PR #32, opened 2026-07-29
feat(operator): upgrade CRDs from an init container

danbar2:
Helm creates the manifests in charts/snapshot/crds/ on a fresh install and
then leaves them alone — `helm upgrade` never updates CRDs by design — so the
operator could end up reconciling against definitions older than the binary
compiled against them.

Add a crd-installer binary that server-side applies the CRDs and runs as the
operator Deployment's init container. It reads the stored resource version
before applying and compares it after, so each definition is reported as
created/updated/unchanged and a run with nothing to do is a visible no-op
rather than a silent write.

The CRDs stay in charts/snapshot/crds/. The operator image build context is
already the repo root, so the Dockerfile copies them to /etc/snapshot/crds
rather than relocating them into the Go module. That also ties the applied
schema to the binary generated from the same api/ types.

crdUpgrade.enabled gates the init container and its RBAC, defaulting to true.
Turning it off makes CRD upgrades the admin's job; values.yaml carries the
notice and the manual apply command.

RBAC is scoped to the two definitions this repo owns. create cannot carry
resourceNames — RBAC has no object name at authorization time for create — so
it stays unscoped, but only ever permits adding a CRD, never modifying one the
operator does not own.

Co-Authored-By: Claude <noreply@anthropic.com>

## Summary by CodeRabbit

* **New Features**
  * Added/updated `PodSnapshot` and `PodSnapshotContent` CRDs with immutable spec validation and `/status` subresources.
  * Operator now server-side applies embedded CRDs during rollout via an init step, configurable with `crdUpgrade.enabled`, `crdUpgrade.logLevel`, and `crdUpgrade.resources` (optional opt-out).
* **Bug Fixes**
  * Updated operator RBAC to support controlled CRD creation/updates for the chart’s CRDs.
* **Tests / Chores**
  * Added installer and embedded-CRD parity tests; improved CRD drift verification and license-header checks.


operator/cmd/crdinstaller/main.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/go-logr/logr"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/ai-dynamo/snapshot/operator/internal/crdinstaller"
+	"github.com/ai-dynamo/snapshot/operator/internal/logging"
+)
+
+// version is overridable at build time via -ldflags "-X main.version=<tag>".
+var version = "dev"
```
Ronkahn21: should it be exported ?

operator/cmd/crdinstaller/main.go
```diff
@@ -0,0 +1,72 @@
...
+
+func main() {
+	crdDir := flag.String("crd-dir", "/etc/snapshot/crds",
```
Ronkahn21: instead of passing the crds as yamls dir
we can import it as string as we do in grove, so we can just import it from the api package

operator/cmd/crdinstaller/main.go
```diff
@@ -0,0 +1,72 @@
...
+		return err
+	}
+
+	ctx := ctrl.SetupSignalHandler()
+	results, err := crdinstaller.InstallCRDs(ctx, cl, log, docs)
+	if err != nil {
+		return err
+	}
+
+	if !crdinstaller.Changed(results) {
```
Ronkahn21: can we make it a property of reuaselt

```suggestion
	if !results.Changed() {
```

operator/Dockerfile
```diff
@@ -9,7 +9,15 @@ COPY api/ api/
 COPY operator/ operator/
 WORKDIR /workspace/operator
 RUN go build -o bin/manager ./cmd/manager
+# The CRD installer ships in the same image; the chart runs it as the operator's
+# init container by overriding the entrypoint.
```
Ronkahn21: redundant

operator/Dockerfile
```diff
@@ -9,7 +9,15 @@ COPY api/ api/
...
+RUN go build -o bin/crd-installer ./cmd/crdinstaller
 
 FROM gcr.io/distroless/static:nonroot
 COPY --from=builder /workspace/operator/bin/manager /manager
+COPY --from=builder /workspace/operator/bin/crd-installer /crd-installer
+# The CRDs controller-gen generates from api/, carried at the path crd-installer
+# reads by default. Baking them in ties the definitions to the binary compiled
+# against the same types, so an image can never apply a schema it disagrees with.
```
Ronkahn21: same

operator/internal/crdinstaller/installer_test.go
```diff
@@ -0,0 +1,241 @@
...
+// built by client.ApplyConfigurationFromUnstructured, which wraps it in an
+// unexported type that promotes the *unstructured.Unstructured methods.
+type applyConfigUnstructured interface {
+	GetName() string
+	SetResourceVersion(string)
+}
+
+// fakeClient records applies and models the API server's resource version
+// behaviour: it only bumps when the stored definition actually changes.
+type fakeClient struct {
```
Ronkahn21: why can we use the regular fake client ?

danbar2: Good question — I tried it, and controller-runtime's fake client can't model the case this installer exists to detect.

The installer decides created/updated/**unchanged** by comparing the stored `resourceVersion` before and after the apply. A real apiserver leaves the resource version alone when a server-side apply changes nothing; the fake client bumps it unconditionally. Probing v0.24.1 directly:

```
after create RV="1"
after identical re-apply RV="2"  -> unchanged detected: false
```

Against a real apiserver (kind v1.33.1) the same sequence reports `unchanged`, which is what produces the `"CRDs already up to date, no changes applied"` no-op path.

So `TestInstallCRDsIsNoOpWhenUpToDate` would assert the opposite of production behaviour if it used the fake client. The hand-written stub is ~25 lines and also lets the tests assert the field manager and `ForceOwnership` are actually passed. I've added a comment above it recording this so it doesn't come up again.

Ronkahn21 (approved):

Ronkahn21 (approved):

PR #40, opened 2026-08-02
chore(agent): add //go:build linux constraints to Linux-only files

oleg-kushniriov:
## Summary by CodeRabbit

* **New Features**
  * Linux builds now include the `nsrestore` utility.

* **Bug Fixes**
  * Checkpoint tests can run across supported platforms instead of being limited to Linux.

* **Chores**
  * Added Docker-based Linux build and test commands.
  * Added a default Go version for build tooling.
  * Improved containerized build configuration with configurable Linux images and more consistent behavior across development environments.


danbar2 (approved):

Ronkahn21 (commented):
I debirtly didn't add buld agent Linux as it silencely fails build, when you build it just bulld empty binary 
I rather it fails on Mac, we should just exposed it better

Makefile
```diff
@@ -13,7 +13,8 @@ TAGS              ?= $(VERSION)
 DOCKER_BUILD_ARGS ?=
 
 .PHONY: tidy generate test build lint verify-generate verify-crds check fmt add-license-headers \
-        verify-license-headers govulncheck helm-lint docker-build-agent docker-build-operator
+        verify-license-headers govulncheck helm-lint docker-build-agent docker-build-operator \
+        linux-build linux-test
```
Ronkahn21, line 17: In the CI it always use linux, so lets remove this and just keep the two target as local only

agent/internal/runtime/overlay_test.go
```diff
@@ -1,6 +1,8 @@
 // SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+//go:build linux
```
Ronkahn21: lets remove this, it should failed on Mac
once we support ARM as well we will add tags

Makefile
```diff
@@ -78,8 +78,26 @@ govulncheck: $(GOVULNCHECK)
 helm-lint: $(HELM)
 	$(HELM) lint charts/snapshot/
 
+LINUX_GO_IMAGE ?= golang:$(GO_VERSION)-bookworm
```
Ronkahn21: move this to the top with all the consts

Ronkahn21 (approved):

PR #61, opened 2026-08-17
feat(agent): own checkpoint PVC and mount restore artifacts

Ronkahn21:
## Summary

Move checkpoint-storage ownership into the Snapshot agent and restore CRIU images from a zero-copy mounted artifact view.

- Mount one shared `ReadWriteMany` checkpoint PVC into every snapshot-agent at the fixed `/checkpoints` path.
- Remove checkpoint PVC injection and physical storage discovery from workload pods, the operator, and `snapshotctl`.
- Mount the selected checkpoint artifact read-only into the restore placeholder instead of hard-linking or copying CRIU image files.
- Rewrite and duplicate only the small `files.img` metadata file when socket state must be unique for a restore.

## Why

Checkpoint artifacts may live on a remote PVC while the agent's temporary storage is local. The old restore flow hard-linked every unchanged CRIU image into a temporary directory, which fails with `EXDEV` across devices. Copying those files is not acceptable because page images may be gigabytes.

CRIU accepts one image-directory FD. A real mount gives it one stable directory containing the original large image files plus an optional locally rewritten `files.img`, without copying the large files or relying on symlinks during namespace reconstruction.

## Restore flow

1. The agent resolves `/checkpoints/<checkpointID>/versions/<version>` from its own fixed configuration. IDs and versions must be clean single path elements.
2. The agent reads and validates the checkpoint manifest from that artifact.
3. The role-specific mount helper installs the executable bundle at `/tmp/snapshot-binaries` and the checkpoint artifact at `/tmp/checkpoint` in the placeholder's mount namespace.
4. Both mounts reuse the same pinned mount-namespace FD. The checkpoint mount is read-only, `nosuid`, `nodev`, and `noexec`.
5. CRIU reads every unchanged image directly from the PVC-backed artifact mount.
6. If socket metadata changes, the agent writes a local temporary `files.img`, creates a detached read-only mount for that inode, moves it over `/tmp/checkpoint/files.img`, and unlinks the temporary pathname.
7. Restore cleanup attempts the replacement-file, checkpoint, and bundle unmounts independently. A cleanup-only failure is logged and emitted as `RestoreCleanupFailed` without killing an otherwise restored workload.

This removes `linkOrCopy`: unchanged image files are neither copied nor linked.

## Deployment impact

- Snapshot uses one agent-owned RWX PVC mounted at `/checkpoints`.
- Workload checkpoint and restore pods no longer mount checkpoint storage.
- The agent watches assigned checkpoint and restore workloads cluster-wide.
- Chart-created checkpoint PVCs use `ReadWriteMany` and are retained across Helm uninstall.
- The obsolete `storage.accessMode`, `storage.pvc.accessMode`, and `rbac.namespaceRestricted` settings are removed.
- Storage and kernel capabilities are exercised by the real checkpoint/restore operation rather than speculative startup probes. Checkpoint publication still fails safely if atomic no-replace rename is unsupported.
- Published checkpoint versions are treated as immutable infrastructure written by trusted snapshot agents. Descriptor-relative `openat2` hardening against a malicious concurrent storage writer is outside this PR's trust model.

## Validation

- CI `build`, `test`, and `check` jobs passed for commit `81a78ca7527b3e92923bf64fba35dc3eb770367b`.
- [Artifact workflow 32247602459](https://github.com/ai-dynamo/snapshot/actions/runs/32247602459) passed and published operator, agent, and chart artifacts as `v0.0.0-g81a78ca7`.
- Linux agent tests, focused namespace-FD tests, C helper compilation with warnings as errors, Helm lint, and Helm unit tests passed locally.
- The configured `ron-poc/ron-vcluster` environment validation passed (`1 passed`).
- The GPU snapshot lifecycle suite passed once against `v0.0.0-g81a78ca7` (`4 passed in 2m43s`), covering successful GPU checkpoint/restore and expected failure paths.

The shared test cluster required its missing `snapshotjobs.nvidia.com` CRD to be installed and the unrelated CRD-upgrade init container disabled because current `main` RBAC does not include that new CRD name.

## Summary by CodeRabbit

- **New Features**
  - Restores now use versioned checkpoint artifacts from shared storage.
  - Cluster-wide agents can restore workloads across namespaces.
  - Bundles and checkpoint artifacts are mounted separately with security protections.
  - Rewritten restore images are mounted read-only.

- **Bug Fixes**
  - Unsafe artifact paths and invalid storage locations are rejected.
  - Cleanup failures are reported without masking successful restores.

- **Breaking Changes**
  - Namespace-restricted deployment and checkpoint storage configuration options were removed.
  - Shared storage requires ReadWriteMany access at `/checkpoints`.
  - Checkpoint location fields were removed from command output.


agent/internal/controller/controller.go
```diff
@@ -116,10 +111,11 @@ func NewNodeController(
 		return nil, fmt.Errorf("failed to create dynamic client: %w", err)
 	}
 
-	injector, err := nsmount.New(nsmount.SnapshotBinSrc, nsmount.SnapshotBinDst, log)
+	nsm, err := nsmount.New(log)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create binary injector: %w", err)
 	}
+	injector := executor.Mounters{Bundle: nsm, Artifact: nsm}
```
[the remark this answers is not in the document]
Ronkahn21: Fixed in 8726e89. Artifact mounts now use a role-specific noexec policy while the injected binary bundle remains executable. MountOptions and ns-bind-mount propagate MOUNT_ATTR_NOEXEC, with unit coverage for both roles.

agent/internal/criu/restore.go
```diff
@@ -42,15 +42,17 @@ func ExecuteRestore(
...
 		return 0, nil, fmt.Errorf("failed to prepare CRIU image directory: %w", err)
 	}
 	var confTmpDir string
 	cleanup := func() {
 		closeFiles(inheritedFiles)
 		closeFiles(openFiles)
-		removeImageDir()
+		if err := removeImageView(); err != nil {
+			log.Error(err, "failed to remove private CRIU image view", "path", imageDirPath)
+		}
```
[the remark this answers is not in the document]
Ronkahn21: Fixed in 8726e89, with one correction to the proposed destination: the image directory is read-only, so it cannot be the fallback work directory. ExecuteRestore now always sets WorkDirFd to either the configured writable directory or the writable temporary directory created for the CRIU override, and diagnostics use that same effective path. Both cases have unit coverage.

charts/snapshot/templates/pvc.yaml
```diff
@@ -11,9 +11,11 @@ metadata:
     app.kubernetes.io/name: {{ include "snapshot.name" . }}
     app.kubernetes.io/instance: {{ .Release.Name }}
     app.kubernetes.io/component: storage
+  annotations:
+    helm.sh/resource-policy: keep
 spec:
   accessModes:
-    - {{ .Values.storage.pvc.accessMode }}
+    - ReadWriteMany
```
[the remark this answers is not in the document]
Ronkahn21: Fixed in 8726e89. The chart now uses lookup to reject both retained chart-owned and reused external PVCs that lack ReadWriteMany, without attempting to mutate accessModes. Documentation gives the data-preserving migration path to a new RWX claim, and Helm tests cover retained RWO rejection, reused RWO rejection, and existing RWX acceptance.

agent/internal/criu/restore_image_mount_linux.go
```diff
@@ -0,0 +1,87 @@
...
+	closeFD:      unix.Close,
+}
+
+func mountFilesImage(checkpointPath, replacementFilesImagePath string) (func() error, error) {
+	// checkpointPath is the artifact mount in one restore's private placeholder
+	// mount namespace. Mounting two replacements over the same path in one
+	// namespace is unsupported because the mounts would stack and cleanup order
+	// could expose the wrong restore's metadata.
+	return mountFilesImageWithOps(checkpointPath, replacementFilesImagePath, realMountFilesImageOps)
+}
```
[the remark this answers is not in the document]
Ronkahn21: No code change for this thread. snapshot-agent is a Linux-only privileged daemon and its existing runtime package already depends on Linux mount and namespace syscalls, so non-Linux builds are unsupported before this restore-image path is reached. Adding a narrow stub or tags here would imply a portability contract the agent does not provide; the Linux-specific mount implementation remains explicitly isolated in the _linux file.

charts/snapshot/README.md
```diff
@@ -141,20 +121,17 @@ kubectl get pods -n ${NAMESPACE} -l app.kubernetes.io/name=snapshot -o wide
 | `image.agent.repository` | Agent image repository | `ghcr.io/ai-dynamo/snapshot/agent` |
 | `image.agent.tag` | Agent image tag (empty = chart appVersion) | `""` |
 | `storage.type` | Snapshot-owned storage backend | `pvc` |
-| `storage.accessMode` | `agentMount` for namespace-local agent PVC mount, or `podMount` for a single cluster agent that accesses workload-mounted PVCs | `agentMount` |
-| `storage.pvc.create` | Create `snapshot-pvc` instead of using an existing PVC in `agentMount` mode | `true` |
-| `storage.pvc.name` | Checkpoint PVC name. Mounted by the agent in `agentMount` mode, or by workload pods in `podMount` mode | `snapshot-pvc` |
+| `storage.pvc.create` | Create `snapshot-pvc` instead of using an existing shared PVC | `true` |
+| `storage.pvc.name` | Shared RWX checkpoint PVC mounted by every agent | `snapshot-pvc` |
```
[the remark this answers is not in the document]
Ronkahn21: Fixed in 8726e89. The verification and explicit cleanup examples now use PVC_NAME with snapshot-pvc only as the default, and instruct users to match storage.pvc.name.

agent/internal/executor/restore.go
```diff
@@ -62,35 +67,51 @@ func Restore(ctx context.Context, rt snapshotruntime.Runtime, log logr.Logger, r
...
+	artifactMount, err := mounts.Artifact.Mount(ctx, snap.PlaceholderPID, artifactPath, nsmount.CheckpointDst)
+	if err != nil {
+		return 0, fmt.Errorf("mount checkpoint artifact into placeholder: %w", err)
+	}
+	defer func() {
+		if cleanupErr := artifactMount.Unmount(context.Background()); cleanupErr != nil {
+			log.Error(cleanupErr, "failed to clean artifact mount from placeholder namespace")
+		}
+	}()
+	injectDuration := time.Since(injectStart)
```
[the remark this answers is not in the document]
Ronkahn21: Fixed in 8726e89. Both deferred unmount handlers now propagate cleanup failure through retErr when restore otherwise succeeded, while retaining any pre-existing restore error. The successful return still carries the placeholder PID, and cleanup error precedence has unit coverage.

agent/internal/nsmount/mount.go
```diff
@@ -153,13 +109,11 @@ func (m *execMounter) Mount(ctx context.Context, pid int, src, dst string, opts
...
 	m.log.Info("mounted into namespace", "src", src, "dst", dst, "readonly", opts.ReadOnly, "pid", pid)
 
-	createdDst := strings.Contains(stdout.String(), "created_dst=1")
-
 	return &execMountRef{
 		binaryPath: m.binaryPath,
 		nsFd:       nsFd,
 		dst:        dst,
-		createdDst: createdDst,
+		createdDst: strings.Contains(stdout.String(), "created_dst=1"),
```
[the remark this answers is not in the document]
Ronkahn21: Fixed in 8726e89. The createdDst assignment now documents that mount-fd emits created_dst=1 after attachment and that the value controls whether umount-fd removes the helper-created destination.

operator/internal/protocol/restore.go
```diff
@@ -161,7 +143,7 @@ func ensureRestoreStartupProbe(container *corev1.Container) {
 func ValidateRestorePodSpec(
 	podSpec *corev1.PodSpec,
 	annotations map[string]string,
-	storage snapshotv1alpha1.Storage,
+	_ snapshotv1alpha1.Storage,
```
Ronkahn21: lets remove this

agent/internal/criu/restore_images.go
```diff
@@ -47,48 +48,87 @@ func prepareRestoreImageDir(checkpointPath string) (string, func(), error) {
...
 
 	rewritten := false
 	for _, entry := range image.Entries {
 		fileEntry := entry.Message.(*fdinfo.FileEntry)
-		if fileEntry.GetType() != fdinfo.FdTypes_UNIXSK || fileEntry.Usk == nil {
-			continue
-		}
-		if rewriteCloneConflictingUnixSocketAddress(fileEntry.Usk, restoreID) {
+		if fileEntry.GetType() == fdinfo.FdTypes_UNIXSK && fileEntry.Usk != nil &&
+			rewriteCloneConflictingUnixSocketAddress(fileEntry.Usk, restoreID) {
```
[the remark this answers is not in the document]
Ronkahn21: Addressed in b0784db. We fail closed rather than silently skipping the entry because a decoded files.img entry is expected to be *fdinfo.FileEntry. An unexpected type now returns an indexed error and closes any reserved TCP-port descriptors. A regression test covers this path.

api/v1alpha1/protocol.go
```diff
@@ -22,6 +22,7 @@ type Storage struct {
 type RestoreStatusAnnotationKeys struct {
 	Status      string
 	ContainerID string
+	Reason      string
```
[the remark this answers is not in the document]
Ronkahn21: Addressed in b0784db by adding the required // +optional marker to RestoreStatusAnnotationKeys.Reason.

api/v1alpha1/protocol.go
```diff
@@ -129,13 +131,19 @@ func RestoreStatusAnnotationKeysFor(containerName string) (RestoreStatusAnnotati
...
+// RestoreStatusAnnotationsWithReason includes the controller-facing condition reason.
+func RestoreStatusAnnotationsWithReason(containerName, status, containerID, reason string) (map[string]string, error) {
 	keys, err := RestoreStatusAnnotationKeysFor(containerName)
 	if err != nil {
 		return nil, err
 	}
 	return map[string]string{
 		keys.Status:      status,
 		keys.ContainerID: containerID,
+		keys.Reason:      reason,
```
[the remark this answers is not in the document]
Ronkahn21: Intentionally unchanged. This PR introduces a new restore-status annotation contract rather than preserving the legacy two-key shape. Both helpers emit the reason key; an empty value explicitly represents no failure reason.

agent/cmd/ns-bind-mount/main.c
```diff
@@ -60,30 +61,31 @@ struct mount_attr {
...
- *   SnapshotBinSrc = "/snapshot-binaries"       → ALLOWED_SRC_PREFIX
- *   SnapshotBinDst = "/tmp/snapshot-binaries"   → ALLOWED_DST_PREFIX
- * Keep them in sync when either changes. */
-#define ALLOWED_SRC_PREFIX "/snapshot-binaries"     /* = nsmount.SnapshotBinSrc */
-#define ALLOWED_DST_PREFIX "/tmp/snapshot-binaries" /* = nsmount.SnapshotBinDst */
-
-/* Returns 0 if path is absolute and begins with allowed_prefix, -1 otherwise. */
 static int
-check_path_prefix(const char* path, const char* allowed_prefix, const char* label)
+check_source(const char* path)
```
oleg-kushniriov: `check_source` only checks that the path is absolute (plus a separate ..-component check) - any absolute path is now accepted as a mount source. This binary runs privileged `(setns(CLONE_NEWNS), effectively CAP_SYS_ADMIN)` and was previously the last line of defense against a Go-side bug feeding it an arbitrary host path. That defense-in-depth is now gone: all source validation lives solely in nsmount.ResolveArtifact/ResolveArtifactPath in Go. A future bug, or any new caller of the mounter that doesn't route through ResolveArtifact, would let this privileged helper clone-mount any absolute host path (/etc, /, /proc, etc.) read-only into a workload's `/tmp/checkpoint` or `/tmp/snapshot-binaries`, with no compile-time or runtime guard inside the privileged binary itself. 

Suggestion: Have the C helper accept the configured checkpoint base path (and `/snapshot-binaries`) as an argument/env var and validate src against it, rather than accepting any absolute path unconditionally. Keep the check in the privileged component, not only in the caller.

agent/internal/nsmount/artifact.go
```diff
@@ -0,0 +1,38 @@
...
+	"path/filepath"
+	"strings"
+
+	snapshotprotocol "github.com/ai-dynamo/snapshot/api/v1alpha1"
+)
+
+// ResolveArtifactPath returns the existing checkpoint artifact layout rooted
+// at the agent-owned base path. All variable components must be single clean
+// path elements.
+func ResolveArtifactPath(basePath, artifactID, version string) (string, error) {
```
oleg-kushniriov: The artifact layout is` <basePath>/<checkpointID>/versions/<version>` - no namespace, no pod UID, no owner component. Checkpoint IDs are caller-supplied (snapshotctl --checkpoint-id, PodOptions.CheckpointID, or a generated manual-snapshot-<unixnano>), and are stamped onto pods as a plain label. Before this PR two independent mechanisms kept namespaces apart: the PVC was namespace-local (agentMount + rbac.namespaceRestricted), or workload pods mounted their own namespace-local PVC (podMount). Both are gone. Now a single PVC backs the whole cluster and every agent watches all namespaces.

The only validation the restore path performs is manifest.CheckpointID != req.CheckpointID (executor/restore.go:82). Nothing compares manifest.SourcePod.Namespace - which is present in the manifest - against the restore pod's namespace.

Failure scenario: A user in namespace tenant-b creates a pod with labels nvidia.com/snapshot-restore-target: "true" and nvidia.com/snapshot-checkpoint-id: <tenant-a's checkpoint id> plus the standard restore annotations. The agent on that node resolves <basePath>/<id>/versions/1, mounts tenant-a's artifact read-only into the placeholder, and CRIU restores tenant-a's process image - full process memory, open FDs, in-memory credentials and tokens - as a running container in tenant-b. Checkpoint IDs are guessable (manual-snapshot-<nanos>) or discoverable from any pod label the user can read.

Secondary failure scenario: Two namespaces independently pick the same checkpoint ID. executor/checkpoint.go:111 does os.RemoveAll(finalDir) then os.Rename(tmpDir, finalDir) - namespace B's capture silently destroys namespace A's artifact.

Suggestion: Put the tenancy boundary back into the path and enforce it at restore. 

Concretely: 
(a) key the layout on namespace, e.g.<basePath>/<namespace>/<checkpointID>/versions/<version> - the agent already knows `pod.Namespace` at `artifactPathForPod`, and `validatePathElement` applies cleanly to a namespace component; 
and (b) in `executor.Restore`, reject when `manifest.SourcePod.Namespace != req.PodNamespace `unless an explicit cross-namespace-restore opt-in is set. Either alone is weaker than both. If cross-namespace restore is an intended product feature, it needs to be an explicit, authorized operation rather than an emergent consequence of the storage layout, and it needs to be called out in the PR description and the chart README as a security-relevant deployment property.

agent/internal/executor/restore.go
```diff
@@ -128,12 +161,12 @@ func Restore(ctx context.Context, rt snapshotruntime.Runtime, log logr.Logger, r
 	return snap.PlaceholderPID, nil
 }
 
-func mountBundle(ctx context.Context, mounter Mounter, pid int) (nsmount.MountPoint, error) {
-	mp, err := mounter.Mount(ctx, pid)
-	if err != nil {
-		return nil, fmt.Errorf("mount agent bundle into placeholder: %w", err)
+func setCleanupErrorIfSuccessful(retErr *error, action string, cleanupErr error) {
```
oleg-kushniriov: The name reads as "set the cleanup error if the cleanup was successful", which is self-contradictory.

Consider to rename to something that states the rule, e.g. `recordCleanupFailureIfRestoreSucceeded`

agent/internal/controller/controller.go
```diff
@@ -528,109 +513,172 @@ func (w *NodeController) startRestoreForContainer(
...
 		restoreCtx, cancel = context.WithTimeout(ctx, timeout)
 		defer cancel()
 	}
+
+	ready, err := op.beginRestore(ctx)
+	if err != nil || !ready {
+		return err
+	}
+	placeholderHostPID, err := op.executeRestore(restoreCtx)
+	if err != nil {
```
oleg-kushniriov: Didn't we agree this should stay log-only as default behaviour when the restore itself succeeded and only the unmount failed? And make it configurable later with both kill/log options?

agent/cmd/ns-bind-mount/main.c
oleg-kushniriov: Before, helper checked that src starts with /snapshot-binaries. Now check_source only checks that path is absolute.

The old comment said the reason: "Both the Go layer and this binary enforce these prefixes so that a bug in the controller cannot cause arbitrary host paths to be mounted". So this was second layer on purpose, and now we have only Go side.

If some bug or new call site passes wrong src, any host directory can be mounted into workload container. It is read-only + nosuid + nodev + noexec, so no escalation, but data can still be read.

Can we pass the configured base path to the helper (argument or env) and check it there too?

agent/internal/controller/podsnapshotcontent.go
```diff
@@ -497,9 +491,13 @@ func isContentTerminal(content *snapshotv1alpha1.PodSnapshotContent) bool {
 }
 
 // artifactPresent reports whether a completed checkpoint directory already exists on disk.
-func artifactPresent(destination string) bool {
+func artifactPresent(destination, checkpointID string) bool {
```
oleg-kushniriov: artifactPresent now also reads manifest, and any error returns false.

If manifest read fails only temporary (NFS returns EIO/ESTALE), we think artifact is missing. Then capture runs again and executor/checkpoint.go (Checkpoint) does os.RemoveAll(finalDir) - so we delete a good artifact. And because storage is now shared between nodes, another agent can restore from this artifact in same time, and its CRIU read will fail.

I think we should separate "manifest not exist" from "manifest cannot be read". For the second case better to return error and retry, not to recapture.

agent/internal/executor/restore.go
```diff
@@ -26,23 +25,39 @@ import (
...
 
-// Mounter mounts the agent binary bundle into a placeholder container's mount namespace.
+// Mounter mounts src at dst inside a placeholder container's mount namespace.
 type Mounter interface {
-	Mount(ctx context.Context, pid int) (nsmount.MountPoint, error)
+	Mount(ctx context.Context, pid int, src, dst string) (nsmount.MountPoint, error)
 }
 
+// Mounters carries the executable bundle and non-executable artifact policies.
+type Mounters struct {
```
oleg-kushniriov: Mounters is a struct with two interface fields and no constructor. Zero value gives nil panic in mounts.Bundle.Mount. Also the two fields differ only by NoExec.

Maybe simpler: one Mounter and pass options per call, or add constructor that checks nil.

agent/internal/types/config.go
```diff
@@ -88,9 +66,8 @@ func (c *AgentConfig) Validate() error {
 
 // StorageSpec holds snapshot storage settings that are local to the agent deployment.
 type StorageSpec struct {
-	Type       string `yaml:"type"`
-	BasePath   string `yaml:"basePath"`
-	AccessMode string `yaml:"accessMode"`
+	Type     string `yaml:"type"`
```
oleg-kushniriov, line 66: After this PR nothing reads storage.Type in agent, but we still render it to ConfigMap. Same for ResolveCheckpointStorage, ResolveRestoreStorage, CheckpointVolumeName - now only tests use them. Some of them already have TODO: dead code, the new ones do not. 

Can we mark them too?

agent/internal/executor/restore.go
oleg-kushniriov: Also *error is pointer to interface (gocritic ptrToRefParam), and the function is 3 lines used 2 times in the same file. Inside defer we already have the named return, so we can just write if retErr == nil { ... } there.

agent/internal/criu/restore_images.go
```diff
@@ -109,56 +153,58 @@ func prepareRestoreImageDirForRestoreID(checkpointPath string, restoreID uint64)
...
-		return "", nil, err
+	if err := filesImage.Close(); err != nil {
+		_ = os.Remove(path)
+		return "", fmt.Errorf("failed to close private %s: %w", filesImageFilename, err)
 	}
+	return path, nil
+}
 
-	entries, err := os.ReadDir(checkpointPath)
+func mountReplacementFilesImage(
```
oleg-kushniriov, line 190: In the error path we call removeImageView() and then os.Remove(replacementFilesImagePath) again, on the path where remove already failed. Second call cannot help and error is ignored. Probbaly can delete it.

agent/internal/executor/restore.go
```diff
@@ -62,35 +77,53 @@ func Restore(ctx context.Context, rt snapshotruntime.Runtime, log logr.Logger, r
...
 
+	artifactPath, err := nsmount.ResolveArtifact(req.BasePath, req.CheckpointID, req.ArtifactVersion)
+	if err != nil {
+		return 0, fmt.Errorf("resolve checkpoint artifact: %w", err)
+	}
+	manifest, err := types.ReadManifest(artifactPath)
+	if err != nil {
+		return 0, fmt.Errorf("read checkpoint manifest: %w", err)
+	}
+	if manifest.CheckpointID != req.CheckpointID {
```
oleg-kushniriov: Consider validating `manifest.SourcePod.Namespace != req.PodNamespace` too

agent/cmd/ns-bind-mount/main.c
Ronkahn21: Addressed in 8341159. I agree that the privileged helper must independently enforce the source policy, but a caller-provided root argument or environment variable would defeat that boundary. The helper now derives a compiled-in source root solely from the fixed destination: `/tmp/snapshot-binaries` maps to `/snapshot-binaries`, and `/tmp/checkpoint` maps to `/checkpoints`. It validates clean portable-ASCII absolute paths, component-boundary containment, and rejects traversal, sibling prefixes, special characters, Unicode, and role swaps before any mount or namespace syscall. Go independently enforces the same fixed mapping.

Ronkahn21: Addressed in 8341159. I agree that the privileged helper must independently enforce the source policy, but a caller-provided root argument or environment variable would defeat that boundary. The helper now derives a compiled-in source root solely from the fixed destination: `/tmp/snapshot-binaries` maps to `/snapshot-binaries`, and `/tmp/checkpoint` maps to `/checkpoints`. It validates clean portable-ASCII absolute paths, component-boundary containment, and rejects traversal, sibling prefixes, special characters, Unicode, and role swaps before any mount or namespace syscall. Go independently enforces the same fixed mapping.

operator/internal/protocol/restore.go
Ronkahn21: Removed in 8341159 together with the obsolete restore-storage protocol path.

agent/internal/executor/restore.go
Ronkahn21: Addressed in 8341159. The helper was removed, cleanup handling is inline, and typed errors are created through `NewRestoreCleanupError`.

agent/internal/controller/controller.go
Ronkahn21: Addressed in 8341159 per the agreed default. The executor logs the cleanup failure; the controller recognizes `RestoreCleanupError`, completes the restore, emits no failure event or failure reason, and does not kill the restored workload. Configurable policy remains a follow-up.

agent/internal/controller/podsnapshotcontent.go
Ronkahn21: Addressed in 8341159. `artifactPresent` now returns `(bool, error)`; only a missing final directory means absent, while manifest errors including EIO and ESTALE propagate. Existing artifacts are never deleted, publication is no-replace, and a post-lease recheck closes the race. Tests cover storage errors, manifest mismatch, and immutable collisions.

agent/internal/executor/restore.go
Ronkahn21: Addressed in 8341159. The two-field wrapper was replaced by one `RestoreMounter` interface with role methods. `NSMounter` owns the fixed paths and policies, and restore rejects a nil mounter.

agent/internal/types/config.go
Ronkahn21: Addressed in 8341159. Removed the dead storage protocol constants, types, and resolver/apply helpers and regenerated deepcopy code. `AgentConfig.Storage.Type` remains live because startup validation and the chart use it to reject unsupported backends.

agent/internal/executor/restore.go
Ronkahn21: Addressed in 8341159. Deleted the pointer-to-interface helper and inlined the `retErr == nil` rule in each defer.

agent/internal/criu/restore_images.go
Ronkahn21: Addressed in 8341159. Removed the redundant ignored unlink retry.

agent/internal/executor/restore.go
Ronkahn21: Addressed in 8341159. Restore now validates that manifest `K8s.PodNamespace` equals the requested pod namespace before inspection or mounting; mismatch coverage was added.

agent/internal/controller/controller.go
```diff
@@ -528,109 +510,166 @@ func (w *NodeController) startRestoreForContainer(
...
+		var cleanupErr *executor.RestoreCleanupError
+		if !errors.As(err, &cleanupErr) {
+			return op.failRestore(ctx, err)
 		}
-		return nil
 	}
+	return op.completeRestore(ctx, placeholderHostPID)
```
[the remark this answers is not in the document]
Ronkahn21: Keeping this log-only in the current PR by design. The executor retains an error log with the cleanup action and failure context, while the controller completes the successful restore without a failure event, failure reason, or kill. Operator-visible signaling will be considered together with the configurable cleanup policy in a follow-up.

agent/internal/controller/podsnapshotcontent_test.go
```diff
@@ -98,6 +99,53 @@ func makeNodeController(t *testing.T, fc *fakeCheckpointer, objs ...client.Objec
...
+			present, err := artifactPresentWithManifestReader(
+				t.TempDir(),
+				"checkpoint-123",
+				func(string) (*snapshottypes.CheckpointManifest, error) { return nil, readErr },
+			)
+			require.ErrorIs(t, err, readErr)
+			assert.False(t, present)
+		})
+	}
+}
```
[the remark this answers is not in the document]
Ronkahn21: Addressed in 883d5c3. Added the existing-directory-without-manifest boundary case and verified that it returns an error rather than triggering recapture.

agent/internal/executor/publish_linux.go
```diff
@@ -0,0 +1,12 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build linux
+
+package executor
+
+import "golang.org/x/sys/unix"
+
+func renameNoReplace(oldPath, newPath string) error {
+	return unix.Renameat2(unix.AT_FDCWD, oldPath, unix.AT_FDCWD, newPath, unix.RENAME_NOREPLACE)
```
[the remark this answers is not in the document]
Ronkahn21: Addressed in 883d5c3. Agent startup now probes the mounted checkpoint filesystem before runtime/controller initialization. It creates an isolated temporary probe under the fixed checkpoint root and requires atomic no-replace rename to return EEXIST when the destination exists; EINVAL, ENOTSUP, ENOSYS, or any other failure produces a clear startup error. The probe is removed afterward and has Linux coverage.

agent/internal/executor/publish_other.go
```diff
@@ -0,0 +1,20 @@
...
+	"syscall"
+)
+
+func renameNoReplace(oldPath, newPath string) error {
+	if _, err := os.Lstat(newPath); err == nil {
+		return syscall.EEXIST
+	} else if !os.IsNotExist(err) {
+		return err
+	}
+	return os.Rename(oldPath, newPath)
```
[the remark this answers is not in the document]
Ronkahn21: Addressed in 883d5c3. The non-Linux implementation no longer uses the racy Lstat-plus-Rename fallback; it returns ENOTSUP, with tests covering both the primitive and checkpoint publication.

agent/internal/nsmount/mount_test.go
```diff
@@ -33,155 +31,117 @@ func newMounterForTest(t *testing.T, bin string) *execMounter {
...
+			}
+			cmd := exec.Command(binary, fmt.Sprintf("%d", os.Getpid()), tc.source, destination, "ro", "noexec")
+			output, err := cmd.CombinedOutput()
+			if err == nil {
+				t.Fatalf("helper accepted source %q", tc.source)
+			}
+			if strings.Contains(string(output), "open_tree") || strings.Contains(string(output), "setns") {
+				t.Fatalf("helper reached mount syscall for invalid source: %s", output)
+			}
+		})
```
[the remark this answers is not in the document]
Ronkahn21: Addressed in 883d5c3. The C-helper regression test now uses the production `mount-fd`, inherited-FD-number, source, destination argument layout and still verifies rejection before open_tree or setns.

charts/snapshot/tests/config_test.yaml
```diff
@@ -0,0 +1,18 @@
...
+    asserts:
+      - isKind:
+          of: ConfigMap
+
+  - it: rejects a caller-selected checkpoint mount path
+    set:
+      storage.pvc.basePath: /other
+    asserts:
+      - failedTemplate:
+          errorPattern: "storage.pvc.basePath is fixed at /checkpoints"
```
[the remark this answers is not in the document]
Ronkahn21: Addressed in 883d5c3. Added independent Helm regression cases for removed `storage.accessMode` and `storage.pvc.accessMode`; the chart suite now passes 7 tests.

.github/workflows/e2e.yaml
```diff
@@ -33,6 +33,7 @@ jobs:
       SNAPSHOT_E2E_SNAPSHOT_RELEASE: snapshot
       SNAPSHOT_E2E_PVC_NAME: snapshot-pvc
       SNAPSHOT_E2E_PVC_SIZE: 2Gi
+      SNAPSHOT_E2E_STORAGE_CLASS: azurefile-csi
```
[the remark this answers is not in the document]
Ronkahn21: No change here. This E2E workflow tests a previously published operator/agent image pair, not the source checkout. The repository publication workflow checks out the exact image SHA, runs `make check`, and makes both image-build jobs depend on that validation. Running `make check` again in E2E could validate a different checkout than the selected image and would duplicate the publication gate.

e2e/snapshot_e2e/infra/setup.py
```diff
@@ -795,30 +799,43 @@ def snapshot_chart_fullname(release: str) -> str:
...
+        detail = (
+            f"phase={pvc.status.phase} storageClass={pvc.spec.storage_class_name}"
+        )
         mismatches = []
         if requested != size:
             mismatches.append(f"requested storage={requested}, configured size={size}")
-        if access_modes != ["ReadWriteOnce"]:
-            mismatches.append(f"accessModes={access_modes}, expected ['ReadWriteOnce']")
+        if access_modes != ["ReadWriteMany"]:
+            mismatches.append(f"accessModes={access_modes}, expected ['ReadWriteMany']")
```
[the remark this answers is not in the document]
Ronkahn21: Addressed in 883d5c3. Existing PVC validation now accepts any access-mode list containing `ReadWriteMany`, regardless of ordering or additional requested modes, and still reports a mismatch when RWX is absent.

agent/internal/nsmount/injector.go
```diff
@@ -1,121 +1,104 @@
...
-	src     string
-	dst     string
 	mounter mounter
 	log     logr.Logger
 }
 
-// probeKernelMountAPI verifies that mount_setattr (Linux 5.12) is available.
-// Called with a NULL attr, so a kernel that implements the syscall fails in
-// copy_struct_from_user with EFAULT; ENOSYS means the kernel is too old.
-func probeKernelMountAPI() error {
```
oleg-kushniriov, line 61: dropped startup fail-fast checks. The kernel `mount_setattr` probe and the ns-bind-mount binary-presence check were both removed with nothing replacing them. A bad node now fails during the first checkpoint/restore instead of at startup.

agent/internal/types/config.go
```diff
@@ -11,33 +11,24 @@ import (
...
 func (c *AgentConfig) LoadEnvOverrides() {
 	if v := os.Getenv("NODE_NAME"); v != "" {
 		c.NodeName = v
 	}
-	if v := os.Getenv("RESTRICTED_NAMESPACE"); v != "" {
-		c.RestrictedNamespace = v
-	}
 }
 
 func (c *AgentConfig) Validate() error {
```
oleg-kushniriov, line 34: PR description says a startup check that the checkpoint filesystem supports `RENAME_NOREPLACE`. No such check exists; 
The real capability is only exercised lazily via `publishCheckpoint/renameNoReplace` on first checkpoint.

Ronkahn21: Agreed; the code exercises RENAME_NOREPLACE only during real checkpoint publication. I corrected the PR description so it no longer claims a startup filesystem-capability check.

agent/internal/nsmount/artifact_linux_test.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+	return base, artifact
+}
+
+func TestResolveArtifact(t *testing.T) {
+	base, artifact := makeArtifactTree(t)
+	got, err := ResolveArtifact(base, "checkpoint-123", "1")
+	if err != nil || got != artifact {
+		t.Fatalf("ResolveArtifact() = %q, %v; want %q", got, err, artifact)
+	}
```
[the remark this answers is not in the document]
Ronkahn21: Fixed in 3c0a236: makeArtifactTree now canonicalizes t.TempDir() once with filepath.EvalSymlinks before constructing and returning the artifact path. The focused nsmount/controller tests also pass in a real Linux container.

agent/internal/nsmount/artifact.go
```diff
@@ -0,0 +1,67 @@
...
+
+func validateWithin(root, source string) error {
+	if err := validateAbsolutePath(source); err != nil {
+		return err
+	}
+	if source != root && !strings.HasPrefix(source, root+"/") {
+		return fmt.Errorf("mount source %q must be within %q", source, root)
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
Ronkahn21: Fixed in 3c0a236: validateWithin now validates root with validateAbsolutePath before validating source and comparing containment. Focused tests cover relative, trailing-slash, and traversal roots.

agent/cmd/ns-bind-mount/main.c
```diff
@@ -60,378 +52,236 @@ struct mount_attr {
...
-  int tree_fd = sys_open_tree(AT_FDCWD, src, OPEN_TREE_CLONE | O_CLOEXEC);
+  int tree_fd = (int)syscall(
+      __NR_open_tree,
+      AT_FDCWD,
+      src,
+      OPEN_TREE_CLONE | O_CLOEXEC);
   if (tree_fd < 0) {
     fprintf(stderr, "open_tree %s: %s\n", src, strerror(errno));
     return 1;
   }
```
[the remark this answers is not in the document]
Ronkahn21: This is the same trust-boundary decision as the Go-side TOCTOU thread. Published artifact directories are treated as immutable and PVC writers/agents are trusted. The helper still rejects malformed paths and existing traversal/symlink inputs, while atomic descriptor-relative openat2 resolution is documented as defense-in-depth follow-up rather than part of this storage-mount PR.

agent/internal/controller/controller.go
```diff
@@ -528,109 +539,147 @@ func (w *NodeController) startRestoreForContainer(
...
+		}
+		return fmt.Errorf("failed to update restore status %q: %w", value, err)
+	}
+	if terminalRestoreStatus(value) {
+		// Keep the attempt key for this controller lifetime so a stale runtime
+		// resolver cannot restart a terminal restore.
+		op.releaseOnExit = false
+	}
 	return nil
 }
```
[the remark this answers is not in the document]
Ronkahn21: Good catch. This is valid at the controller status-authority layer, but it is intentionally deferred from this PVC/namespace-mount PR. The follow-up changes the current lease/status ownership, so terminal-annotation retry semantics should be implemented there without coupling them to this storage-mount change. No code change in this PR.

agent/internal/nsmount/injector_test.go
```diff
@@ -7,142 +7,140 @@ import (
...
 
-func (h *fakemountRef) TargetPath() string { return h.dst }
-func (h *fakemountRef) NsFd() *os.File     { return nil }
-
-func (h *fakemountRef) Unmount(_ context.Context) error {
+func (h *fakeMountRef) NsFd() *os.File { return os.Stdin }
+func (h *fakeMountRef) Unmount(context.Context) error {
 	*h.unmountLog = append(*h.unmountLog, h.dst)
 	return nil
 }
```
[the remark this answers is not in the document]
Ronkahn21: Fixed in 3c0a236. fakeMountRef now owns a unique namespace fd, the nsmount test verifies MountArtifact receives the bundle namespace fd, and the controller test records and asserts the exact MountPoint returned by MountBundle. The focused Linux tests pass.

agent/internal/nsmount/injector.go
Ronkahn21: The startup probes were removed intentionally. This flow only checks that the requested checkpoint exists; it does not run speculative kernel/helper behavior probes. The helper binary is guaranteed by the built image/chart, and any kernel or helper incompatibility is reported by the real mount operation. The PR description now reflects that behavior.

agent/internal/nsmount/artifact_mount.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+	info, err := os.Stat(artifactPath)
+	if err != nil {
+		return "", fmt.Errorf("stat artifact %s: %w", artifactPath, err)
+	}
+	if !info.IsDir() {
+		return "", fmt.Errorf("artifact %s is not a directory", artifactPath)
+	}
+	return artifactPath, nil
+}
```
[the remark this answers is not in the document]
Ronkahn21: This is intentionally outside the current PR's trust boundary. Published checkpoint version directories are treated as immutable, and PVC writers/agents are trusted. The path checks protect against malformed, traversal, and pre-existing symlink input; they do not claim to defeat a malicious concurrent writer. Descriptor-relative openat2 resolution is useful defense-in-depth and is explicitly documented as follow-up work in the PR description.

agent/internal/executor/restore.go
```diff
@@ -158,35 +213,29 @@ func validateRestoredProcess(targetRoot string, restoredPID int, log logr.Logger
...
-		return nil, 0, fmt.Errorf("failed to resolve checkpoint path: %w", err)
-	}
-	if checkpointAbs != baseAbs && !strings.HasPrefix(checkpointAbs, baseAbs+string(os.PathSeparator)) {
-		return nil, 0, fmt.Errorf("invalid checkpoint id %q", req.CheckpointID)
-	}
-
-	m, err := types.ReadManifest(checkpointPath)
-	if err != nil {
-		return nil, 0, fmt.Errorf("failed to read checkpoint manifest: %w", err)
+func validateRestoreManifest(req RestoreRequest, manifest *types.CheckpointManifest) error {
```
oleg-kushniriov, line 216: doesn't check the checkpoint's namespace against the requesting pod

charts/snapshot/templates/role.yaml
```diff
@@ -52,10 +15,6 @@ rules:
   - apiGroups: ["nvidia.com"]
     resources: ["podsnapshotcontents/status"]
     verbs: ["update", "patch"]
-{{- end }}
-
-{{- if .Values.rbac.create }}
-{{- if not .Values.rbac.namespaceRestricted }}
```
oleg-kushniriov, line 58: Is it a breaking change?
Should we note somewhere about it?

charts/snapshot/README.md
```diff
@@ -29,22 +22,16 @@ can require that they already exist.
...
+mount it concurrently. Chart-created claims always request `ReadWriteMany` and
+are retained when the Helm release is removed.
 
-Because `podMount` reaches storage through `/host/proc/<pid>/root`, the target
-container must still be alive and visible through host proc when the agent starts
-checkpoint or restore. A container restart, exited placeholder, runtime PID
-lookup failure, or node security policy that blocks host proc traversal will fail
-or defer that attempt until reconciliation sees a fresh container.
+An existing `ReadWriteOnce` claim cannot be upgraded in place because PVC access
+modes are immutable. The chart rejects retained and externally supplied claims
```
oleg-kushniriov: Despite the doc - pvc.yaml only sets ReadWriteMany when storage.pvc.create=true. The documented create=false migration path has no check on the existing PVC's access mode - failure surfaces later as confusing mount contention.

oleg-kushniriov (approved):

agent/internal/executor/restore.go
Ronkahn21: This is intentional in this PR. manifest.K8s.PodNamespace identifies the source pod and must not be required to equal the target pod namespace; that comparison would prevent cross-namespace restore. The follow-up makes the PodSnapshotContent UID authoritative, so source/target namespace equality does not belong in this validation.

charts/snapshot/README.md
Ronkahn21: Fixed in 7a7af7c. The README no longer claims the chart validates externally supplied claims. It now states that storage.pvc.create=false assumes the named existing claim supports ReadWriteMany and that its access mode must be verified before install or upgrade.

charts/snapshot/templates/role.yaml
Ronkahn21: This is not a published breaking change for this repository. The standalone snapshot repository and chart are new, and the namespace-restricted mode was never released from it, so there is no released chart upgrade contract to document here.

PR #63, opened 2026-08-17
feat: publish checkpoint source facts

leeZardNav:
## Summary

- `PodSnapshotContent` gains an optional `status.source` publishing what a checkpoint was captured on: the node, the container, and the devices.
- The block is projected from the artifact's `manifest.yaml` and written in the same status patch that marks the capture `Ready`, so a content that is Ready already carries its facts.
- Device facts are grouped under one property per vendor, so each payload stays typed and validated and a second vendor can join without a discriminator or a CEL rule.
- Nothing reads the block back. The restore compatibility gates compare the manifest, so this cannot become a second source of truth.

Design: [#99](https://github.com/ai-dynamo/snapshot/pull/99), section 7.1. Stacked on [#140](https://github.com/ai-dynamo/snapshot/pull/140), which records the facts this publishes.

## What it looks like

```yaml
status:
  conditions:
    - type: Ready
      status: "True"
      reason: Captured
  source:
    devices:
      nvidia:
        driverVersion: 580.82.07
        instances:
          - productName: NVIDIA A100-SXM4-80GB
          - productName: NVIDIA A100-SXM4-80GB
    node:
      name: gpu-node-3
      architecture: amd64
      kernelVersion: 5.15.0-1071-aws
    pod:
      image: nvcr.io/nvidia/ai-dynamo/vllm-runtime:0.6.1
      imageDigest: sha256:9f2c...
      memory: 64Gi
```

`instances[].productName` and `driverVersion` are the names the NVIDIA DRA driver publishes on a `ResourceSlice`, so the same GPU reads the same way in both places. The GPU count is the length of the instance list rather than a field of its own, so the two cannot disagree.

## How to read this

The 7 commits form a ladder; each introduces one idea with its own tests.

| Commits | What it does |
| ------- | ------------ |
| 1-3 | The published shape, one block per commit: node, then pod, then vendor-keyed devices. Each regenerates both CRD copies and the deepcopy |
| 4 | The projection from manifest to status as a pure function, reusing `compat.ImageDigest`. Nothing calls it yet |
| 5 | Wires it into the Ready write, at all three call sites |
| 6 | A test per Ready path — fresh dump, artifact recovery, and a work order reconciled after the source pod is gone — plus the conflict retry and the do-not-erase guard |
| 7 | E2E assertion against the node object and nvidia-smi rather than against the manifest |

## Decisions worth a look

**Absent, not empty.** A fact the agent could not read is left out, and a block with nothing in it is omitted entirely. An artifact captured before a fact was ever recorded still publishes; a reader can tell unknown from a value that happens to be blank.

**The published digest is the compared digest.** `compat.ImageDigest` is exported rather than reimplemented, so the reduction that decides whether two images are the same content is the one this publishes. Runtimes wrap a digest differently and the artifact keeps whichever form it was handed, so a second implementation could show a value the gate never compares.

**One read, not one per retry.** The Ready write retries on optimistic-lock conflict, so the manifest is read once before the loop.

**Losing the facts never costs the capture.** By the time this runs the artifact is committed, so an unreadable manifest logs and Ready is written anyway.

**No `node.agentVersion` yet.** The design lists it, but [#140](https://github.com/ai-dynamo/snapshot/pull/140) dropped agent-version recording along with the major/minor comparison, so the field would only ever be empty. It arrives with the deferred agent-version compatibility mapping, which is also what will give it a meaning.

**No new RBAC.** The agent already holds `podsnapshotcontents/status` `update` and `patch`. `PodSnapshot` keeps mirroring only Ready and Failed.

**Single container.** Capture is still one container per content (`PodReference.Containers` is `MinItems=1`/`MaxItems=1`), so one `source` block per content is correct. Multi-container restore does not change that; if multi-container capture arrives, this block is where it grows a key.

## Verification

- Every one of the 7 commits builds, vets and tests on its own (`go vet ./...` and `go test ./...` for the agent under Linux, `go test ./...` for the api module, at each rung).
- Full suites green for api, agent and operator; `make verify-crds` clean, so the chart CRD copy matches the generated one; license headers, fmt, tidy and helm lint clean; govulncheck clean.
- Manual cluster validation pending.

## Summary by CodeRabbit

- **New Features**
  - Pod snapshot content now reports checkpoint provenance, including source node, pod, container image, and captured device details.
  - NVIDIA GPU information includes driver version, product names, and device counts.
  - Vendor-specific device fields are preserved for broader hardware compatibility.
  - Source information is recorded across fresh captures and artifact recovery workflows.

- **Bug Fixes**
  - Snapshot readiness now completes even when provenance data cannot be read, while retaining previously available metadata.
  - Image digest handling is consistently normalized for compatibility checks.

Superseded by [#141](https://github.com/ai-dynamo/snapshot/pull/141) after removing the internal task reference from the head branch.


leeZardNav:
@coderabbitai review

PR #65, opened 2026-08-17
feat: add SnapshotJob v1alpha1 types and CRD

oleg-kushniriov:
### Summary

- Adds the `SnapshotJob` v1alpha1 CRD types: `SnapshotJob`, `SnapshotJobSpec`, `PodSnapshotTemplate`, `SnapshotJobStatus`, `SnapshotJobList`, the four condition types and their reasons, an owner label, and helpers to check completion/failure/terminal state.
- Two CEL validation rules replace what would otherwise need an admission webhook: spec immutability, and `targetContainers` must name a container actually present in the pod template.
- Registers the new types in the scheme, regenerates deepcopy and the CRD manifest, and mirrors it into the chart's CRD directory.
- No reconciler in this PR — the CRD ships inert. Nothing watches or acts on a `SnapshotJob` object yet.

### Scope

- New types, constants, and helpers for the CRD.
- Scheme registration.
- Generated deepcopy and CRD manifest, with the chart copy verified identical to the source copy.
- A named accessor for the new CRD's embedded manifest, for consistency with the existing accessors for the other two CRDs.
- Added `k8s.io/api` as an explicit direct dependency (needed for the embedded pod template type).

**Explicitly out of scope:** reconciler, RBAC, chart templates other than the mirrored CRD.

### Known gap

The two CEL rules are covered by validation-marker inspection only, not by a test that actually creates/updates an invalid object against a real API server and asserts rejection — doing that requires test infrastructure that starts a real (embedded) API server, which isn't set up in this codebase yet. This isn't a new regression — the existing CRD's own immutability rule has the same gap today. Standing up that infrastructure and closing this gap (for both the new and existing rule) is planned for a follow-up PR rather than this one.

### Test plan

- [x] Scheme recognizes the new types; deepcopy is independent, including through the embedded pod template; all three status helpers behave correctly.
- [x] The new CRD's embedded manifest is a valid `CustomResourceDefinition` and is included in the full manifest set.
- [x] `go build`/`go vet`/`go test ./...` clean.
- [x] Lint clean.
- [x] Formatting and license headers clean.
- [x] No drift between the generated CRD and its chart copy.

## Summary by CodeRabbit

* **New Features**
  * Added support for the v1alpha1 SnapshotJob resource, including pod and snapshot templates, deadlines, status tracking, and lifecycle conditions.
  * Added SnapshotJob CRD availability and Kubernetes scheme registration.
  * Added helpers for identifying completed, failed, and terminal SnapshotJobs.
* **Bug Fixes**
  * Improved embedded CRD installation validation.
  * Enforced immutability for PodSnapshot and PodSnapshotContent specifications.
* **Chores**
  * Updated Kubernetes dependencies to newer patch releases.


api/v1alpha1/snapshotjob_types.go
```diff
@@ -0,0 +1,187 @@
...
+
+// SnapshotJob condition reasons.
+const (
+	// Running
+	ReasonPodPending = "PodPending"
+	ReasonPodReady   = "PodReady"
+
+	// Captured
+	ReasonDumpInProgress = "DumpInProgress"
+	ReasonDumpCompleted  = "DumpCompleted"
```
Ronkahn21: Can we change this to Capture
Dump sound bad

api/v1alpha1/snapshotjob_types.go
```diff
@@ -0,0 +1,187 @@
...
+
+	// WrapLaunchJob wraps the target container's entrypoint with
+	// cuda-checkpoint --launch-job. This is a pass-through of Dynamo's existing
+	// mechanism, carried forward so migrating from DynamoCheckpoint does not
+	// regress it — general multi-GPU capture is not supported or validated in
+	// v1alpha1. Requires targetContainer.command to be non-empty — the command
+	// is the entrypoint that gets prepended. The controller fails with
+	// Failed=True if this is true and command is absent.
+	// +optional
+	WrapLaunchJob bool `json:"wrapLaunchJob,omitempty"`
```
Ronkahn21: this should be removed, the user can just add it themself

api/v1alpha1/snapshotjob_types.go
```diff
@@ -0,0 +1,187 @@
...
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope=Namespaced,shortName=snapjob
+// +kubebuilder:printcolumn:name="Running",type="string",JSONPath=".status.conditions[?(@.type=='Running')].status"
+// +kubebuilder:printcolumn:name="Captured",type="string",JSONPath=".status.conditions[?(@.type=='Captured')].status"
+// +kubebuilder:printcolumn:name="Completed",type="string",JSONPath=".status.conditions[?(@.type=='Completed')].status"
+// +kubebuilder:printcolumn:name="PodSnapshot",type="string",JSONPath=".status.podSnapshotName"
+// +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
+// +kubebuilder:validation:XValidation:rule="!has(oldSelf.spec) || self.spec == oldSelf.spec",message="spec is immutable"
+// +kubebuilder:validation:XValidation:rule="self.spec.podSnapshotTemplate.targetContainers.all(c, c in self.spec.podTemplate.spec.containers.map(x, x.name))",message="targetContainers must name containers present in podTemplate"
```
Ronkahn21: can we move this under the relevant structs not here

api/v1alpha1/snapshotjob_types.go
```diff
@@ -0,0 +1,178 @@
...
+
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope=Namespaced,shortName=snapjob
+// +kubebuilder:printcolumn:name="Running",type="string",JSONPath=".status.conditions[?(@.type=='Running')].status"
+// +kubebuilder:printcolumn:name="Captured",type="string",JSONPath=".status.conditions[?(@.type=='Captured')].status"
+// +kubebuilder:printcolumn:name="Completed",type="string",JSONPath=".status.conditions[?(@.type=='Completed')].status"
+// +kubebuilder:printcolumn:name="PodSnapshot",type="string",JSONPath=".status.podSnapshotName"
+// +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
+// +kubebuilder:validation:XValidation:rule="!has(oldSelf.spec) || self.spec == oldSelf.spec",message="spec is immutable"
```
Ronkahn21: can you move this under spec field

oleg-kushniriov: good point. moved

Ronkahn21 (approved):
LGTM

PR #66, opened 2026-08-17
Use PodSnapshot references for restore

Ronkahn21:
Stacked on #61.

## Summary

- make nvidia.com/restore-from the only snapshot-owned restore annotation and reject snapshot annotations on capture pods
- resolve the same-namespace PodSnapshot and bound PodSnapshotContent in the node agent, scoped by spec.nodeName
- key artifacts by PodSnapshotContent UID at artifacts/<content-uid>/containers/<container-name>
- publish restore progress through the Restored pod condition using server-side apply on pods/status
- update snapshotctl, RBAC, documentation, unit tests, and E2E flows for the new contract

## Testing

- go test ./... in api and operator
- full agent go test ./... in Linux
- golangci-lint for api, operator, and Linux agent
- Helm lint and CRD generation/parity
- Python syntax validation for updated E2E files

## Summary by CodeRabbit

* **New Features**
  * Checkpoints require a snapshot name and explicitly selected container.
  * Restores require a manifest and snapshot name, creating a new pod for the captured container.
  * Restore progress is reported through the pod’s `Restored` condition, including success and failure details.
  * Snapshot, artifact identity, readiness, and container selection are validated.
  * Artifact storage now uses content- and container-specific paths.
  * Restore operations recover safely after interruptions and handle cleanup and failures consistently.

* **Documentation**
  * Updated `snapshotctl` usage examples for the revised workflow.


agent/internal/controller/controller.go
```diff
@@ -538,80 +590,237 @@ func (w *NodeController) runRestore(ctx context.Context, pod *corev1.Pod, contai
...
+		return fmt.Errorf("restore failed and placeholder PID could not be resolved: %w", err)
+	}
+	if err := snapshotruntime.SendSignalToPID(op.log, placeholderHostPID, syscall.SIGKILL, "restore failed"); err != nil {
+		return fmt.Errorf("restore failed and placeholder could not be killed: %w", err)
 	}
+	if err := w.convergeRestoredFailureCondition(ctx, op.pod, op.artifact.ContainerName, reason, restoreErr); err != nil {
+		return fmt.Errorf("re-persist failed restore condition after stopping placeholder: %w", err)
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
Ronkahn21: Not changing this behavior in this PR. The current flow persists Restored=False/RestoreFailed before placeholder termination, and the convergence loop referenced by this comment was intentionally removed to avoid changing post-preflight restore failure behavior. The Restored condition is now owned through forced server-side apply.

agent/internal/executor/checkpoint_test.go
```diff
@@ -0,0 +1,60 @@
...
+
+func TestCheckpointRejectsMismatchedArtifactPath(t *testing.T) {
+	cfg := &types.AgentConfig{Storage: types.StorageSpec{BasePath: t.TempDir()}}
+	err := Checkpoint(context.Background(), checkpointPathRuntime{}, logr.Discard(), CheckpointRequest{
+		ContentUID:         "content-uid",
+		ContainerName:      "main",
+		CheckpointLocation: filepath.Join(cfg.Storage.BasePath, "wrong"),
+	}, cfg)
+	require.ErrorContains(t, err, "does not match content artifact path")
+}
```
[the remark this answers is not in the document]
Ronkahn21: This comment is obsolete after the artifact API simplification. CheckpointLocation was removed, and ContentUID/ContainerName validation is handled by the artifact path resolvers with dedicated path tests.

agent/internal/types/manifest.go
```diff
@@ -19,27 +19,38 @@ const manifestFilename = "manifest.yaml"
...
-		K8s:          k8s,
-		Overlay:      overlay,
+		Artifact: ArtifactManifest{
+			ContentUID:    contentUID,
+			ContainerName: containerName,
+		},
+		CreatedAt: time.Now().UTC(),
+		CRIUDump:  criuDump,
+		K8s:       k8s,
+		Overlay:   overlay,
```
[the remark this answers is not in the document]
Ronkahn21: Deferred by design for this PR. Legacy checkpoint-ID artifacts are unsupported; migration, cleanup, and upgrade documentation are intentionally not being added in this change.

operator/internal/protocol/checkpoint.go
```diff
@@ -40,7 +34,7 @@ func NewCheckpointJob(podTemplate *corev1.PodTemplateSpec, opts CheckpointJobOpt
 		podTemplate.Annotations = map[string]string{}
 	}
 	podTemplate.Annotations = DisableCheckpointJobSidecarInjection(podTemplate.Annotations)
-	snapshotv1alpha1.ApplyCheckpointSourceMetadata(podTemplate.Labels, podTemplate.Annotations, opts.CheckpointID, opts.ArtifactVersion)
+	podTemplate.Labels[snapshotv1alpha1.CheckpointSourceLabel] = "true"
```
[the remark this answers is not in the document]
Ronkahn21: Fixed in b813901. NewCheckpointJob now rejects the presence of nvidia.com/restore-from, including an empty value, and regression coverage was added.

agent/internal/controller/controller.go
```diff
@@ -276,264 +323,313 @@ func (w *NodeController) Run(ctx context.Context) error {
...
-	if livePod.DeletionTimestamp != nil ||
-		(livePod.Status.Phase != corev1.PodPending && livePod.Status.Phase != corev1.PodRunning) {
-		w.log.V(1).Info("Skipping restore; pod became ineligible while polling runtime",
-			"pod", podKey,
-			"container", containerName,
-			"phase", livePod.Status.Phase,
-		)
-		return nil, false
+		return nil, newRestorePendingError("SnapshotPending", fmt.Sprintf("Unable to read PodSnapshot %s: %v", key.String(), err))
 	}
```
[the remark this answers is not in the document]
Ronkahn21: Fixed in b813901. Non-NotFound PodSnapshot and PodSnapshotContent read errors are logged internally, while the Restored condition receives stable pending messages without raw API error text. Regression coverage verifies both branches.

agent/internal/controller/controller.go
```diff
@@ -276,264 +323,315 @@ func (w *NodeController) Run(ctx context.Context) error {
...
-	if annotationContainerID == containerID && (annotationStatus == snapshotv1alpha1.RestoreStatusCompleted || annotationStatus == snapshotv1alpha1.RestoreStatusFailed) {
-		return
-	}
-	if w.config.CRIU.TcpEstablished && pod.Status.PodIP == "" {
-		w.log.V(1).Info("Restore pod has no PodIP yet; waiting before TCP-established restore",
-			"pod", podKey,
-			"container", containerName,
-		)
-		return
+	if !podSpecHasContainer(&pod.Spec, containerName) {
```
shayasoolin, line 560: In case the user adds the restore annotation on a pod that has the right container name, here we may start restoring even that the preconditions are not there.
Do we plan to have the planned validations run here, before the restore takes place?
Otherwise, imagine a plain pod - which is not ready for restore in any way - but has the annotation and the container name. It will go through the destructive restore procedure, which may be risky.

agent/internal/controller/controller.go
```diff
@@ -276,264 +323,315 @@ func (w *NodeController) Run(ctx context.Context) error {
...
-		if _, err := snapshotv1alpha1.RestoreStatusAnnotationKeysFor(containerName); err != nil {
-			w.log.Error(
-				err,
-				"Restore target container name cannot be used in restore status annotation key",
-				"pod", podKey,
-				"container", containerName,
-			)
-			return
-		}
+	if err := terminalRestoreConditionError(pod); err != nil {
```
shayasoolin: This check uses the informer-cached pod, which may be stale.
After an attempt writes RestoreFailed, an already queued update may still contain RestoreInProgress and start another restore before the cache receives the failure.
You should check the latest pod state from the API before starting another attempt, or add another guard that prevents replay after a terminal failure.

agent/internal/controller/controller.go
```diff
@@ -276,264 +323,315 @@ func (w *NodeController) Run(ctx context.Context) error {
...
+		key, shutdown := w.restoreQueue.Get()
+		if shutdown {
+			return
+		}
+		go w.processRestoreQueueItem(ctx, key)
 	}
+}
 
-	targets, err := snapshotv1alpha1.TargetContainersFromAnnotations(pod.Annotations, 1, 0)
+func (w *NodeController) processRestoreQueueItem(ctx context.Context, key client.ObjectKey) {
```
oleg-kushniriov, line 393: Once a restore finishes (or fails), the agent keeps posting a new event for that pod every 30 seconds, forever. . Only emit when something actually changes.

agent/internal/controller/controller.go
```diff
@@ -542,96 +640,223 @@ func (w *NodeController) runRestore(ctx context.Context, pod *corev1.Pod, contai
...
+	}
+	return w.restoreFn(ctx, w.runtime, op.log, req, w.injector)
+}
+
+func (op *restoreOperation) failRestore(ctx context.Context, restoreErr error) error {
+	w := op.controller
+	op.log.Error(restoreErr, "External restore failed")
+	reason := restoreFailedReason
+	emitPodEvent(ctx, w.clientset, op.log, op.pod, "snapshot", corev1.EventTypeWarning, reason, restoreErr.Error())
+	if err := op.setRestoredCondition(ctx, corev1.ConditionFalse, reason, restoreErr.Error()); err != nil {
```
oleg-kushniriov: When a restore fails, the agent marks the pod "failed" first and only then tries to kill the placeholder process. If that kill fails, it's never retried - the result is a pod that looks Running, does nothing. Kill first, mark failed after.

agent/internal/nsmount/artifact.go
```diff
@@ -8,27 +8,42 @@ import (
...
 )
 
-// ResolveArtifactPath returns the existing checkpoint artifact layout rooted
-// at the agent-owned base path. All variable components must be single clean
-// path elements.
-func ResolveArtifactPath(basePath, artifactID, version string) (string, error) {
+// ResolveArtifactPath returns the checkpoint artifact path owned by one
+// PodSnapshotContent and captured container. All variable components must be
+// single safe path elements.
+func ResolveArtifactPath(basePath, contentUID, containerName string) (string, error) {
```
oleg-kushniriov, line 21: The on-disk artifact path and manifest format both changed, with no migration and no mention anywhere. After upgrading, every existing snapshot becomes unrestorable and old checkpoint files leak on the PVC. Worth documenting it.

api/v1alpha1/constants.go
```diff
@@ -17,39 +17,19 @@ const (
...
-	RestoreContainerIDAnnotation = "nvidia.com/snapshot-restore-container-id"
+	// RestoreFromAnnotation names the PodSnapshot to restore in the pod's
+	// namespace. Other annotations are ignored by the snapshot protocol.
+	RestoreFromAnnotation = "nvidia.com/restore-from"
 
-	DefaultCheckpointArtifactVersion = "1"
-	DefaultCheckpointJobTTLSeconds   = int32(300) // TODO: dead code — remove once no longer synced from Dynamo
-	DefaultSeccompLocalhostProfile   = "profiles/block-iouring.json"
+	// RestoredCondition is the Pod status condition owned by the node agent.
+	RestoredCondition = "Restored"
```
oleg-kushniriov: use `nvidia.com/Restored` instead, in the future it can collide with k8s-native condition

agent/internal/controller/controller.go
```diff
@@ -152,48 +214,33 @@ func newDefaultController(
...
-		return fmt.Errorf("failed to build restore label selector: %w", err)
-	}
-	restoreSelector := restoreSel.String()
-
 	restoreFactoryOpts := []informers.SharedInformerOption{
-		informers.WithTweakListOptions(tweakNodePodListOptions(restoreSelector, w.config.NodeName)),
+		informers.WithTweakListOptions(tweakNodePodListOptions(w.config.NodeName)),
 	}
 
 	restoreFactory := informers.NewSharedInformerFactoryWithOptions(
```
oleg-kushniriov, line 232: Consider using metadata-only informer .
Once label selector was removed in favor of NodeName selector - we can watch all node pods cheaply

agent/internal/controller/controller.go
```diff
@@ -542,96 +640,223 @@ func (w *NodeController) runRestore(ctx context.Context, pod *corev1.Pod, contai
...
+	}
+	err = op.completeRestore(ctx, placeholderHostPID)
+	return err != nil, err
+}
+
+// recoverCompletedRestore finalizes an interrupted status update without
+// replaying CRIU. RestoreInProgress is normally retryable after an agent
+// restart, but the restore-complete sentinel proves that execution already
+// finished and only the Restored=True status write remains.
+func (op *restoreOperation) recoverCompletedRestore(ctx context.Context) (bool, error) {
```
oleg-kushniriov, line 716: f the agent crashes right after a restore finishes but before recording success, and someone deletes the PodSnapshot meanwhile, the pod is actually restored and working but stays labeled "pending" forever. Run the recovery check before requiring the snapshot to exist.

oleg-kushniriov (commented):
1/n

agent/internal/controller/controller.go
Ronkahn21: its the same behavior has it used to be, I kept it so the changes are smaller
this is not the full migration to workquee once I moved there it will not be needed

agent/internal/nsmount/artifact.go
Ronkahn21: we already taked this is breaking changed

api/v1alpha1/constants.go
Ronkahn21: I will add condition but not nvidia.com pref

agent/internal/controller/podsnapshotcontent.go
```diff
@@ -150,35 +150,27 @@ func (w *NodeController) reconcileSourcePod(ctx context.Context, pod *corev1.Pod
...
-	}
-	leaseName := checkpointLeaseName(id)
-	if errs := validation.IsDNS1123Subdomain(leaseName); len(errs) > 0 {
-		return w.setSnapshotContentFailed(ctx, content, "InvalidCheckpointID",
-			fmt.Errorf("checkpoint ID %q produces invalid Lease name %q: %s", id, leaseName, strings.Join(errs, "; ")))
+	containerName, err := singleTargetContainer(content)
+	if err != nil {
+		return w.setSnapshotContentFailed(ctx, content, "InvalidTargetContainer", err)
 	}
+	artifactKey := contentUID + "/" + containerName
```
oleg-kushniriov, line 162: The old dedup guard was keyed on the checkpoint ID, which survived a PodSnapshot delete/recreate - that was deliberate. Both were removed. 
The new guard and Lease are keyed on the content UID, which a recreation always changes, so nothing stops two dumps of the same container running at once.

agent/internal/nsmount/artifact.go
oleg-kushniriov, line 21: Each PodSnapshotContent now owns a unique artifact directory, and nothing ever deletes it. Under the old scheme a re-dump overwrote the same checkpoint-ID path.
Now every delete/recreate of a snapshot permanently leaks a multi-GB GPU checkpoint on the PVC

We need to delete `artifacts/<content-uid>` when the owning PodSnapshotContent is deleted

agent/internal/controller/controller.go
```diff
@@ -276,264 +323,315 @@ func (w *NodeController) Run(ctx context.Context) error {
...
+}
 
-	go func() {
-		if err := w.runRestore(ctx, pod, containerName, containerID, checkpointID, restoreAttemptKey, startedAt); err != nil {
-			opLog := w.log.WithValues("pod", podKey, "checkpoint_id", checkpointID, "container", containerName)
-			opLog.Error(err, "Restore controller worker failed")
-			emitPodEvent(ctx, w.clientset, opLog, pod, "snapshot", corev1.EventTypeWarning, "RestoreWorkerFailed", err.Error())
+// resolveRestoreContainerID uses the kubelet-published status fast path, then
+// polls the node runtime inline while this Pod's queue key remains in progress.
+func (w *NodeController) resolveRestoreContainerID(ctx context.Context, pod *corev1.Pod, artifact *restoreArtifact, podKey string) (string, bool) {
```
oleg-kushniriov, line 626: The old code re-fetched the pod from the API after resolving the container via the node runtime and skipped the restore if the pod had been deleted or gone terminal during the poll `refreshRestorePodForStart`. That guard was deleted with no replacement - the restore now starts against a pod copy that can be up to 30 seconds stale.

agent/internal/controller/controller.go
```diff
@@ -542,96 +640,223 @@ func (w *NodeController) runRestore(ctx context.Context, pod *corev1.Pod, contai
...
+	emitPodEvent(ctx, w.clientset, w.log, pod, "snapshot", corev1.EventTypeWarning, restoreFailedReason, cause.Error())
+	return err != nil
 }
 
-func (w *NodeController) tryAcquire(podKey string) bool {
+func (w *NodeController) handleRestorePreflightError(ctx context.Context, pod *corev1.Pod, cause error) bool {
+	var ignored *restoreIgnoredError
+	if errors.As(cause, &ignored) {
+		if ignored.reason != "" {
+			emitPodEvent(ctx, w.clientset, w.log, pod, "snapshot", ignored.eventType, ignored.reason, ignored.message)
```
oleg-kushniriov: consider defining a const for "snapshot"

agent/internal/controller/controller.go
```diff
@@ -542,96 +640,223 @@ func (w *NodeController) runRestore(ctx context.Context, pod *corev1.Pod, contai
...
+		w.log.Error(err, "Failed to apply restore failure condition", "pod", podKey)
+		emitPodEvent(ctx, w.clientset, w.log, pod, "snapshot", corev1.EventTypeWarning, restoreStatusUpdateFailedReason, fmt.Sprintf("Failed to record restore failure status: %v", err))
+	}
+	w.log.Error(cause, "Restore request failed", "pod", podKey)
+	emitPodEvent(ctx, w.clientset, w.log, pod, "snapshot", corev1.EventTypeWarning, restoreFailedReason, cause.Error())
+	return err != nil
 }
 
-func (w *NodeController) tryAcquire(podKey string) bool {
+func (w *NodeController) handleRestorePreflightError(ctx context.Context, pod *corev1.Pod, cause error) bool {
```
oleg-kushniriov, line 964: The preflight error handling defaults to terminal error.
Anything that isn't explicitly a pending or ignored error stamps Restored=False/RestoreFailed, which the terminal short-circuit then makes permanent. For example a one-off stat error (EIO, permission flake on the PVC mount) or an unexpected artifact-path error kills the restore pod forever, while the old code just logged and retried on the next event. 

Is it intentional?

operator/internal/controller/podsnapshot_reconciler.go
```diff
@@ -157,12 +157,12 @@ func (sr *PodSnapshotReconciler) captureFromSourcePod(ctx context.Context, snap
...
 
-// verifyContentBacklink errors when a content's backref does not point at this PodSnapshot
-// (namespace/name, and uid when recorded). It is pod-free: the content↔pod relationship is the
+// verifyContentBacklink errors when a content's complete backref does not point at this PodSnapshot
+// (namespace, name, and UID). It is pod-free: the content↔pod relationship is the
 // PodSnapshotContent's own concern, not the PodSnapshot reconciler's.
 func verifyContentBacklink(snap *snapshotv1alpha1.PodSnapshot, content *snapshotv1alpha1.PodSnapshotContent) error {
 	if ref := content.Spec.PodSnapshotRef; ref.Namespace != snap.Namespace || ref.Name != snap.Name ||
-		(ref.UID != "" && ref.UID != snap.UID) {
+		ref.UID == "" || ref.UID != snap.UID {
```
oleg-kushniriov, line 319: Any PodSnapshotContent already stored in etcd without a UID backlink (created by an older operator version) flips its PodSnapshot to `Failed/ContentConflict` after the operator upgrades - and Failed is sticky/terminal, so a previously healthy, restorable snapshot is permanently dead.

agent/internal/controller/controller.go
```diff
@@ -542,96 +640,223 @@ func (w *NodeController) runRestore(ctx context.Context, pod *corev1.Pod, contai
...
+	return err != nil, err
+}
+
+// recoverCompletedRestore finalizes an interrupted status update without
+// replaying CRIU. RestoreInProgress is normally retryable after an agent
+// restart, but the restore-complete sentinel proves that execution already
+// finished and only the Restored=True status write remains.
+func (op *restoreOperation) recoverCompletedRestore(ctx context.Context) (bool, error) {
+	condition := findRestoredCondition(op.pod)
+	if condition == nil || condition.Status != corev1.ConditionFalse || condition.Reason != restoreInProgressReason {
```
oleg-kushniriov, line 718: The crash-recovery path ("restore already finished, only the status write is missing - don't replay CRIU") triggers only when the pod's condition reason is exactly `RestoreInProgress`. But preflight's pending path can overwrite that reason (for example when transient API error reading the PodSnapshot writes SnapshotPending). Once overwritten, the recovery gate never fires - and the next successful pass runs a full CRIU restore over a workload that already restored and is actively running.

As a suggestion 
- Check the sentinel whenever the condition is not Restored=True 
OR
- Stop the pending path from overwriting an InProgress reason.

agent/internal/controller/controller.go
```diff
@@ -276,264 +323,315 @@ func (w *NodeController) Run(ctx context.Context) error {
...
+func (w *NodeController) processRestoreQueueItem(ctx context.Context, key client.ObjectKey) {
+	requeue := false
+	defer func() {
+		w.restoreQueue.Done(key)
+		if requeue && ctx.Err() == nil {
+			w.restoreQueue.AddAfter(key, restoreSafetyRequeueInterval)
+		}
+	}()
+
+	pod, err := w.restorePodLister.Pods(key.Namespace).Get(key.Name)
```
oleg-kushniriov, line 402: Re-check restorePodRequested(pod) and drop the key silently if the `nvidia.com/restore-from` annotation  is gone (user deliberatly removed it)

agent/internal/controller/controller.go
```diff
@@ -276,264 +323,315 @@ func (w *NodeController) Run(ctx context.Context) error {
...
+				"pod", podKey,
+				"container", artifact.ContainerName,
+				"container_id", containerID,
+			)
+			return containerID, false
 		}
-	}()
+
+		select {
+		case <-deadline.C:
```
oleg-kushniriov, line 654: If the container never starts (ImagePullBackOff etc.), the agent retries silently forever - no `Restored` condition, no event - so kubectl describe gives an operator no way to tell "restore is waiting for the container" from "the agent never saw this pod". 

All the other wait states report themselves through a pending reason on the `Restored` condition (`SnapshotPending`, `ArtifactPending`, …) - waiting for the container should do the same, e.g. `Restored=False` with reason `ContainerPending`

agent/internal/controller/controller.go
```diff
@@ -542,96 +640,223 @@ func (w *NodeController) runRestore(ctx context.Context, pod *corev1.Pod, contai
...
+	if err != nil {
+		emitPodEvent(ctx, op.controller.clientset, op.log, op.pod, "snapshot", corev1.EventTypeWarning, restoreStatusUpdateFailedReason, fmt.Sprintf("Failed to record %s restore status: %v", reason, err))
+	}
+	return err
+}
+
+// applyRestoredCondition uses server-side apply against the status subresource.
+// Pod conditions are an associative list keyed by type, so this field manager
+// owns only Restored and does not replace kubelet-owned conditions.
+func (w *NodeController) applyRestoredCondition(ctx context.Context, pod *corev1.Pod, status corev1.ConditionStatus, reason, message string) error {
```
oleg-kushniriov, line 846: in case Agent shutdown during restore - pod will nbe left with RestoreInProgress 
COnsider using `context.WithoutCancel(ctx)` -

agent/internal/controller/controller.go
Ronkahn21: Kept intentionally for this PR. Resync can emit the already-completed/already-failed event again; event aggregation or deduplication is separate work. The important behavior is that terminal or handled Pods return before preflight and never rerun restore.

agent/internal/controller/controller.go
Ronkahn21: Fixed in 0738a41. The failure path now resolves and kills the placeholder first, then records RestoreFailed and marks the Pod UID handled. If killing the placeholder fails, terminal state is not recorded and the full restore remains retryable.

agent/internal/controller/controller.go
Ronkahn21: Addressed in 0738a41 with a process-local handled-restores cache keyed by Pod UID. After success, or after failure cleanup completes, stale queued updates only retry finalizer removal and cannot replay CRIU in the same agent process. The rare agent-restart replay edge is intentionally accepted here.

api/v1alpha1/constants.go
Ronkahn21: Updated in 0738a41 to the vendor-neutral condition type snapshot/Restored. There is no released legacy condition type that needs compatibility handling.

agent/internal/controller/controller.go
Ronkahn21: Addressed in 0738a41. Restore now runs a preflight before container resolution or execution: it validates the Pod phase, referenced PodSnapshot state, bound PodSnapshotContent state and UID backlink, captured container presence, and node-local artifact. Full restore-Pod shaping validation is outside this reference-only PR.

agent/internal/nsmount/artifact.go
Ronkahn21: This is an intentional breaking change for the first standalone release. There are no supported checkpoint-ID artifacts to migrate, so legacy layout migration and cleanup are not part of this restore-by-reference PR.

agent/internal/controller/controller.go
Ronkahn21: Not changed in this PR. Sentinel recovery requires resolving the target container/PID and reconstructing the artifact identity from the referenced snapshot/content. Recovery without those objects needs a separate durable identity design; this rare deletion-during-agent-restart case is accepted here.

agent/internal/controller/controller.go
Ronkahn21: Fixed in 0738a41 with the shared snapshotEventComponent constant.

agent/internal/controller/podsnapshotcontent.go
Ronkahn21: Deferred as capture-lifecycle work. This PR intentionally uses PodSnapshotContent UID as artifact/lease identity; delete/recreate concurrency needs an owner/finalizer design and is outside the restore-by-reference scope.

agent/internal/controller/controller.go
Ronkahn21: Deliberate. A live refetch only moves the same delete/status race window. In this PR the finalizer is added immediately after basic filtering, workqueue serialization prevents concurrent processing of the same key, and the handled UID cache prevents same-agent replay. Restore-from immutability is planned separately; the remaining replacement race is accepted here.

agent/internal/nsmount/artifact.go
Ronkahn21: Agree that content-owned artifact garbage collection needs an explicit owner lifecycle. It is separate from restore-by-reference and should be implemented with deletion/finalizer semantics rather than inferred path cleanup, so no cleanup behavior is added in this PR.

agent/internal/controller/controller.go
Ronkahn21: Not taking this optimization in this PR. Restore reconciliation needs annotations, nodeName, container names, Pod phase/conditions/IP, and running container IDs. The informer is already server-side node-scoped; a trimming transform can be considered separately.

agent/internal/controller/controller.go
Ronkahn21: Addressed in 0738a41. os.PathError failures from artifact inspection, including transient stat permission/I/O errors, are logged and exposed as ArtifactPending. Transient Kubernetes reads similarly return stable SnapshotPending/ContentPending messages; only semantic validation failures remain terminal.

agent/internal/controller/controller.go
Ronkahn21: Not adding another live/cached recheck in this PR. restore-from will be made immutable in the follow-up validation work; this controller treats the request as immutable once queued. An extra check here would only move the same annotation race window.

agent/internal/controller/controller.go
Ronkahn21: Fixed in 0738a41. Pending dependency handling now preserves RestoreInProgress and emits the pending event without overwriting the condition, so the completion-sentinel recovery gate remains usable. Coverage was added for this behavior.

operator/internal/controller/podsnapshot_reconciler.go
Ronkahn21: This is intentionally breaking for the first standalone release. Older PodSnapshotContent objects without a UID backlink are not a supported compatibility surface and are treated as stale conflicts; no migration is added in this PR.

agent/internal/controller/controller.go
Ronkahn21: Not using context.WithoutCancel. Agent shutdown should cancel node/runtime work; continuing detached restore work risks overlap with the replacement agent. The Pod remains RestoreInProgress behind the restore finalizer, and the next agent retries or recovery-handles it.

agent/internal/controller/controller.go
Ronkahn21: Fixed in 0738a41. Runtime resolution timeout now records Restored=False with reason ContainerPending, emits an event, and requeues.

Ronkahn21: sddress, it will lead to an error if failed

api/v1alpha1/constants.go
```diff
@@ -17,39 +17,19 @@ const (
...
-	RestoreTargetLabel = "nvidia.com/snapshot-is-restore-target"
-
-	CheckpointArtifactVersionAnnotation = "nvidia.com/snapshot-artifact-version"
-
 	// SnapshotNodeLabel mirrors PodSnapshotContent.spec.source.nodeName onto the
 	// object so the per-node agent's cache can label-select work for its node.
 	SnapshotNodeLabel = "nvidia.com/snapshot-node"
 
-	// Required comma-separated checkpoint/restore target container list.
-	TargetContainersAnnotation = "nvidia.com/snapshot-target-containers"
```
julienmancuso, line 31: this MR removes `snapshot-target-containers` annotation, derives exactly one target container from the captured PodSnapshot and removes the current one-to-many restore path.
That conflicts with Dynamo’s intra-pod failover behavior. let's not remove this until there is a first-class replacement

api/v1alpha1/constants.go
```diff
@@ -17,39 +17,19 @@ const (
...
-	RestoreContainerIDAnnotation = "nvidia.com/snapshot-restore-container-id"
+	// RestoreFromAnnotation names the PodSnapshot to restore in the pod's
+	// namespace. Other annotations are ignored by the snapshot protocol.
+	RestoreFromAnnotation = "nvidia.com/restore-from"
 
-	DefaultCheckpointArtifactVersion = "1"
-	DefaultCheckpointJobTTLSeconds   = int32(300) // TODO: dead code — remove once no longer synced from Dynamo
-	DefaultSeccompLocalhostProfile   = "profiles/block-iouring.json"
+	// RestoredCondition is the Pod status condition owned by the node agent.
+	RestoredCondition = "snapshot/Restored"
```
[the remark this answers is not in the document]
Ronkahn21: No compatibility layer is being added. This is the first standalone release, so the old bare Restored condition and the removed checkpoint-ID/restore-annotation API are not a supported upgrade or external-consumer contract. The break is intentional and there are no released legacy restore Pods to migrate.

agent/internal/controller/controller.go
```diff
@@ -542,96 +691,316 @@ func (w *NodeController) runRestore(ctx context.Context, pod *corev1.Pod, contai
...
+	finalizerErr := w.removeRestoreFinalizer(ctx, pod)
+	if finalizerErr != nil {
+		w.log.Error(finalizerErr, "Failed to remove restore protection finalizer", "pod", fmt.Sprintf("%s/%s", pod.Namespace, pod.Name))
+		emitPodEvent(ctx, w.clientset, w.log, pod, snapshotEventComponent, corev1.EventTypeWarning, restoreFinalizerUpdateFailedReason, finalizerErr.Error())
+	}
+	if statusErr != nil {
+		return statusErr
+	}
+	return finalizerErr
+}
```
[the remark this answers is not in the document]
Ronkahn21: Fixed in cb71850 with the simpler ordering discussed here: the terminal condition must apply successfully before the Pod UID is cached as handled or the restore finalizer is removed. A failed status apply leaves the finalizer in place and requeues. The success-path regression proves the retry uses the existing completion sentinel, persists RestoreSucceeded, removes the finalizer, and does not replay CRIU.

agent/internal/controller/controller.go
```diff
@@ -542,96 +691,316 @@ func (w *NodeController) runRestore(ctx context.Context, pod *corev1.Pod, contai
...
-	emitPodEvent(ctx, w.clientset, log, pod, "snapshot", corev1.EventTypeNormal, "RestoreSucceeded", fmt.Sprintf("Restore completed from checkpoint %s", checkpointID))
-	return setRestoreStatus(snapshotv1alpha1.RestoreStatusCompleted)
+func (w *NodeController) patchRestoreFinalizers(ctx context.Context, pod *corev1.Pod, finalizers []string) error {
+	patch, err := json.Marshal(map[string]any{"metadata": map[string]any{"finalizers": finalizers}})
+	if err != nil {
+		return err
+	}
+	_, err = w.clientset.CoreV1().Pods(pod.Namespace).Patch(ctx, pod.Name, ktypes.MergePatchType, patch, metav1.PatchOptions{})
+	return err
+}
```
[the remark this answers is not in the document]
Ronkahn21: Kept intentionally for this PR. Including the informer object's resourceVersion would make finalizer patches contend with frequent kubelet status updates. Finalizer patch failures already requeue, this controller is the single node-scoped restore writer, and we are explicitly accepting the narrow concurrent-finalizer race instead of adding a live-GET/conflict loop here.

Ronkahn21:
Regarding CodeRabbit's outside-diff SnapshotJob adoption finding in the latest review: the concern is valid, but it is pre-existing behavior from PR #80 rather than part of the restore-by-reference flow. Correctly hardening it requires defining a normalized capture-critical Pod template identity and admission/defaulting behavior. We are deferring that work to a dedicated SnapshotJob hardening PR so #66 remains restore-focused.

agent/internal/controller/podsnapshotcontent.go
oleg-kushniriov: lets create an issue or task to follow up that this is fixed in future PRs

agent/internal/nsmount/artifact.go
oleg-kushniriov: Did you documented those breaking changes?

oleg-kushniriov (approved):
Approving!

a few conditions before or after merge

Breaking changes need to be documented:

- Artifact path/manifest format change - existing snapshots become unrestorable, old checkpoint files are left on the PVC.
- PodSnapshotContent objects without a UID backlink (created by older operator versions) will flip to Failed/ContentConflict after upgrade.
- Both are reasonable calls for a pre-GA release, but neither is captured anywhere outside this PR thread. Please add a line to the release notes / upgrade doc for each before we ship, so this doesn't become a support surprise later.

Follow-up tasks needed:

Two items were deferred as out of scope here but are real gaps, not just cleanup debt - please open tracking issues for both so they don't get lost:

- Artifact directories (artifacts/<content-uid>/...) are never garbage collected on PodSnapshotContent deletion - unbounded PVC growth with no current visibility. A stopgap log/metric on orphaned artifact dirs would help until the real owner/finalizer-based GC lands.
- The dedup guard/Lease is now keyed on PodSnapshotContent UID, which changes on delete/recreate - nothing currently prevents two concurrent CRIU dumps against the same container in that window.

api/v1alpha1/constants.go
Ronkahn21: will be address in new PR

PR #68, opened 2026-08-18
Snapshotjob source job

oleg-kushniriov:
### Summary

- Adds the Job phase of the `SnapshotJob` reconciler: `buildSourceJob` builds the source `batch/v1 Job` via the existing `protocol.NewCheckpointJob` (unchanged — it's the agent contract), and `SnapshotJobReconciler` creates it with a controller ownerRef, derives `Running` from `job.status.ready`, and handles crash-recovery (adopts an already-owned Job, including the `Create`-time `AlreadyExists` race) and classification of a foreign object holding the deterministic name.
- Spec validation is terminal, not retried: an empty/multi-entry `targetContainers` list or a target container missing from `podTemplate` is recorded as `Failed=True/InvalidSpec` with an event, rather than retried forever against a spec that's CEL-immutable. Adds `ReasonInvalidSpec` — none of the seven previously documented `Failed=True` reasons covered a spec-validation failure; also documented as an eighth reason in the design docs, which previously described a closed set of seven.
- No storage is injected (the agent's own config owns it) and `WrapLaunchJob` is always `false` (`PodSnapshotTemplate` carries no field for it).

### Scope

- `operator/internal/controller/snapshotjob_job.go` — `buildSourceJob`.
- `operator/internal/controller/snapshotjob_reconciler.go` — `SnapshotJobReconciler`: the Job phase of `Reconcile` (create/adopt/classify), `SetupWithManager` with `For`/`Owns(batchv1.Job)`/`MaxConcurrentReconciles: 1`.
- `api/v1alpha1/snapshotjob_types.go` — adds `ReasonInvalidSpec`.
- Matching test files for both.

**Explicitly out of scope:** PodSnapshot creation, the completion gate, cleanup, and — deliberately — registration in `cmd/manager/main.go`. The reconciler exists and is fully tested but does not run in production yet; wiring it in is a later PR, so a half-finished controller can never create a Job it can't see through to completion.

### Test plan

- [x] `buildSourceJob` — options wiring (identity, target container, deadline, backoff, seccomp), owner label without clobbering existing labels, no storage injected, `WrapLaunchJob` always `false`, terminal errors on empty/multi-entry `targetContainers`, target container absent from `podTemplate`, and empty `podTemplate.containers`.
- [x] `SnapshotJobReconciler.Reconcile` — first-pass Job creation with ownerRef and `Running=False/PodPending`; `Running` transitions to `True/PodReady` on `job.status.ready > 0` with `startedAt` set exactly once; adopts an already-owned Job without recreating it; the `Create`-time `AlreadyExists` race is handled for both an owned and a foreign Job; a foreign object at the Job name is classified without being acted on; a non-`NotFound` `Get` error is returned (retryable) rather than swallowed; terminal and deleted `SnapshotJob`s are skipped; `SnapshotJob` not found is a no-op.
- [x] `golangci-lint` clean, `gofmt` clean, license headers pass.

## Summary by CodeRabbit

* **New Features**
  * SnapshotJobs now create and monitor source workloads, report progress, preserve start times, and handle completion or failure states.
  * Added validation, ownership tracking, deadlines, security settings, and terminal status handling.

* **Bug Fixes**
  * Checkpoint IDs containing dots are now accepted when otherwise valid.
  * Improved handling of invalid specifications, conflicting workloads, and reconciliation races.
  * Improved recovery when workloads are created concurrently or already exist.

* **Documentation**
  * Clarified that `startedAt` reflects when the controller observes the source workload as ready.


agent/internal/controller/podsnapshotcontent.go
```diff
@@ -153,9 +153,13 @@ func (w *NodeController) reconcileSourcePod(ctx context.Context, pod *corev1.Pod
 		return w.setSnapshotContentFailed(ctx, content, "MissingCheckpointID",
 			fmt.Errorf("source pod %q missing %s label", pod.Name, snapshotv1alpha1.CheckpointIDLabel))
 	}
-	if errs := validation.IsDNS1123Label(id); len(errs) > 0 {
+	// The checkpoint ID is a label value (CheckpointIDLabel), not a DNS-1123 label —
+	// dots are valid in a label value, so a stricter DNS-1123-label check would
+	// wrongly reject a valid SnapshotJob name like "warm.worker" that made it this
+	// far as a pod label.
```
Ronkahn21: remove

api/v1alpha1/snapshotjob_types.go
```diff
@@ -47,6 +47,12 @@ const (
 	ReasonPodSnapshotNameConflict = "PodSnapshotNameConflict"
 	ReasonJobNameConflict         = "JobNameConflict"
 	ReasonJobDeleted              = "JobDeleted"
+	// ReasonInvalidSpec is a defense-in-depth reason for spec-level validation
+	// failures caught by the controller before building the source Job (e.g. no
+	// container named by spec.podSnapshotTemplate.targetContainers[0]). The CEL
+	// XValidation rule on SnapshotJobSpec already rejects most of these at
+	// admission time; this reason only fires for objects that bypassed it.
```
Ronkahn21: this could be ignore no need to do defence in depth here

operator/internal/controller/snapshotjob_job.go
```diff
@@ -0,0 +1,67 @@
...
+// up themselves in spec.podTemplate).
+func buildSourceJob(sj *snapshotv1alpha1.SnapshotJob) (*batchv1.Job, error) {
+	// sj.Name becomes a label value (SnapshotJobOwnerLabel, CheckpointIDLabel) as
+	// well as the source Job's own name. Kubernetes object names allow dots and up
+	// to 253 characters (RFC 1123 subdomain), but label values are capped at 63
+	// chars with a looser charset (RFC 1123 label plus '_' and '.') — the CRD does
+	// not constrain metadata.name, so this must be checked here. A stricter
+	// DNS-1123-label check would wrongly reject a valid dotted name like
+	// "warm.worker"; without any check, a long-named SnapshotJob would fail Job
+	// creation with an apiserver validation error on every retry, forever.
```
Ronkahn21: can be reduced

operator/internal/controller/snapshotjob_job.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+	podTemplate.Labels[snapshotv1alpha1.SnapshotJobOwnerLabel] = sj.Name
+
+	return protocol.NewCheckpointJob(podTemplate, protocol.CheckpointJobOptions{
+		Namespace:             sj.Namespace,
+		Name:                  sj.Name,
+		TargetContainer:       targetContainers[0],
+		CheckpointID:          sj.Name,
+		ArtifactVersion:       sourceJobArtifactVersion,
+		SeccompProfile:        snapshotv1alpha1.DefaultSeccompLocalhostProfile,
```
Ronkahn21, line 60: this is opinionated profile I am not sure we want to keep it like this hidden in the code

operator/internal/controller/snapshotjob_job.go
```diff
@@ -0,0 +1,67 @@
...
+		ActiveDeadlineSeconds: sj.Spec.ActiveDeadlineSeconds,
+		TTLSecondsAfterFinish: nil,
+		WrapLaunchJob:         false,
```
Ronkahn21, line 63: also this

operator/internal/controller/snapshotjob_reconciler.go
```diff
@@ -0,0 +1,218 @@
...
+// Reconcile drives a SnapshotJob's source Job into existence and derives Running
+// from it. It is a thin orchestrator: each branch delegates to a helper that owns
+// that path's detail.
+func (r *SnapshotJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	sj := &snapshotv1alpha1.SnapshotJob{}
+	if err := r.Get(ctx, req.NamespacedName, sj); err != nil {
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	// No finalizer: the batch/v1 Job's ownerReference cascade reaps it when the
```
Ronkahn21: redundant

operator/internal/controller/snapshotjob_reconciler.go
```diff
@@ -0,0 +1,218 @@
...
+	// SnapshotJob is deleted, and the PodSnapshot (added in PR 4) must survive
+	// deletion anyway, so there is nothing to clean up synchronously here.
+	if !sj.GetDeletionTimestamp().IsZero() {
+		return ctrl.Result{}, nil
+	}
+	if snapshotv1alpha1.IsSnapshotJobTerminal(sj) {
+		return ctrl.Result{}, nil
```
Ronkahn21, line 61: if it terminal but you didn't update the status the CR will be hanged without proper status

operator/internal/controller/snapshotjob_job.go
```diff
@@ -0,0 +1,67 @@
...
+	if podTemplate.Labels == nil {
+		podTemplate.Labels = map[string]string{}
+	}
+	podTemplate.Labels[snapshotv1alpha1.SnapshotJobOwnerLabel] = sj.Name
+
+	return protocol.NewCheckpointJob(podTemplate, protocol.CheckpointJobOptions{
+		Namespace:             sj.Namespace,
+		Name:                  sj.Name,
+		TargetContainer:       targetContainers[0],
+		CheckpointID:          sj.Name,
```
Ronkahn21, line 58: Using only `sj.Name` makes the checkpoint artifact identity collide across namespaces and across delete/recreate cycles. The agent resolves this to `<base>/<checkpointID>/versions/1` and treats an existing artifact directory as successful crash recovery, so a new `SnapshotJob` can reuse stale data; cross-namespace captures also use different namespaced Leases and can write the same path concurrently. Please derive a label-safe ID from the object instance (for example, its UID or a hash including namespace/name/UID) and persist or expose that ID for restore.

operator/internal/controller/snapshotjob_reconciler.go
```diff
@@ -0,0 +1,218 @@
...
+// since 1.24 behind the JobReadyPods feature gate) — the controller watches only
+// the Job, not the pod, per the design's "SnapshotJob observes the Job, not the
+// Pod, for failure status." startedAt is recorded once, the first time the pod is
+// observed ready, and never rewritten afterward.
+func (r *SnapshotJobReconciler) observeJob(ctx context.Context, sj *snapshotv1alpha1.SnapshotJob, job *batchv1.Job) (ctrl.Result, error) {
+	ready := job.Status.Ready != nil && *job.Status.Ready > 0
+
+	var changed bool
+	if ready {
+		if sj.Status.StartedAt == nil {
```
Ronkahn21, line 142: `StartedAt` is mutated here without marking the status as changed. If `Running=True/PodReady` is already persisted while `StartedAt` is nil, `setCondition` returns false and the early return below discards this timestamp on every reconcile. Mark `changed = true` when assigning `StartedAt` (and OR it with the condition result), and add a regression test with a preloaded ready condition and nil `StartedAt`.

Ronkahn21 (commented):
Two non-blocking correctness findings from the review.

operator/internal/controller/snapshotjob_reconciler.go
```diff
@@ -0,0 +1,218 @@
...
+// name: it emits a Warning event so the conflict is visible to an operator (e.g.
+// via `kubectl describe`) without writing to sj.Status. The terminal
+// Failed=True/JobNameConflict status write is added by the completion gate (a
+// later phase), so this reconciler stops without taking an action it can't yet
+// finish recording — the event is a cheap interim signal, not a substitute.
+func (r *SnapshotJobReconciler) reportJobNameConflict(sj *snapshotv1alpha1.SnapshotJob) (ctrl.Result, error) {
+	r.Recorder.Event(sj, corev1.EventTypeWarning, snapshotv1alpha1.ReasonJobNameConflict,
+		fmt.Sprintf("an object not controlled by this SnapshotJob already holds the name %q", sj.Name))
+	return ctrl.Result{}, nil
+}
```
Ronkahn21: do we want to failed the snapshotJob

operator/internal/controller/snapshotjob_reconciler.go
```diff
@@ -0,0 +1,218 @@
...
+// completedAt. This phase only ever reaches it via InvalidSpec — the completion
+// gate (a later phase) owns the rest of the Failed=True reason matrix, but
+// completedAt must be set here regardless: IsSnapshotJobTerminal short-circuits
+// every later reconcile once Failed=True is persisted, so this is the only
+// chance to record it for an InvalidSpec failure.
+func (r *SnapshotJobReconciler) failSnapshotJob(ctx context.Context, sj *snapshotv1alpha1.SnapshotJob, reason string, cause error) (ctrl.Result, error) {
+	r.Recorder.Event(sj, corev1.EventTypeWarning, reason, cause.Error())
+	setCondition(sj, snapshotv1alpha1.SnapshotJobConditionFailed, metav1.ConditionTrue, reason, cause.Error())
+	if sj.Status.CompletedAt == nil {
+		now := metav1.Now()
```
Ronkahn21: you need to make sure all the other conditions are set accordingly

operator/internal/controller/snapshotjob_reconciler.go
```diff
@@ -0,0 +1,218 @@
...
+	})
+}
+
+// SetupWithManager wires the controller: it owns the batch/v1 Job it creates (the
+// Job carries a controller ownerRef, so this is the built-in mapping; Create is
+// filtered since we just made it). PodSnapshot watching is added in a later phase.
+func (r *SnapshotJobReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&snapshotv1alpha1.SnapshotJob{}).
+		Owns(&batchv1.Job{}, builder.WithPredicates(predicate.Funcs{
```
Ronkahn21, line 213: where is the watch for podSnapshot ?

oleg-kushniriov: Out of scope for this PR by design: 
the PR description states explicitly that PodSnapshot creation and the Watches(&PodSnapshot{}) registration in follow up PR

operator/internal/controller/snapshotjob_job.go
oleg-kushniriov: WrapLaunchJob is always false because PodSnapshotTemplate has no field to source it from

operator/internal/controller/snapshotjob_job.go
oleg-kushniriov: seccomp is a compiled constant (no new values.yaml knob), matching existing snapshotctl precedent.

operator/internal/controller/snapshotjob_job.go
oleg-kushniriov: Agree.
But this is not SnapshotJob introduced/related to fix in those PRs.
I will create a separate task

operator/internal/controller/snapshotjob_reconciler.go
oleg-kushniriov: This logic is not full yet and will be updated in follow-up PRs.  
It doesn't yet account for `Completed`, since that condition doesn't exist until the completion-gate PR.

operator/internal/controller/snapshotjob_job.go
```diff
@@ -0,0 +1,62 @@
...
+//
+// No storage is injected (spec §5.3: the agent falls back to its own config), and
+// WrapLaunchJob is always false (spec §5.4: PodSnapshotTemplate has no field to
+// source it from — a caller needing cuda-checkpoint --launch-job wrapping sets it
+// up themselves in spec.podTemplate).
+func buildSourceJob(sj *snapshotv1alpha1.SnapshotJob) (*batchv1.Job, error) {
+	// sj.Name is also used as a label value (SnapshotJobOwnerLabel,
+	// CheckpointIDLabel); the CRD doesn't cap metadata.name length, so this must
+	// be checked here or a long name fails Job creation forever.
+	if errs := contentvalidation.IsLabelValue(sj.Name); len(errs) > 0 {
```
Ronkahn21, line 39: IsLabelValue does not explain what it does

operator/internal/controller/snapshotjob_reconciler.go
```diff
@@ -0,0 +1,215 @@
...
+//
+// This phase (source Job creation + Running) only handles the batch/v1 Job it
+// creates: build/create it, adopt an existing one it already owns, classify a
+// foreign object holding its deterministic name, and derive Running from
+// job.status.ready. PodSnapshot creation (Captured), the completion gate
+// (Completed/Failed beyond spec validation), and cleanup are added in later
+// phases — this reconciler is not registered in main.go yet, so none of this
+// runs in production until the feature is complete.
+type SnapshotJobReconciler struct {
+	client.Client
```
Ronkahn21, line 43: please not embedded the client it in the struct

operator/internal/controller/snapshotjob_reconciler.go
Ronkahn21: still not cover

operator/internal/controller/snapshotjob_reconciler.go
```diff
@@ -0,0 +1,215 @@
...
+// readiness, so a consumer must not see Failed=True with Running entirely
+// absent from status.conditions (missing is not the same as "known False").
+// An already-set Running is left untouched, since it may already reflect a
+// real observation.
+func (r *SnapshotJobReconciler) failSnapshotJob(ctx context.Context, sj *snapshotv1alpha1.SnapshotJob, reason string, cause error) (ctrl.Result, error) {
+	r.Recorder.Event(sj, corev1.EventTypeWarning, reason, cause.Error())
+	if meta.FindStatusCondition(sj.Status.Conditions, snapshotv1alpha1.SnapshotJobConditionRunning) == nil {
+		setCondition(sj, snapshotv1alpha1.SnapshotJobConditionRunning, metav1.ConditionFalse,
+			snapshotv1alpha1.ReasonPodPending, "waiting for the source pod to become ready")
+	}
```
Ronkahn21, line 177: reason and the msg is not true if the snapshot failed

operator/internal/controller/snapshotjob_reconciler.go
```diff
@@ -0,0 +1,215 @@
...
+	setCondition(sj, snapshotv1alpha1.SnapshotJobConditionFailed, metav1.ConditionTrue, reason, cause.Error())
+	if sj.Status.CompletedAt == nil {
+		now := metav1.Now()
+		sj.Status.CompletedAt = &now
+	}
+	if err := r.Status().Update(ctx, sj); err != nil {
```
Ronkahn21, line 191: you should set all the conditions to reflect the failed

operator/internal/controller/snapshotjob_reconciler.go
oleg-kushniriov: th PodSnapshotReconciler already uses the same embedded-client.Client shape, changing only this reconciler would create inconsistency, so I'm proposing it as a separate repo-wide follow-up instead.

operator/internal/controller/snapshotjob_reconciler.go
```diff
@@ -0,0 +1,221 @@
...
+type SnapshotJobReconciler struct {
+	client.Client
+	Recorder record.EventRecorder
+}
+
+// +kubebuilder:rbac:groups=nvidia.com,resources=snapshotjobs,verbs=get;list;watch;update;patch
+// +kubebuilder:rbac:groups=nvidia.com,resources=snapshotjobs/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=nvidia.com,resources=snapshotjobs/finalizers,verbs=update
+// +kubebuilder:rbac:groups=batch,resources=jobs,verbs=create;get;list;watch
+// +kubebuilder:rbac:groups=core,resources=events,verbs=create;patch
```
Ronkahn21: can you move it above the stract

Ronkahn21 (approved):

PR #69, opened 2026-08-18
Delete snapshot artifacts during finalization

Ronkahn21:
## Summary

- mount the shared checkpoint PVC in the operator and finalize `PodSnapshotContent` artifacts before cascade deletion completes
- coordinate cleanup with the agent capture Lease and prevent deleting contents from starting a new capture
- remove the complete content-UID root, including abandoned staging data, while retaining finalizers on cleanup failure
- add controller unit coverage and a CPU E2E lifecycle test that verifies the PVC artifact root is gone

## Validation

- `go test ./...` in `api` and `operator`
- `go test ./...` in `agent` inside a Linux Go container
- `golangci-lint` for all Go modules, with the agent linted for `GOOS=linux GOARCH=amd64`
- `helm lint ./charts/snapshot` and OpenShift/default template rendering
- E2E pytest collection


Ronkahn21:
Superseded by the approved PodSnapshotContent artifact-cleanup design. Closing without merge; a clean replacement implementation will follow from current main.

PR #70, opened 2026-08-18
create the SnapshotJob PodSnapshot and derive Captured

oleg-kushniriov:
### Summary

- Adds the PodSnapshot phase of the `SnapshotJob` reconciler: `findOwnedPodSnapshot` (List by `SnapshotJobOwnerLabel`), `buildPodSnapshot` (**no ownerReference** — the artifact-survival guarantee; pins the source pod's name and UID; carries `targetContainers` into `spec.source.podRef.containers`), `createPodSnapshot`, and `classifyExistingPodSnapshot` — ported from Dynamo's `DynamoCheckpoint` controller, adapted for label-based ownership (no ownerRef exists to check `IsControlledBy` against).
- The reconciler creates the PodSnapshot as soon as the source pod exists (scheduled or not — `PodSnapshotReconciler`, unchanged, owns waiting for scheduling), records `status.podSnapshotName`, and sets `Captured=False/CaptureInProgress`.
- Mirrors the PodSnapshot's `Ready`/`Failed` into `Captured`. A capture failure (`PodSnapshot Failed=True`) is immediately terminal — `Failed=True/CaptureFailed` with `completedAt` set — since it's fully determined by the PodSnapshot's own signal, unlike a Job failure after a successful capture, which needs the completion gate (a later PR) to track Job state.
- Adds `Watches(&PodSnapshot{})` with a tombstone-aware, label-based map function (mirroring this package's existing `PodSnapshotContent` map function), Update+Delete only.
- Restructures the Job phase from the prior PR to route into this new PodSnapshot phase instead of a standalone Running-only observe step: both the initial Get-owned path and the `Create` `AlreadyExists` adopt path now share one `observe()` that sets `Running` and `Captured` together, matching the combined reconcile-flow design.

### Scope

- `operator/internal/controller/snapshotjob_podsnapshot.go` (new) — `findSourcePod`, `findOwnedPodSnapshot`, `buildPodSnapshot`, `createPodSnapshot`, `classifyExistingPodSnapshot`, `mapPodSnapshotToSnapshotJob`.
- `operator/internal/controller/snapshotjob_reconciler.go` — `reconcilePodSnapshot`, `createPodSnapshotPhase`, `observe` (replaces the old `observeJob`); `SetupWithManager` gets the `PodSnapshot` watch; RBAC markers extended (`podsnapshots` create/get/list/watch, `core/pods` get/list/watch).
- Matching test files for both.

**Explicitly out of scope:** the completion gate (`Completed`, and `Failed` reasons beyond spec validation/capture failure), cleanup, and — deliberately — registration in `cmd/manager/main.go`.

### Test plan

- [x] `buildPodSnapshot` — no ownerReference (named test, the artifact-survival guarantee), owner label, pod name/UID pinned, `targetContainers` carried through, name matches the SnapshotJob's own name, terminal error on empty `targetContainers`.
- [x] PodSnapshot created for an unscheduled pod (no `spec.nodeName`); `status.podSnapshotName` recorded; `Running=False/PodPending` set (not created) when the source pod doesn't exist yet.
- [x] `Captured=True/CaptureCompleted` on `PodSnapshot Ready=True`; `Captured=False/CaptureFailed` + `Failed=True/CaptureFailed` + `completedAt` on `PodSnapshot Failed=True`.
- [x] `classifyExistingPodSnapshot` — ours (adopted), foreign (`PodSnapshotNameConflict`), cache-lag (`NotFound` surfaces the original `AlreadyExists` for requeue).
- [x] `findSourcePod` — not found when no pod exists; finds the Job-owned pod while ignoring a same-labeled foreign pod.
- [x] `mapPodSnapshotToSnapshotJob` — maps via the owner label; no label maps to nothing; malformed object maps to nothing; the tombstone-unwrap helper is tested directly for the delete-event path.
- [x] Full existing PR3 suite re-verified with no regressions after the Job-phase restructure (first-pass creation, `AlreadyExists` race adopt/classify, foreign-Job classification, `Running` transitions, terminal/deleted skip, `Get` error propagation).
- [x] `golangci-lint` clean, `gofmt` clean, license headers pass, `make verify-crds` clean (no schema change).

## Summary by CodeRabbit

- **New Features**
  - Improved SnapshotJob tracking and reconciliation for source Pods and PodSnapshots.
  - Snapshot progress and capture results are recovered and reflected more reliably, including after conflicts or delayed resource visibility.
  - Added stronger resource ownership tracking to prevent collisions between SnapshotJob instances.

- **Bug Fixes**
  - Improved handling of missing, unscheduled, or ambiguous source Pods.
  - Invalid checkpoint IDs that generate unsupported Lease names now fail with a clear error.
  - Added validation for supported security and naming requirements.
  - Improved handling of long and dotted SnapshotJob names.


operator/internal/controller/snapshotjob_podsnapshot.go
```diff
@@ -0,0 +1,189 @@
...
+		client.MatchingLabels{batchv1.JobNameLabel: job.Name},
+	); err != nil {
+		return nil, err
+	}
+	for i := range pods.Items {
+		if metav1.IsControlledBy(&pods.Items[i], job) {
+			return &pods.Items[i], nil
+		}
+	}
+	return nil, apierrors.NewNotFound(corev1.Resource("pods"), job.Name)
```
Ronkahn21: lets check the pods that controll by is 1

operator/internal/controller/snapshotjob_podsnapshot.go
```diff
@@ -0,0 +1,189 @@
...
+			schema.GroupResource{Group: snapshotv1alpha1.GroupVersion.Group, Resource: "podsnapshots"},
+			sj.Name,
+		)
+	case 1:
+		return &snaps.Items[0], nil
+	default:
+		err := fmt.Errorf("multiple PodSnapshots owned by SnapshotJob %q; expected at most one", sj.Name)
+		r.Recorder.Event(sj, corev1.EventTypeWarning, "PodSnapshotLookupAmbiguous", err.Error())
+		return nil, err
+	}
```
Ronkahn21: can we simplify it with just if elfse

operator/internal/controller/snapshotjob_job_test.go
```diff
@@ -141,9 +141,9 @@ func requireContainer(t *testing.T, containers []corev1.Container, name string)
 	return nil
 }
 
-// batchJobByName finds a Job in a fake client's tracked objects by name — used by
-// reconciler-level tests below that only care whether/what got created, not the
-// full build matrix already covered above.
+// batchJobByName finds a Job in a fake client's tracked objects by name. Shared
+// across this file and snapshotjob_reconciler_test.go — those tests only care
+// whether/what got created, not the full build matrix already covered above.
 func batchJobByName(jobs *batchv1.JobList, name string) *batchv1.Job {
```
Ronkahn21:
```suggestion
// across this file and snapshotjob_reconciler_test.go — those tests only care
// whether/what got created, not the full build matrix already covered above.
func getBatchJobByName(jobs *batchv1.JobList, name string) *batchv1.Job {
```

operator/internal/controller/snapshotjob_podsnapshot.go
```diff
@@ -0,0 +1,189 @@
...
+// PodSnapshotContent — artifacts must outlive the SnapshotJob, and a controller
+// ownerRef would make Kubernetes GC delete this artifact along with its owner.
+func buildPodSnapshot(sj *snapshotv1alpha1.SnapshotJob, pod *corev1.Pod) (*snapshotv1alpha1.PodSnapshot, error) {
+	targetContainers := sj.Spec.PodSnapshotTemplate.TargetContainers
+	if len(targetContainers) != 1 {
+		return nil, fmt.Errorf("spec.podSnapshotTemplate.targetContainers must have exactly one entry, got %d", len(targetContainers))
+	}
+	// Copy: the produced PodSnapshot must not share a backing array with the
+	// SnapshotJob's own spec slice.
+	containers := append([]string(nil), targetContainers...)
```
Ronkahn21, line 89: there is deepcopy package

Ronkahn21 (commented):
1/n review

operator/internal/controller/snapshotjob_reconciler.go
```diff
@@ -41,7 +45,9 @@ type SnapshotJobReconciler struct {
 // +kubebuilder:rbac:groups=nvidia.com,resources=snapshotjobs,verbs=get;list;watch;update;patch
 // +kubebuilder:rbac:groups=nvidia.com,resources=snapshotjobs/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=nvidia.com,resources=snapshotjobs/finalizers,verbs=update
+// +kubebuilder:rbac:groups=nvidia.com,resources=podsnapshots,verbs=create;get;list;watch
 // +kubebuilder:rbac:groups=batch,resources=jobs,verbs=create;get;list;watch
+// +kubebuilder:rbac:groups=core,resources=pods,verbs=list;watch
 // +kubebuilder:rbac:groups=core,resources=events,verbs=create;patch
```
Ronkahn21, line 37: we dont use kustomize currently so this is redundant but if we want to keep can you move it above the strcut

operator/internal/controller/snapshotjob_reconciler.go
```diff
@@ -126,18 +133,86 @@ func (r *SnapshotJobReconciler) adoptExistingJob(ctx context.Context, sj *snapsh
...
+// the PodSnapshot. The pod may not exist yet even after the Job does (Job
+// creation and pod creation are not atomic); in that case Running=False/PodPending
+// is (re)set and this reconcile stops — the next Job status change (job.status.active
+// 0→1 when the pod is assigned) re-enqueues, since the controller watches only
+// the Job, not the pod.
+func (r *SnapshotJobReconciler) createPodSnapshotPhase(ctx context.Context, sj *snapshotv1alpha1.SnapshotJob, job *batchv1.Job) (ctrl.Result, error) {
+	pod, err := findSourcePod(ctx, r.Client, job)
+	if apierrors.IsNotFound(err) {
+		if setCondition(sj, snapshotv1alpha1.SnapshotJobConditionRunning, metav1.ConditionFalse,
+			snapshotv1alpha1.ReasonPodPending, "waiting for the source pod to be created") {
```
Ronkahn21: the pod is not pending for reason

Ronkahn21 (commented):
Can we separate desired-state reconciliation from status reporting?

`spec` should remain read-only desired input. Resource helpers should create/find/classify the Job, Pod, and PodSnapshot and return their observed state without directly mutating `SnapshotJob.Status`. A single status function should then derive and persist the complete status once per reconcile.

Status should be reconstructable from the current spec and observed resources; it should not behave as a workflow journal that depends on which branch ran previously. Resource creation and status updates are not atomic, so every status-write failure must be fully recoverable during the next reconcile.

The explicit exception is immutable spec validation: if the spec cannot produce a valid Job, that is immediately terminal and should result in `Failed=True/InvalidSpec` even though no child resource exists. The validation helper should still return a typed terminal outcome and let the centralized status function perform the update. Transient Kubernetes creation errors should remain retryable, while a Job that is created and later fails should derive its failure from `Job.Status`.

Conceptually:

```go
observed, reconcileErr := r.reconcileResources(ctx, sj)
desiredStatus := deriveStatus(sj, observed, reconcileErr)
return r.patchStatus(ctx, sj, desiredStatus)
```

operator/internal/controller/snapshotjob_reconciler.go
```diff
@@ -126,18 +133,86 @@ func (r *SnapshotJobReconciler) adoptExistingJob(ctx context.Context, sj *snapsh
...
+
+	snap, err := r.createPodSnapshot(ctx, sj, pod)
+	if err != nil {
+		if errors.Is(err, errPodSnapshotNameConflict) {
+			return r.failSnapshotJob(ctx, sj, snapshotv1alpha1.ReasonPodSnapshotNameConflict, err)
+		}
+		return ctrl.Result{}, err
+	}
+
+	sj.Status.PodSnapshotName = snap.Name
```
Ronkahn21: `createPodSnapshot()` does not necessarily return a newly created object; after `AlreadyExists`, it can return an existing PodSnapshot carrying `Ready=True` or `Failed=True`. This path ignores those conditions and unconditionally reports `CaptureInProgress`. If that PodSnapshot is already terminal, there may be no later update to correct the SnapshotJob status.

Could we route both newly created and adopted PodSnapshots through the same complete status derivation? That derivation should refresh `Running` from the current Job and explicitly set `Captured=False/CaptureInProgress` for a pending PodSnapshot. The pending case is also needed to recover `Captured` after a previous `Status().Update` conflict; `observe()` currently repairs only `podSnapshotName`.

operator/internal/controller/snapshotjob_podsnapshot.go
```diff
@@ -0,0 +1,189 @@
...
+// SnapshotJob — see buildPodSnapshot), so the label match from List *is* the
+// ownership check; there is no IsControlledBy to additionally verify. More than
+// one match is a controller invariant violation (deterministic naming makes it
+// impossible under normal operation), so it emits a warning and returns a
+// non-terminal error to requeue rather than silently picking one.
+func (r *SnapshotJobReconciler) findOwnedPodSnapshot(ctx context.Context, sj *snapshotv1alpha1.SnapshotJob) (*snapshotv1alpha1.PodSnapshot, error) {
+	var snaps snapshotv1alpha1.PodSnapshotList
+	if err := r.List(ctx, &snaps,
+		client.InNamespace(sj.Namespace),
+		client.MatchingLabels{snapshotv1alpha1.SnapshotJobOwnerLabel: sj.Name},
```
Ronkahn21: `SnapshotJobOwnerLabel` contains only the SnapshotJob name, which identifies a logical name but not a particular object instance. Because PodSnapshots deliberately survive SnapshotJob deletion, deleting and recreating a SnapshotJob with the same name causes the new object to adopt the old PodSnapshot and mirror its capture result.

Please record the SnapshotJob UID in an annotation or label and validate both name and UID during lookup and `classifyExistingPodSnapshot`. An artifact belonging to an older SnapshotJob incarnation should be treated as a deterministic-name conflict, not as owned.

operator/internal/controller/snapshotjob_reconciler.go
```diff
@@ -126,18 +133,86 @@ func (r *SnapshotJobReconciler) adoptExistingJob(ctx context.Context, sj *snapsh
...
+func (r *SnapshotJobReconciler) createPodSnapshotPhase(ctx context.Context, sj *snapshotv1alpha1.SnapshotJob, job *batchv1.Job) (ctrl.Result, error) {
+	pod, err := findSourcePod(ctx, r.Client, job)
+	if apierrors.IsNotFound(err) {
+		if setCondition(sj, snapshotv1alpha1.SnapshotJobConditionRunning, metav1.ConditionFalse,
+			snapshotv1alpha1.ReasonPodPending, "waiting for the source pod to be created") {
+			if err := r.Status().Update(ctx, sj); err != nil {
+				return ctrl.Result{}, fmt.Errorf("update SnapshotJob status: %w", err)
+			}
+		}
+		return ctrl.Result{}, nil
```
Ronkahn21: Returning without requeueing assumes another Job status update will always occur after the Pod enters the Pod informer cache. Job and Pod informer updates have no ordering guarantee: the Job event can enqueue reconciliation before the Pod becomes visible, and Pod events currently do not enqueue this controller.

Please either watch/map source Pods back to the SnapshotJob or return a bounded `RequeueAfter` while the Pod is not found. Otherwise capture can remain stuck even though the source Pod exists.

operator/internal/controller/snapshotjob_podsnapshot.go
oleg-kushniriov: Fixed

operator/internal/controller/snapshotjob_podsnapshot.go
oleg-kushniriov: I’d keep the switch: 
zero, one, and multiple matches are distinct outcomes
if/else hides the invariant-violation case

operator/internal/controller/snapshotjob_podsnapshot.go
oleg-kushniriov: Generated DeepCopy copies the whole API object. 
Only one `[]string` crosses objects here, so cloning that slice is smaller and explicit

operator/internal/controller/snapshotjob_reconciler.go
```diff
@@ -19,179 +22,240 @@ import (
...
 	}
-	if snapshotv1alpha1.IsSnapshotJobTerminal(sj) {
-		return ctrl.Result{}, nil
+
+	observed, result, err := r.reconcileResources(ctx, sj)
+	if err != nil {
+		return result, err
+	}
+	if observed.failure != nil {
+		r.Recorder.Event(sj, corev1.EventTypeWarning, observed.failure.reason, observed.failure.cause.Error())
```
Ronkahn21, line 82: can you explain this, if it a failure why only event... in what case we dont want return error here

operator/internal/controller/snapshotjob_podsnapshot.go
```diff
@@ -0,0 +1,201 @@
...
+	if !podSnapshotBelongsToSnapshotJob(existing, sj) {
+		return nil, fmt.Errorf("%w: PodSnapshot %q", errPodSnapshotNameConflict, name)
+	}
+	return existing, nil
+}
+
+func podSnapshotBelongsToSnapshotJob(snap *snapshotv1alpha1.PodSnapshot, sj *snapshotv1alpha1.SnapshotJob) bool {
+	return snap.Labels[snapshotv1alpha1.SnapshotJobOwnerLabel] == sj.Name &&
+		snap.Annotations[snapshotv1alpha1.SnapshotJobOwnerUIDAnnotation] == string(sj.UID)
+}
```
Ronkahn21, line 154: why the split ? I preder to have both of them as label

operator/internal/controller/snapshotjob_reconciler.go
```diff
@@ -27,171 +31,231 @@ import (
...
+
+// Reconcile first drives child resources toward the desired state, then derives
+// and patches SnapshotJob status once from the resulting observation.
 func (r *SnapshotJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	sj := &snapshotv1alpha1.SnapshotJob{}
 	if err := r.Get(ctx, req.NamespacedName, sj); err != nil {
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
 	// No finalizer: GC cascade-deletes the owned Job; nothing else to clean up here.
```
Ronkahn21: you can remove this comment

operator/internal/controller/snapshotjob_reconciler.go
```diff
@@ -27,171 +31,231 @@ import (
...
-// not exist yet, classify a foreign object holding its name, or observe an
-// owned one for Running.
-func (r *SnapshotJobReconciler) reconcileJob(ctx context.Context, sj *snapshotv1alpha1.SnapshotJob) (ctrl.Result, error) {
+// reconcileResources validates immutable input, then creates or observes the
+// source Job and its PodSnapshot. Retryable API failures are returned as errors;
+// immutable validation and deterministic-name conflicts are typed observations.
+func (r *SnapshotJobReconciler) reconcileResources(ctx context.Context, sj *snapshotv1alpha1.SnapshotJob) (snapshotJobObservation, ctrl.Result, error) {
+	desiredJob, err := buildSourceJob(sj)
+	if err != nil {
+		return terminalObservation(snapshotv1alpha1.ReasonInvalidSpec, err), ctrl.Result{}, nil
```
Ronkahn21: it should first get existing and if not foumd build it

operator/internal/controller/snapshotjob_reconciler.go
```diff
@@ -27,171 +31,231 @@ import (
...
+		if err := controllerutil.SetControllerReference(sj, desiredJob, r.Scheme()); err != nil {
+			return snapshotJobObservation{}, ctrl.Result{}, fmt.Errorf("set owner reference on source Job %q: %w", sj.Name, err)
+		}
+		if err := r.Create(ctx, desiredJob); err != nil {
+			if apierrors.IsAlreadyExists(err) {
+				return r.observeExistingSourceJob(ctx, sj)
+			}
+			r.Recorder.Event(sj, corev1.EventTypeWarning, "SourceJobCreateFailed", err.Error())
+			return snapshotJobObservation{}, ctrl.Result{}, fmt.Errorf("create source Job %q: %w", sj.Name, err)
+		}
```
Ronkahn21: export this to separate function it to heavy on eye inside this swich case

operator/internal/controller/snapshotjob_reconciler.go
```diff
@@ -27,171 +31,231 @@ import (
...
-	if err := r.Create(ctx, job); err != nil {
-		if apierrors.IsAlreadyExists(err) {
-			return r.adoptExistingJob(ctx, sj)
+	return r.reconcilePodSnapshotResources(ctx, sj, job)
+}
+
+// reconcilePodSnapshotResources observes an existing PodSnapshot or creates it
+// once exactly one controlled source Pod exists.
+func (r *SnapshotJobReconciler) reconcilePodSnapshotResources(ctx context.Context, sj *snapshotv1alpha1.SnapshotJob, job *batchv1.Job) (snapshotJobObservation, ctrl.Result, error) {
+	observed := snapshotJobObservation{job: job}
```
Ronkahn21, line 150: same here

operator/internal/controller/snapshotjob_reconciler.go
```diff
@@ -27,171 +31,231 @@ import (
...
-		if err := r.Status().Update(ctx, sj); err != nil {
-			return ctrl.Result{}, fmt.Errorf("update SnapshotJob status: %w", err)
+		snap, err = r.createPodSnapshot(ctx, sj, pod)
+		if errors.Is(err, errPodSnapshotNameConflict) {
+			observed.failure = &snapshotJobFailure{reason: snapshotv1alpha1.ReasonPodSnapshotNameConflict, cause: err}
+			return observed, ctrl.Result{}, nil
+		}
+		if err != nil {
+			return snapshotJobObservation{}, ctrl.Result{}, err
 		}
```
Ronkahn21: export to function

operator/internal/controller/snapshotjob_reconciler.go
```diff
@@ -27,171 +31,231 @@ import (
...
-	if ready {
-		if sj.Status.StartedAt == nil {
-			now := metav1.Now()
-			sj.Status.StartedAt = &now
-			changed = true
+// deriveSnapshotJobStatus is a pure derivation over current status and observed
+// resources. Existing timestamps are monotonic; conditions and references are
+// reconstructed whenever their source resource is observed.
+func deriveSnapshotJobStatus(sj *snapshotv1alpha1.SnapshotJob, observed snapshotJobObservation) snapshotv1alpha1.SnapshotJobStatus {
+	desired := sj.DeepCopy()
```
Ronkahn21: desired is not good name

operator/internal/controller/snapshotjob_reconciler.go
```diff
@@ -27,171 +31,231 @@ import (
...
+		if meta.FindStatusCondition(desired.Status.Conditions, snapshotv1alpha1.SnapshotJobConditionRunning) == nil {
+			setCondition(desired, snapshotv1alpha1.SnapshotJobConditionRunning, metav1.ConditionFalse,
+				snapshotv1alpha1.ReasonPodPending, "source pod was never observed ready before this SnapshotJob failed")
+		}
+		setCondition(desired, snapshotv1alpha1.SnapshotJobConditionCaptured, metav1.ConditionFalse,
+			failure.reason, "capture did not complete: "+failure.cause.Error())
+		setCondition(desired, snapshotv1alpha1.SnapshotJobConditionCompleted, metav1.ConditionFalse,
+			failure.reason, "the SnapshotJob failed before completing: "+failure.cause.Error())
+		setCondition(desired, snapshotv1alpha1.SnapshotJobConditionFailed, metav1.ConditionTrue,
+			failure.reason, failure.cause.Error())
```
Ronkahn21: split this entire function to smaller function

operator/internal/controller/snapshotjob_reconciler.go
oleg-kushniriov: in `deriveSnapshotJobStatus` converts it into terminal conditions

oleg-kushniriov: This is terminal failure, not retryable error. 
When we emit the event and then save `Failed=True` in status. If we return error here, status will not be saved and controller will retry for no reason.

Ronkahn21 (approved):

PR #77, opened 2026-08-18
fix(agent): release checkpoint after ready status

galletas1712:
## Summary
- persist `PodSnapshotContent` Ready before writing the `snapshot-complete` sentinel that releases the checkpoint target
- keep the target blocked when the Ready status patch fails
- fail closed by killing the target when release fails after Ready is durable
- make the Ready patch optimistic so a concurrent terminal Failed write wins; on conflict, re-read and skip release if the object is already terminal
- kill the still-blocked target when a successful dump hits `LeaseCancelled`, so a terminal Failed cannot leave it polling forever
- port of [ai-dynamo/dynamo#13366](https://github.com/ai-dynamo/dynamo/pull/13366) with the [Ready-vs-Failed race](https://github.com/ai-dynamo/dynamo/pull/13366#discussion_r3799008606) from @hhzhang16 addressed here

## Test plan
- [x] `go test ./agent/internal/controller -count=1`
- [x] `go test -race ./agent/internal/controller -count=1`
- [x] `go test ./agent/internal/controller -count=10 -run 'TestRunCheckpoint_|TestMarkCheckpointReadyAndRelease_|TestReconcileSnapshotContent_ResumeWritesReady'`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved checkpoint and resume reliability by recording readiness before releasing processes.
  * Prevented releases when readiness updates fail or checkpoint content is already failed.
  * Added safer cleanup when cancellation, release operations, or checkpoint processes fail.
  * Improved handling of concurrent updates, conflicts, retries, stale requests, and artifact recovery.
  * Ensured newer pending checkpoint content is not overlooked during resume.
  * Treated already-exited checkpoint processes as successfully completed.

* **Tests**
  * Added coverage for readiness ordering, release failures, cancellation, resume behavior, and conflict handling.


agent/internal/controller/podsnapshotcontent.go
```diff
@@ -408,6 +403,52 @@ func (w *NodeController) setSnapshotContentSucceeded(ctx context.Context, conten
...
+		if isContentReady(current) {
+			return w.writeSentinelOrKill(logger, containerPID, content.Name)
+		}
+		if err := w.setSnapshotContentSucceeded(ctx, current); err != nil {
+			logger.Error(err, "Failed to write PodSnapshotContent ready status", "content", content.Name)
+			return err
+		}
+		ready = current
+	}
+	return w.writeSentinelOrKill(logger, containerPID, ready.Name)
```
julienmancuso: [P1] Preserve a post-Ready finalization path. This call persists Ready before writing snapshot-complete. If the agent exits between the successful status patch and this sentinel write, both reconcilePodSnapshotContent and chooseActiveContent treat the content as terminal, so no later reconcile writes the sentinel. The source process remains blocked indefinitely and can retain its GPU. Please retain an idempotent finalizer for Ready, non-Failed content without letting it starve newer active work.

agent/internal/controller/podsnapshotcontent.go
```diff
@@ -270,19 +271,13 @@ func (w *NodeController) runCheckpoint(
...
-	// renewal failure). A clean context.Canceled (outer ctx shutdown) is not a lease failure.
-	if cause := context.Cause(leaseCtx); cause != nil && !errors.Is(cause, context.Canceled) {
-		logger.Error(cause, "Lease cancelled during checkpoint")
-		if patchErr := w.setSnapshotContentFailed(ctx, content, "LeaseCancelled", cause); patchErr != nil {
-			logger.Error(patchErr, "Failed to write PodSnapshotContent failed status", "content", content.Name)
-		}
+	// The dump is on disk. Persist Ready and write snapshot-complete even if the lease
+	// was cancelled during the dump. A clean context.Canceled (outer ctx shutdown) is
+	// not a lease failure and is also not a reason to skip Ready-and-release.
+	if err := w.markCheckpointReadyAndRelease(ctx, content, containerPID); err != nil {
```
julienmancuso, line 362: [P1] Do not finalize after lease ownership is lost. renewLease cancels leaseCtx both on renewal failures and when HolderIdentity changes; production checkpointing can still return nil after cancellation because CRIU Dump and the final RemoveAll/Rename are not context-aware. A stale holder can therefore mark Ready and release while another holder is writing the same artifact, so this Ready content can refer to an artifact later replaced by the competing writer. Preserve the LeaseCancelled failure/kill behavior or revalidate ownership with fencing before finalizing.

agent/internal/controller/podsnapshotcontent.go
```diff
@@ -408,6 +403,52 @@ func (w *NodeController) setSnapshotContentSucceeded(ctx context.Context, conten
...
+			logger.Error(getErr, "Failed to re-read PodSnapshotContent after Ready conflict", "content", content.Name)
+			return getErr
+		}
+		if isContentFailed(current) {
+			return w.killCheckpointProcess(logger, containerPID, "checkpoint content already failed")
+		}
+		if isContentReady(current) {
+			return w.writeSentinelOrKill(logger, containerPID, content.Name)
+		}
+		if err := w.setSnapshotContentSucceeded(ctx, current); err != nil {
```
julienmancuso: [P2] Re-read when this retry conflicts too. If the first Ready patch conflicts for a nonterminal update, current is re-read, and then a concurrent Failed writer wins before this second optimistic patch, this branch returns the conflict without observing Failed or killing the target. The content is now terminal, so later reconciles skip it and the target remains blocked. Handle conflicts in a loop, or re-read this conflict, until Ready succeeds or a terminal state is observed.

julienmancuso (changes requested):
Requesting changes for three concurrency and liveness issues. The restart window is acknowledged in the thread, but it can leave the target blocked indefinitely.

julienmancuso (approved):

PR #80, opened 2026-08-19
feat: wire the SnapshotJob completion gate, failure handling, and cleanup

oleg-kushniriov:
### Summary

- Adds the two-signal completion gate: `Completed=True` only once both the `PodSnapshot`'s `Ready=True` and the source `Job`'s `Complete=True` are observed. `Captured=True` alone sets `Completed=False/WaitingForPodCompletion`; `Job Complete=True` alone leaves `Completed` untouched, since the CRIU dump — not the helper containers — is what's pending.
- Fills in the rest of the `Failed=True` reason matrix that earlier PRs left classified-but-not-acted-on: `JobNameConflict` (foreign object at the Job's deterministic name, both on the initial `Get` and the `Create`-time `AlreadyExists` race), `PodSnapshotNameConflict` (same, for the PodSnapshot), and `JobDeleted` (the source Job vanishes after a PodSnapshot was already created from it — a non-execution failure, not a fresh object to create).
- Adds `JobFailed`/`DeadlineExceeded`, derived from the Job's own `Failed` condition — `activeDeadlineSeconds` expiry gets its own reason, distinct from a plain container exit failure. Job failure wins over a successful capture: `observeOwnedJob` checks the Job's `Failed` condition before the PodSnapshot phase gets a chance to report `Captured=True`.
- Splits `CaptureFailed` from `PodSnapshotFailed`: a PodSnapshot `Failed=True` is now classified by its own `Failed` condition's reason — a bind-stage failure (`ContentConflict`, `SourcePodNotFound`, `StalePodReference`, set by `PodSnapshotReconciler` before any agent interaction) maps to `PodSnapshotFailed`; anything else (a real CRIU/CUDA dump failure, mirrored from the agent-driven `PodSnapshotContent`) maps to `CaptureFailed`.
- Cleanup: on `Completed=True`, the source Job is deleted with `Preconditions{UID}` + `PropagationPolicy(Background)`, retried on every later reconcile of an already-`Completed` SnapshotJob until confirmed gone (mirrors Dynamo's `cleanupTerminalCheckpointJob`) — this is why `Completed=True` still re-enters `Reconcile` instead of short-circuiting like `Failed=True` does. On `Failed=True`, the Job is deliberately never deleted, so the pod and its container logs stay available for debugging.
- Registers `SnapshotJobReconciler` in `cmd/manager/main.go` and grants it the RBAC it needs (`snapshotjobs`, `snapshotjobs/status`, `batch/jobs` with `delete`) — **this is the PR that turns the feature on**. Every earlier PR built and tested pieces of this reconciler without it running in production; this one wires it into the manager.

### Scope

- `operator/internal/controller/snapshotjob_reconciler.go` — completion gate, full `Failed=True` reason matrix, `completedAt` set-once-and-sticky, cleanup (`ensureJobDeleted`, `completeSnapshotJob`).
- `operator/cmd/manager/main.go` — construct and register `SnapshotJobReconciler`.
- `charts/snapshot/templates/operator-rbac.yaml` — `snapshotjobs`, `snapshotjobs/status`, `snapshotjobs/finalizers`, `batch/jobs` (add `delete`), `snapshotjobs.nvidia.com` in `crdUpgrade.resourceNames`.
- `operator/internal/controller/snapshotjob_completion_test.go` (new) — the completion-gate matrix, every failure reason, cleanup retry-until-gone, Job-failure-wins-over-Captured, `PodSnapshotFailed` vs `CaptureFailed` classification.
- `operator/internal/controller/snapshotjob_reconciler_test.go` — updated the two PR3-era foreign-Job tests (`TestSnapshotJobReconcileClassifiesForeignJob(OnAlreadyExistsRace)`) to assert the new terminal `Failed=True/JobNameConflict` behavior instead of the old classify-only behavior.

### Test plan

- [x] Completion gate: `Captured=True` alone → `Completed=False/WaitingForPodCompletion`, Job preserved; `Job Complete=True` alone → `Completed` untouched, Job preserved; both → `Completed=True/JobCompleted`, `completedAt` set, Job deleted.
- [x] Every `Failed=True` reason: `JobNameConflict` (both discovery paths), `PodSnapshotNameConflict`, `JobDeleted`, `JobFailed`, `DeadlineExceeded`, `PodSnapshotFailed` vs `CaptureFailed`.
- [x] Job-failure-wins-over-Captured: `Captured=True` already persisted, then the Job fails → `Failed=True/JobFailed`, `Captured=True` is not reverted, Job preserved.
- [x] Cleanup: Job deleted on `Completed=True`; retried on a later reconcile if not yet gone; tolerates the Job already being gone; never deleted on `Failed=True`; the terminal status write is persisted before the delete is attempted (a failed delete doesn't lose the record of why the Job is gone).
- [x] `helm template` renders every new RBAC rule; the rendered rules match the kubebuilder markers.
- [x] `golangci-lint` clean, `gofmt` clean, license headers pass, `make verify-crds` clean (no drift between the CRD source and the chart copy).

## Summary by CodeRabbit

* **New Features**
  * SnapshotJob status now identifies the specific source Job and captured PodSnapshot.
  * Conditions use consistent keyed entries with clearer progress and failure details.
  * Snapshot processing now handles source completion, deletion, replacement, and conflicts more reliably.
  * SnapshotJob validation requires a specification and limits names to 63 characters.

* **Bug Fixes**
  * Improved handling of deadlines, failed workloads, missing resources, races, and cleanup.
  * Prevented stale or replacement resources from being incorrectly reused.

* **Documentation**
  * Updated installation and permissions documentation for SnapshotJob support.


operator/cmd/manager/main.go
```diff
@@ -34,7 +38,13 @@ func main() {
 	}
 
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
-		Scheme:                        scheme,
+		Scheme: scheme,
+		Cache: cache.Options{ByObject: map[client.Object]cache.ByObject{
+			// Cache only prepared checkpoint source Pods.
+			&corev1.Pod{}: {
+				Label: labels.SelectorFromSet(labels.Set{v1alpha1.CheckpointSourceLabel: "true"}),
+			},
```
Ronkahn21, line 46: we need make sure it will keep the caching to all the other objects

operator/cmd/manager/main.go
```diff
@@ -55,14 +65,25 @@ func main() {
...
+		Recorder:  mgr.GetEventRecorderFor("podsnapshot-controller"),
 	}
 	if err := podSnapshotReconciler.SetupWithManager(mgr); err != nil {
 		ctrl.Log.Error(err, "unable to set up PodSnapshot controller")
 		os.Exit(1)
 	}
 
+	snapshotJobReconciler := &controller.SnapshotJobReconciler{
+		Client:    mgr.GetClient(),
+		APIReader: mgr.GetAPIReader(),
```
Ronkahn21: lets change this non cached API
```suggestion
		NonCacheReadClient: mgr.GetAPIReader(),
```

operator/cmd/manager/main.go
```diff
@@ -55,14 +65,25 @@ func main() {
 	}
 
 	podSnapshotReconciler := &controller.PodSnapshotReconciler{
-		Client:   mgr.GetClient(),
-		Recorder: mgr.GetEventRecorderFor("podsnapshot-controller"),
+		Client:    mgr.GetClient(),
+		APIReader: mgr.GetAPIReader(),
```
Ronkahn21: same as line 79

operator/internal/controller/podsnapshot_reconciler.go
```diff
@@ -123,9 +132,109 @@ func (sr *PodSnapshotReconciler) mirrorBoundContent(ctx context.Context, snap *s
...
+	}
+	return pod, nil
+}
+
+func pendingCaptureSourceFailure(snap *snapshotv1alpha1.PodSnapshot, content *snapshotv1alpha1.PodSnapshotContent, pod *corev1.Pod) (string, error) {
+	key := client.ObjectKeyFromObject(pod)
+	wantUID := snap.Spec.Source.PodRef.UID
+	if wantUID == "" {
+		wantUID = content.Spec.Source.PodRef.UID
+	}
```
Ronkahn21, line 265: redundant, immutable

operator/internal/controller/podsnapshot_reconciler.go
```diff
@@ -123,9 +132,109 @@ func (sr *PodSnapshotReconciler) mirrorBoundContent(ctx context.Context, snap *s
...
+	case pod.Status.Phase == corev1.PodSucceeded:
+		return snapshotv1alpha1.ReasonSourceCompletedWithoutCapture,
+			fmt.Errorf("source pod %q completed before PodSnapshotContent %q reported capture success or failure", key.String(), content.Name)
+	case pod.Status.Phase == corev1.PodFailed || pod.DeletionTimestamp != nil:
+		return "SourcePodGone",
+			fmt.Errorf("source pod %q became unavailable before PodSnapshotContent %q reported capture success or failure", key.String(), content.Name)
+	default:
+		return "", nil
+	}
+}
```
Ronkahn21, line 281: I am not sure why you need to add this, the content should do it so its redeunnt

operator/internal/controller/podsnapshot_reconciler.go
```diff
@@ -52,7 +59,8 @@ var errContentConflict = errors.New("existing PodSnapshotContent belongs to anot
 // back to the PodSnapshot, and cascades deletion to the PodSnapshotContent.
 type PodSnapshotReconciler struct {
 	client.Client
-	Recorder record.EventRecorder
+	APIReader client.Reader
+	Recorder  record.EventRecorder
 }
```
Ronkahn21, line 70: there should be only one source of true once you crated a snasphotconoent with UID 
the failed and ready come from it, getting directly after could cuease edge case senario where stall cache and dealy will failed a good capture, once you create content its bound to pod if it failed it just will raport it al the way up.

operator/internal/controller/podsnapshot_reconciler.go
```diff
@@ -123,9 +132,109 @@ func (sr *PodSnapshotReconciler) mirrorBoundContent(ctx context.Context, snap *s
...
+func (sr *PodSnapshotReconciler) readAuthoritativeBoundContent(ctx context.Context, snap *snapshotv1alpha1.PodSnapshot, cachedContent *snapshotv1alpha1.PodSnapshotContent) (*snapshotv1alpha1.PodSnapshotContent, error) {
+	content := &snapshotv1alpha1.PodSnapshotContent{}
+	if err := sr.APIReader.Get(ctx, client.ObjectKeyFromObject(cachedContent), content); err != nil {
+		return nil, fmt.Errorf("re-read pending PodSnapshotContent %q: %w", cachedContent.Name, err)
+	}
+	if err := verifyContentBacklink(snap, content); err != nil {
+		return nil, fmt.Errorf("%w: %v", errContentConflict, err)
+	}
+	return content, nil
+}
```
Ronkahn21, line 246: using non cache and chached together lead to very wired behavior so mixing them should reserve to very special case

operator/internal/controller/podsnapshot_reconciler.go
```diff
@@ -123,9 +132,109 @@ func (sr *PodSnapshotReconciler) mirrorBoundContent(ctx context.Context, snap *s
...
+	}
+
+	reason, cause := pendingCaptureSourceFailure(snap, content, pod)
+	if cause != nil {
+		return sr.failPodSnapshot(ctx, snap, reason, cause)
+	}
 	return sr.propagateStatus(ctx, snap, content)
 }
 
+func (sr *PodSnapshotReconciler) readAuthoritativeBoundContent(ctx context.Context, snap *snapshotv1alpha1.PodSnapshot, cachedContent *snapshotv1alpha1.PodSnapshotContent) (*snapshotv1alpha1.PodSnapshotContent, error) {
```
Ronkahn21, line 237: all the CRD are immutable so cache is just fine

operator/internal/controller/podsnapshot_reconciler.go
```diff
@@ -369,18 +478,89 @@ func (sr *PodSnapshotReconciler) handleDelete(ctx context.Context, snap *snapsho
...
 		For(&snapshotv1alpha1.PodSnapshot{}).
 		Watches(
 			&snapshotv1alpha1.PodSnapshotContent{},
 			handler.EnqueueRequestsFromMapFunc(podSnapshotContentToPodSnapshot),
 		).
+		Watches(
+			&corev1.Pod{},
+			handler.EnqueueRequestsFromMapFunc(sr.sourcePodToPodSnapshots),
+			builder.WithPredicates(sourcePodCapturePredicate()),
+		).
```
Ronkahn21, line 551: why are you watching the pod, the reconciler is mostly a action function,
I understadnd you want to watch on pod status changed but we need to decide what is the source of true comes from

operator/internal/controller/snapshotjob_source_job.go
```diff
@@ -0,0 +1,250 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
Ronkahn21, line 1: is there a piticilar reason we have so many files for one reconcilers 
its make it hard to follow, can you  explain from where to start or main fiuntion in each file

operator/internal/controller/snapshotjob_podsnapshot.go
```diff
@@ -25,32 +28,80 @@ import (
...
+// owned-Job watch is the normal signal that a source Pod may now exist.
+const sourcePodRequeueBackstop = 30 * time.Second
+
+// createPodSnapshotForSourceJob waits for the source Pod, then creates or
+// classifies the deterministic PodSnapshot. A missing active source Pod gets a
+// bounded backstop requeue; completed source work with no remaining Pod is a
+// concrete missing-capture failure.
+func (r *SnapshotJobReconciler) createPodSnapshotForSourceJob(ctx context.Context, sj *snapshotv1alpha1.SnapshotJob, job *batchv1.Job) (snapshotJobObservation, ctrl.Result, error) {
+	observed := snapshotJobObservation{job: job}
+	pod, found, err := findSourcePod(ctx, r.APIReader, job)
```
Ronkahn21: I am not sure why to use non cache reader

operator/internal/controller/snapshotjob_podsnapshot.go
```diff
@@ -25,32 +28,80 @@ import (
...
-func findSourcePod(ctx context.Context, reader client.Reader, job *batchv1.Job) (*corev1.Pod, error) {
+// sourcePodRequeueBackstop is only a safety net for a missed Job event; the
+// owned-Job watch is the normal signal that a source Pod may now exist.
+const sourcePodRequeueBackstop = 30 * time.Second
+
+// createPodSnapshotForSourceJob waits for the source Pod, then creates or
+// classifies the deterministic PodSnapshot. A missing active source Pod gets a
+// bounded backstop requeue; completed source work with no remaining Pod is a
+// concrete missing-capture failure.
+func (r *SnapshotJobReconciler) createPodSnapshotForSourceJob(ctx context.Context, sj *snapshotv1alpha1.SnapshotJob, job *batchv1.Job) (snapshotJobObservation, ctrl.Result, error) {
```
Ronkahn21, line 38: we should make sure it the the podsnasphot does  not exists before I dont see the logic the calls this function

Ronkahn21 (commented):
1/n review

operator/cmd/manager/main.go
oleg-kushniriov: the `&corev1.Pod{}` entry restricts just the Pod cache to pods labeled `nvidia.com/snapshot-is-checkpoint-source=true`, so the operator doesn't cache every Pod in the cluster. Every other kind (`SnapshotJob, PodSnapshot, PodSnapshotContent, batch/v1 Job`, …) keeps controller-runtime's default full caching

operator/internal/controller/podsnapshot_reconciler.go
oleg-kushniriov: immutability covers the recorded reference, not the live pod - a same-named pod recreated after the original died has a new UID and must not be mistaken for the pinned source.

operator/internal/controller/podsnapshot_reconciler.go
oleg-kushniriov: the failure mode you describe - stale cache failing a good capture - is the reason the direct reader is there.

oleg-kushniriov: It runs only right before writing a sticky failure - and since the agent persists content `Ready` before releasing the source, the re-read always sees a Ready that won the race. Cache-only reads here would introduce the stale-cache failure, not remove it.

operator/internal/controller/podsnapshot_reconciler.go
oleg-kushniriov: agreed on the principle
direct reads only as the final confirmation immediately before an irreversible sticky-failure write, as a confirmation

operator/internal/controller/podsnapshot_reconciler.go
oleg-kushniriov: the spec is immutable, but this read is for the content's `Status`.
the agent's Ready/Failed may be persisted at the API server and not yet in our informer.

operator/internal/controller/podsnapshot_reconciler.go
oleg-kushniriov: the PodSnapshotContent stays the source of truth. It exists for the two states no content event can unblock - initial binding (no content exists yet) and the dead-agent backstop - and it's kept narrow by the predicate plus the checkpoint-source-only pod cache.

operator/internal/controller/snapshotjob_source_job.go
oleg-kushniriov: One reconciler split by the child resource each file manages.
start at Reconcile in snapshotjob_reconciler.go

- snapshotjob_reconciler.go - entrypoint, status derivation, wiring
 - snapshotjob_job.go - buildSourceJob (desired batch/v1 Job from spec)
- snapshotjob_source_job.go - create/adopt/classify the source Job (one-shot identity, terminal classification, cleanup)
- snapshotjob_podsnapshot.go - create/adopt/classify the produced PodSnapshot; locate the source Pod
- snapshotjob_validation.go - spec checks shared by build and adoption

I can merge them but this way it less crowded

operator/internal/controller/snapshotjob_podsnapshot.go
oleg-kushniriov: The SnapshotJob controller has no pod watch by design.

Read using uncached used for irreversible actions identity check, same as in other places - irreversible decisions -> read direct.

operator/internal/controller/snapshotjob_podsnapshot.go
oleg-kushniriov: The check is in the single caller - it looks the PodSnapshot up first and only calls this on "not found". A race can't overwrite anything either: the create fails with "already exists", if it's our own PodSnapshot we continue with it, and if it belongs to someone else the SnapshotJob fails with a name conflict.

operator/internal/controller/snapshotjob_podsnapshot.go
```diff
@@ -25,32 +28,80 @@ import (
...
+	observed := snapshotJobObservation{job: job}
+	pod, found, err := findSourcePod(ctx, r.NonCacheReadClient, job)
```
Ronkahn21: Why is the normal source-Pod lookup using `NonCacheReadClient`? A direct read does not eliminate before/after drift relative to the cached Job and PodSnapshot reads; it only mixes consistency domains within one reconciliation. The Pod is only an input for constructing the PodSnapshot, so the normal path should be cache-first and tolerate eventual consistency through watches/requeues. Please reserve a direct read for confirming cached absence or identity immediately before an irreversible terminal decision.

operator/internal/controller/snapshotjob_podsnapshot.go
```diff
@@ -25,32 +28,80 @@ import (
...
+	if err != nil {
+		return snapshotJobObservation{}, ctrl.Result{}, fmt.Errorf("find source pod for Job %q: %w", job.Name, err)
+	}
+	if !found {
+		observed.sourcePodMissing = true
+		if classifySourceJobTerminal(job).state == sourceJobComplete {
+			observed.failure = &snapshotJobFailure{
+				reason: snapshotv1alpha1.ReasonSourceCompletedWithoutCapture,
+				cause:  fmt.Errorf("source Job completed and its pod is gone before a PodSnapshot capture result was recorded"),
```
Ronkahn21: `SourceCompletedWithoutCapture` currently depends on Pod-retention timing. If the Job is `Complete=True` and its succeeded Pod is still retained, this path proceeds to create a new PodSnapshot; if the same Pod has already been removed, it fails here. Those are the same logical state, and once the containers have exited a new capture cannot start. Please check the deterministic PodSnapshot first: if none exists and the Job is complete, report `SourceCompletedWithoutCapture` consistently without creating a new capture. If it already exists, its status must remain the authoritative capture result.

Ronkahn21 (changes requested):
Requesting changes to the reconciliation model before this controller is enabled.

The current implementation duplicates capture-lifecycle interpretation across PodSnapshot, PodSnapshotContent, the source Pod, and SnapshotJob, then tries to resolve ordering with a mixture of cached reads, direct reads, and grace periods. Neither cached nor sequential API reads form an atomic view; before/after update drift still exists.

The ownership boundary should be explicit:

- Before a PodSnapshot exists, inspect the source Pod only for the identity needed to create it.
- Once created, PodSnapshot.status is the sole capture signal consumed by SnapshotJob.
- Once PodSnapshotContent is bound, its Ready/Failed status should flow to PodSnapshot rather than PodSnapshot independently deriving another terminal result from the Pod.
- Job.status contributes only the independent workload-completion/failure signal.
- SnapshotJob.status is the public aggregate of those authoritative child signals.
- Normal reconciliation should be cache-first; direct reads should be limited to confirming absence or identity immediately before an irreversible decision.

The state machine must be monotonic so stale observations can delay progress but cannot invent a different terminal outcome. The completed-Job path called out inline currently violates this: the result changes depending on whether the succeeded Pod has already been garbage-collected.

If a dead-agent backstop is required, model it as an explicit timeout/failure in the authoritative capture lifecycle rather than making the source Pod a competing truth after binding. Please also address the valid deletion-timestamp/grace-period finding already posted on podTerminalTime.

operator/internal/controller/podsnapshot_reconciler.go
Ronkahn21: The direct reader addresses one cache-lag race, but it does not address the ownership-boundary concern. Reads of Content and Pod are still sequential and can straddle an agent update even when both hit the API server. More importantly, after Content is bound this controller is independently deriving a terminal capture result from the Pod instead of propagating the authoritative Content result. Once bound, `Ready`/`Failed` should flow from PodSnapshotContent to PodSnapshot. If a dead-agent backstop is required, model it as an explicit timeout/failure on that authoritative lifecycle rather than creating a competing Pod-derived outcome.

Ronkahn21:
To clarify the requested reconciliation model:

Once a PodSnapshot has been created, `PodSnapshot.status` must be the only capture signal consumed by SnapshotJob. PodSnapshot already encapsulates the lower-level PodSnapshotContent and source-Pod lifecycle, so SnapshotJob should not independently reinterpret those resources.

The source Job has only two roles:

1. Before PodSnapshot creation, a Job event is a hint that the source Pod may now exist. The Pod is read only to obtain the identity needed to construct the deterministic PodSnapshot.
2. After PodSnapshot creation, `Job.status` supplies only the independent workload completion/failure signal required by the two-signal gate. It must not produce or override the capture result.

The resulting flow should be:

1. Look up the deterministic PodSnapshot.
2. If it is absent, look up the Job/source Pod only to create it or determine that creation is no longer possible.
3. If it exists, derive `Captured` exclusively from `PodSnapshot.status`.
4. Use `Job.status` only for the independent completion/failure gate.
5. Derive and patch `SnapshotJob.status` once.

This separation also removes most of the need for `NonCacheReadClient`. Direct API reads do not make a multi-resource decision atomic: each object can change between sequential reads. For example, the controller can read Content as pending, the agent can publish `Ready`, and then the controller can read the Pod as terminal and infer failure from observations representing different moments. Reversing the order produces equivalent races; a grace period only reduces their probability.

With PodSnapshot as the capture boundary, stale cached status can only delay progress. The PodSnapshot watch will deliver the authoritative update and requeue SnapshotJob; SnapshotJob cannot invent a competing terminal result from Pod or Content state. Direct reads can then be limited to narrow identity or confirmed-absence checks immediately before an irreversible decision.

This gives each condition one owner:

- PodSnapshot controls `Captured`.
- Job controls workload completion/failure.
- SnapshotJob publishes the aggregate `Completed`/`Failed` result.

It also removes Pod/Content update-order races, Pod-retention-dependent outcomes, and much of the grace-period and cache/API-reader test matrix.

Ronkahn21 (approved):

PR #81, opened 2026-08-19
test(e2e): add SnapshotJob capture and restore coverage

oleg-kushniriov:
### Summary

- Adds `e2e/tests/test_snapshotjob.py`: end-to-end coverage of the SnapshotJob CRD itself, not just the PodSnapshot it produces. `test_snapshotjob_completes_and_restore_recovers_state` drives a SnapshotJob through the full lifecycle — source Job creation, Running, PodSnapshot creation, Captured, the two-signal completion gate, `Completed=True`, Job cleanup — then restores from the produced artifact and asserts the CPU/file/GPU state and the artifact contents. `test_snapshotjob_deadline_exceeded_when_never_ready` covers the negative path: a workload that never signals ready hits `activeDeadlineSeconds` and the SnapshotJob reaches `Failed=True/DeadlineExceeded`.
- The SnapshotJob flow needs a workload that can *end itself*: unlike the plain-PodSnapshot flow's test pod (which the test deletes explicitly), a SnapshotJob's `Completed=True` requires the source Job's own `Complete=True`, which requires its container to actually exit 0. Adds self-exiting CPU/CUDA workload script variants (`SNAPSHOTJOB_CPU_SOURCE`/`SNAPSHOTJOB_CUDA_SOURCE`) that watch for the agent's `snapshot-complete` control-volume sentinel (written after a successful dump, confirmed against the chart's `leaveRunning: true` default) and exit once they see it.
- Adds `snapshotjob_pod_template`/`snapshotjob_hang_pod_template` — unlike the plain flow's pod spec, a SnapshotJob's `spec.podTemplate` must **not** pre-wire the control volume or any `nvidia.com/snapshot-*` annotations: the controller derives and injects all of that itself, so the template here is what a real caller would actually write.
- The source pod's name isn't predictable (the Job controller appends a random suffix), so adds `wait_for_job_source_pod`/`k8s.list_job_pods` to find it by the `batch.kubernetes.io/job-name` label instead of assuming a name.
- Adds `create_snapshotjob`, `cleanup_snapshotjob`, and `debug_dump_snapshotjob` — cleanup is shaped differently from the plain flow: deleting the SnapshotJob cascades to its source Job and pod via ownerReference, but the produced PodSnapshot deliberately carries none (artifacts must outlive the SnapshotJob) and needs the same explicit cleanup as the plain-PodSnapshot flow already has.
- Adds a `snapshot_tag` `workflow_dispatch` input to `e2e.yaml` so a manual run can pin an exact operator/agent image tag (e.g. published from this branch via a manual `push-artifacts.yaml` dispatch) instead of always auto-resolving whatever is newest in GHCR — needed to actually validate a feature branch's own image rather than racing against `main`'s latest publish.
- Wires the new test into CI (enumerated explicitly, like `test_snapshot_lifecycle.py`) and into the README's direct/vCluster run instructions.

### Scope

- `e2e/tests/test_snapshotjob.py` (new).
- `e2e/snapshot_e2e/workloads.py` — `base_pod_spec` gains a `control_volume` flag (default `True`, existing callers unaffected); `snapshotjob_pod_template`, `snapshotjob_hang_pod_template`, `snapshotjob_source_command`, `SNAPSHOTJOB_CPU_SOURCE`, `SNAPSHOTJOB_CUDA_SOURCE`, `SNAPSHOT_COMPLETE`.
- `e2e/snapshot_e2e/lifecycle.py` — `create_snapshotjob`, `wait_for_job_source_pod`, `cleanup_snapshotjob`, `debug_dump_snapshotjob`, `print_custom_objects_named`, `SNAPSHOTJOBS`.
- `e2e/snapshot_e2e/k8s.py` — `list_job_pods`, `JOB_NAME_LABEL`.
- `.github/workflows/e2e.yaml` — new `snapshot_tag` input, a new "Run SnapshotJob lifecycle tests" step.
- `e2e/README.md` — run instructions for both local modes.

**Depends on:** the SnapshotJob completion/wiring PR (the reconciler must actually be registered in `main.go` for any of this to run against a live cluster).

### Test plan

- [x] `pytest e2e/tests --collect-only` succeeds for the full suite (imports, fixtures, markers all resolve) — verified locally.
- [x] The generated CUDA C source (`SNAPSHOTJOB_CUDA_SOURCE`) compiles clean with `cc -fsyntax-only -Wall`.
- [x] `base_pod_spec`'s refactor produces byte-identical output for existing callers (`source_pod`/`restore_pod`) — verified via a dry construction diff.
- [ ] Actual execution against a live GPU cluster — **not yet run**; no GPU cluster was reachable in the environment this was authored in. Needs a real CI/GPU-cluster run before merge.

## Summary by CodeRabbit

* **New Features**
  * Added optional snapshot image tag selection for manually triggered E2E workflows.
  * Added SnapshotJob coverage for capture, restore, helper containers, scheduling, deadlines, failures, cleanup, and validation.
  * Added enhanced diagnostics for failed snapshot operations.

* **Documentation**
  * Updated local E2E instructions to include SnapshotJob tests.

* **Tests**
  * Added checks for workload ordering, readiness signaling, termination behavior, lifecycle handling, and failure diagnostics.


.github/workflows/build-and-e2e.yaml
```diff
@@ -0,0 +1,46 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# TEMPORARY: scoped to this branch only, to validate the SnapshotJob e2e
+# suite (test_snapshotjob.py) against this branch's own build without a
+# human babysitting a manual push-artifacts.yaml -> e2e.yaml dispatch chain.
+# Remove this file before merging RUN-39806-snapshotjob-e2e-docs to main.
```
hhzhang16: Can you remove this from the MR? To be safe; you can always add this file as a draft MR as the last element in the PR stack

e2e/snapshot_e2e/workloads.py
```diff
@@ -131,26 +173,39 @@ def base_pod_spec(
...
+    # what the feature actually promises callers they do not have to set up.
     container: dict[str, Any] = {
         "name": CONTAINER,
         "image": run.image,
         "imagePullPolicy": "IfNotPresent",
         "command": ["/bin/bash", "-lc", command],
         "volumeMounts": [
-            {"name": "snapshot-control", "mountPath": CONTROL_DIR},
+            {"name": "checkpoint-storage", "mountPath": CHECKPOINT_DIR},
         ],
```
hhzhang16: This is always adding the checkpoint PVC, I thought we didn't want to do this? Let me know if I'm misunderstanding something

oleg-kushniriov: Good catch. fixing

.github/workflows/build-and-e2e.yaml
oleg-kushniriov: it is temporal. removing

Ronkahn21 (approved):

PR #91, opened 2026-08-20
feat(charts): deploy opted-in tmpfs PageBroker

dfeigin-nv:
Adds the opt-in PageBroker sidecar, shared tmpfs staging, and control-socket mounts.

Validation: enabled and disabled Helm renders.

## Summary by CodeRabbit

* **New Features**
  * Added optional Page Broker support to the Snapshot deployment.
  * Configures Page Broker control and staging paths automatically.
  * Added settings for enabling Page Broker, selecting its image and pull policy, limiting concurrent requests, and configuring resource requests and limits.
  * Added configurable memory-backed staging storage with an optional size limit.
  * Page Broker is disabled by default and can be enabled through chart configuration.

### DIFF
--- a/charts/snapshot/values.yaml
+++ b/charts/snapshot/values.yaml
@@ -92,6 +92,22 @@ storage:
   oci:
     uri: ""
 
+pageBroker:
+  enabled: false
+  maxConcurrentRequests: 16
+  stagingSizeLimit: 64Gi
+  resources:
+    requests:
+      cpu: 8
+      memory: 64Gi
+    limits:
+      cpu: 32
+      memory: 72Gi
+  image:
+    repository: nvcr.io/nvidia/ai-dynamo/pagebroker
+    tag: 1.4.0
+    pullPolicy: Always
+
 # DaemonSet configuration for the checkpoint/restore agent
 daemonset:
   # Agent and nsrestore log level (trace, debug, info, warn, error)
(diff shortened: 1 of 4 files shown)


dfeigin-nv:
@coderabbitai full review

dfeigin-nv:
@coderabbitai full review

charts/snapshot/values.yaml
```diff
@@ -75,6 +75,22 @@ storage:
...
+  resources:
+    requests:
+      cpu: 8
+      memory: 64Gi
+    limits:
+      cpu: 32
+      memory: 72Gi
+  image:
+    repository: nvcr.io/nvidia/ai-dynamo/pagebroker
+    tag: 1.4.0
```
galletas1712, line 108: is this the right tag?

charts/snapshot/values.yaml
```diff
@@ -75,6 +75,22 @@ storage:
...
     uri: ""
 
+pageBroker:
+  enabled: false
+  maxConcurrentRequests: 16
+  stagingSizeLimit: 64Gi
+  resources:
+    requests:
+      cpu: 8
+      memory: 64Gi
```
galletas1712, line 102: since PB stages into tmpfs, wouldn't it count against this? or is it a different namespace? artifacts can be much larger than 64Gi.

charts/snapshot/values.yaml
dfeigin-nv: Nope. I changed it to align with Snapshot-agent's style

charts/snapshot/values.yaml
dfeigin-nv: It would. it's redundant. I'll remove it.
Regarding sizes - streamer keeps at most X (configurable) GB allocated and our custom storage engine should do the same, so this shouldn't be an issue for anyone using Custom Storage. 

I'll make it request 32GB and limit 256GB just in case. WDYT?

galletas1712: ah ok makes sense as long as the amount of staging is controlled. not sure if this is worth flagging in docs

galletas1712 (approved):
Approved as part of PageBroker PR stack review.

hhzhang16 (approved):

PR #92, opened 2026-08-20
fix(charts): allow runtime storage directory override

dfeigin-nv:
Adds the optional rootfs-diff mount used by the PageBroker-enabled agent image.

Validation: Helm lint and render.

## Summary by CodeRabbit

* **New Features**
  * Added an optional setting to configure the host directory for container runtime snapshots and overlay upper directories.
  * When omitted, the appropriate default directory is selected automatically based on the container runtime.

### DIFF
--- a/charts/snapshot/values.yaml
+++ b/charts/snapshot/values.yaml
@@ -58,6 +58,9 @@ crdUpgrade:
 runtime:
   type: containerd  # containerd | crio
   socketPath: ""
+  # Host directory containing container runtime snapshots and overlay upperdirs.
+  # Leave empty to use /var/lib/containerd (or /var/lib/containers for CRI-O).
+  storageDir: ""
 
 # Gates the OCP RBAC template and the DS required-scc annotation. Keep false
 # on vanilla Kubernetes (it references an OCP-only built-in SCC).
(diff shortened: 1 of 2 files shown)


dfeigin-nv:
@coderabbitai full review

dfeigin-nv:
@coderabbitai full review

charts/snapshot/values.yaml
```diff
@@ -41,6 +41,9 @@ crdUpgrade:
 runtime:
   type: containerd  # containerd | crio
   socketPath: ""
+  # Host directory containing container runtime snapshots and overlay upperdirs.
+  # Leave empty to use /var/lib/containerd (or /var/lib/containers for CRI-O).
+  storageDir: ""
```
galletas1712, line 63: not sure how this is related to PB. Does this belong in the stack? What is the purpose of it?

dfeigin-nv: This was needed for k3s: its containerd storage is under /var/lib/rancher/k3s/agent/containerd, not /var/lib/containerd. Without the override, we miss rootfs-diff.tar, and restores can be missing files
  created in the container. It is not PageBroker logic, but it is required for the PageBroker E2E there.

I'll move it to a separate PR and not in this stack.

dfeigin-nv:
Subset of #146

PR #96, opened 2026-08-21
feat: add the gated local CustomStorage POSIX path

hhzhang16:
## Summary

This ports and updates the CUDA CustomStorage work from ai-dynamo/dynamo#11584 in the canonical `ai-dynamo/snapshot` repository.

Stacked on #109, this wires the transfer-neutral helper core into Snapshot's existing one-target checkpoint/restore path and adds the Snapshot-local NIXL POSIX transfer implementation behind an explicit opt-in gate.

This implements the CUDA portion of the Snapshot-local checkpoint backend described by ai-dynamo/dynamo#13220. It does not include PageBroker, GMS, or ordered multi-target restore.

## Review focus

The highest-value review is the Go integration and rollout boundary: backend selection before CUDA mutation, manifest compatibility, restored-PID validation, timeout and unknown-outcome handling, and cleanup after partial restore failures. The NIXL/POSIX implementation in this PR is Snapshot-local. It does not define or constrain PageBroker's Control API, GPU engine, or backend selection.

## Main changes

- Build and ship the #109 helper with the NIXL POSIX transfer implementation.
- Add the Go daemon client, storage-mode selection, structured job files, and phase telemetry.
- Add `config.cudaCheckpoint.storageMode: legacy | posix`, defaulting to `legacy`.
- Require explicit `posix` opt-in for new CustomStorage artifacts and fail before target mutation when the helper lacks the required capability.
- Persist storage mode, extent metadata, device mapping, and launch-job state while preserving legacy-manifest compatibility.
- Restore from the artifact's recorded storage mode rather than silently falling back.
- Defer CUDA restore until CRIU recreates the process and Snapshot resolves and validates the current host PIDs.
- Prefetch extent files while CRIU restores the process. Prefetch is bounded and non-authoritative; the helper still performs the authoritative read after the restored PIDs are known.
- Support one assigned GPU with one or more CUDA-owning processes. TP2+ is rejected before CRIU or CUDA mutation in this first rollout.
- Add chart/image wiring, helper health checks, NIXL runtime packaging, validation, and pinned-memory guidance.
- Harden the shared legacy/POSIX path: preserve the source on confirmed pre-mutation failures, identity-check crash recovery, never replay an operation with an unknown daemon outcome, and terminate the restored placeholder after post-CRIU failures.

## Architecture

```mermaid
flowchart LR
    A["Snapshot agent"] -->|"bounded local RPC"| H["CUDA helper daemon"]
    H -->|"CUDA checkpoint + CustomStorage"| P["target PID"]
    H <-->|"NIXL POSIX"| PVC["Snapshot PVC"]
    C["CRIU"] <--> PVC
    A -->|"select before mutation"| M{"legacy or posix"}
    M --> H
```

The daemon transport and artifact backend are separate decisions. Backend selection happens before state-changing CUDA work and is persisted in the manifest. Restore obeys the recorded mode and does not silently switch after a POSIX artifact has been created.

## Rollout gate

The default:

```yaml
config:
  cudaCheckpoint:
    storageMode: legacy
    transferBufferCount: 4
    transferChunkBytes: 67108864
```

Enable new POSIX CustomStorage artifacts explicitly:

```yaml
config:
  cudaCheckpoint:
    storageMode: posix
    transferBufferCount: 4
    transferChunkBytes: 67108864
```

`legacy` remains the default artifact mode. The helper daemon is deployed for both modes, but the agent controller starts independently. Snapshot waits for the helper only when a checkpoint or restore has CUDA targets, before target mutation; CPU-only operations are not blocked by helper availability. `posix` is fail-closed, and published POSIX artifacts still require a compatible helper to restore. The agent and helper should be deployed and upgraded together.
## Matched end-to-end comparison

The matched experiment used Qwen3-0.6B on one physical B200 with the same DRA class, PVC, driver API, runtime, and Snapshot base. It used counterbalanced A-B-B-A blocks, discarded two warmups per block, and retained 15 measured restores per lane. All 30 measured restores completed, became Ready, and passed worker health without retries or exclusions.

The candidate explicitly used `storageMode=posix`, four 64 MiB buffers, and restored exactly 2,420,113,408 CUDA bytes in every sample.

| Metric | Pinned baseline `273cc46` main median / p95 | CustomStorage median / p95 | Median delta |
| --- | ---: | ---: | ---: |
| Agent restore operation | 5.652 / 6.401 s | **4.837 / 6.194 s** | **-14.4%** |
| Started-to-complete | 5.684 / 6.428 s | **4.866 / 6.225 s** | **-14.4%** |
| CRIU phase | 3.916 / 4.688 s | **3.002 / 4.295 s** | **-23.3%** |
| CUDA phase | 1.510 / 2.064 s | **1.407 / 1.990 s** | **-6.8%** |
| Pod create-to-Ready | 12.000 / 15.900 s | 12.000 / 14.300 s | 0.0% |

Candidate-only medians:

| Phase | Median |
| --- | ---: |
| Helper operation | 1.099 s |
| NIXL POSIX service | 0.435 s |
| CUDA process API | 0.517 s |
| Cleanup | 0.036 s |
| Effective transfer throughput | 3.864 GiB/s |

The deterministic bootstrap 95% interval for the agent-restore median delta was [-1.216 s, -0.260 s]. This matrix was measured on the pre-split candidate against `273cc46`, which was main at the time. The split and final hardening changes do not alter the extent-transfer algorithm or four-buffer configuration.
CustomStorage externalizes the 2.42 GB CUDA payload instead of leaving it in opaque process/driver state. Total durable artifact size differed by less than 0.1%, but the layout changed, so CRIU restored less opaque CUDA state while NIXL moved the external extents. This explains the observed CRIU-phase reduction without claiming that NIXL itself accelerates CRIU.

## #11584 context

#11584 is useful historical context, but it is not an apples-to-apples baseline:

| Result | #11584 one run | Matched pre-split candidate median |
| --- | ---: | ---: |
| CUDA artifact bytes | 1,730,150,400 | 2,420,113,408 |
| CUDA restore phase | 2.167 s | 1.407 s |
| External restore | 4.826 s | 4.837 s |

The measured candidate restores 39.9% more CUDA data, has a 35.1% lower CUDA phase, and is within 0.3% of #11584's external restore result. The table above is historical performance evidence, not a comparison against today's main. Final exact-image smoke timings are the restack regression check; a new matched A/B is required if they regress.

## Correctness validation

The final restack on #109 head `8d3d1a2` passed the full native Linux Go test and race matrices, go vet, Helm lint/render and invalid-configuration gates, and a production agent-image build.
Fresh smokes used Qwen3-0.6B on one DRA-exclusive B200. Both legacy and posix completed checkpoint and restore, became Ready, passed health, and returned exactly snapshot restore works through a local Dynamo frontend. These were correctness smokes; the counterbalanced 15-per-lane experiment above remains the performance comparison.

## Compatibility and limitations

- Legacy artifact creation remains the default, and legacy manifests still restore through the existing path.
- Unknown daemon outcomes are not replayed.
- Backend fallback cannot occur after state-changing CUDA work.
- The qualified profile is one B200 and Qwen3-0.6B; it does not qualify TP2+, GMS, ordered multi-target restore, PageBroker, SGLang, TRT-LLM, or cross-node restore.
- Planned helper restart requires draining live CustomStorage-restored targets. Unexpected helper restart safety is not yet qualified.

Supersedes and closes ai-dynamo/dynamo#11584

## Summary by CodeRabbit

* **New Features**
  * Added CUDA checkpoint and restore with configurable storage modes, transfer settings, and custom-storage support.
  * Added a dedicated CUDA checkpoint helper with health checks, secure communication, watchdogs, and asynchronous transfers.
  * Added NIXL-based POSIX transfers, process identity validation, artifact prefetching, and safer checkpoint cleanup.
  * Added Helm configuration for transfer buffers, chunk sizes, timeouts, storage, and helper resources.

* **Documentation**
  * Documented CUDA checkpoint configuration and helper behavior.

* **Tests**
  * Expanded coverage for daemon communication, manifests, transfers, configuration, cancellation, and process validation.


Ronkahn21:
Lets sync with @dfeigin-nv about this change, using is infra will relax the entire impl

agent/cmd/cuda-checkpoint-helper/main.cpp
```diff
@@ -0,0 +1,1229 @@
...
+        stderr,
+        "warning: retained CUDA primary context release failed with status %d "
+        "after operation acknowledgment\n",
+        static_cast<int>(primary_context_release_status));
+  }
+  if (primary_context_release_status != CUDA_SUCCESS) {
+    return {primary_context_release_status, operation};
+  }
+  return {CUDA_SUCCESS, operation};
+}
```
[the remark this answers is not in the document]
hhzhang16: Will be in a follow-up MR

agent/cmd/cuda-checkpoint-helper/transfer_engine.cpp
```diff
@@ -0,0 +1,772 @@
...
+  return success;
+}
+
+bool TransferPipeline(const std::vector<TransferChunk> &chunks,
+                      const std::vector<FileDescriptor> &files,
+                      std::vector<std::unique_ptr<TransferSlot>> *slots,
+                      nixlAgent *agent, const std::string &agent_name,
+                      CUdeviceptr device_ptr, CUstream stream,
+                      TransferOperation operation, TransferMetrics *metrics,
+                      TransferCancellation *cancellation, std::string *error) {
```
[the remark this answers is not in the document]
hhzhang16: Will be in a follow-up MR

dfeigin-nv (changes requested):
Blocking issues at a449ee774ef26d57e0f4a12f0bcd848e07f6df61:

1. CustomStorage prefetch is described as non-fatal, but restore waits synchronously for it after CRIU. Its worker uses blocking unix.Read and only checks cancellation between reads, so a stalled PVC/NFS read can indefinitely keep restored processes parked before the authoritative CUDA restore. Bound/detach prefetch so it cannot block the restore critical path.

2. The shipped restore deadline is 2h while the helper permits operations for 6h. When the controller deadline expires, the client closes the RPC and reports the state-changing CUDA operation outcome as unknown, while the daemon may still be restoring. Couple the deadlines and retain/resolve the operation outcome before failing the restore and killing the placeholder.

I traced the transfer cleanup, identity validation, and manifest paths; these are the two source-backed blockers I found.

dfeigin-nv (approved):
Withdrawn: supersedes my previous request-changes review.

dfeigin-nv (changes requested):
.

dfeigin-nv (commented):
Sorry about the previous comments They were written by mistake. The idea is to publish this and move logic into PageBroker? Will this require all users to compile NIXL? will it be part of the image? will all users need CUDA 13.4 to use snapshot?

hhzhang16:
@dfeigin-nv Yes, the intent here is to keep the driver-facing CustomStorage operation in a common place that can be used by both Snapshot-local and PageBroker. This MR also includes the Snapshot-local path. I separated the NIXL POSIX adapter from the common CustomStorage code so PageBroker can supply its own backend(s) without depending on NIXL. The standard/default Snapshot-local work will still use NIXL but PageBroker does not need it. Right now, since CustomStorage requires a driver exposing the CUDA 13.4 API, the implementation capability-detects for CUDA 13.4 before checkpointing; if it's not available, it'll fall back to the existing legacy CUDA mode.

agent/internal/cuda/cuda.go
```diff
@@ -328,57 +418,145 @@ func BuildDeviceMap(sourceUUIDs, targetUUIDs []string, log logr.Logger) (string,
...
-	for _, pid := range cudaPIDs {
-		if err := restoreProcess(ctx, pid, deviceMap, helperBinaryPath, log); err != nil {
+	for index, pid := range cudaPIDs {
+		processDir := ""
+		var selectedDevices []string
+		if storageMode == types.CUDAStorageModePOSIX {
+			processDir = customStorageProcessDir(checkpointDir, index)
+			selectedDevices = targetGPUUUIDs
+		}
+		if err := runner.run(ctx, helperAction{PID: pid, Action: actionRestore, DeviceMap: deviceMap, StorageMode: storageMode, StorageDir: processDir, JobFile: jobFile, GPUUUIDs: selectedDevices, Transfer: transferSettings}, log); err != nil {
```
dfeigin-nv: Is this fails, do we want to call something similar to terminateCUDAProcessesAfterCheckpointFailure?

agent/internal/cuda/daemon_client.go
```diff
@@ -0,0 +1,383 @@
...
+	daemonActionHealth     = uint16(0)
+	daemonActionCheckpoint = uint16(1)
+	daemonActionRestore    = uint16(2)
+	daemonActionLock       = uint16(3)
+	daemonActionUnlock     = uint16(4)
+
+	daemonResponseFatal           = uint32(1 << 0)
+	daemonCapabilityDeferredCUDA  = uint32(1 << 1)
+	daemonCapabilityCustomStorage = uint32(1 << 2)
+	daemonHealthWait              = 30 * time.Second
```
dfeigin-nv: The helper must be ready in 30 seconds, but the helper startup probe is much more permissive. Could we align them?

agent/internal/cuda/prefetch.go
```diff
@@ -0,0 +1,125 @@
...
+		return 0, fmt.Errorf("stat CUDA CustomStorage extent %s: %w", path, err)
+	}
+	if stat.Mode&unix.S_IFMT != unix.S_IFREG || stat.Size <= 0 {
+		return 0, fmt.Errorf("CUDA CustomStorage extent %s is not a nonempty regular file", path)
+	}
+	_ = unix.Fadvise(fd, 0, stat.Size, unix.FADV_WILLNEED)
+
+	var total int64
+	for {
+		if err := ctx.Err(); err != nil {
```
dfeigin-nv, line 110: Why do you need this loop? You already called fadvise.

agent/cmd/agent/main.go
```diff
@@ -52,6 +53,9 @@ func main() {
 
 	rootCtx, stop := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
 	defer stop()
+	if err := cuda.WaitForDaemon(rootCtx); err != nil {
```
oleg-kushniriov: The executor already skips all CUDA work when there are no CUDA PIDs (executor/checkpoint.go), so CPU-only pods never touch CUDA. But `WaitForDaemon` at startup means an unhealthy CUDA helper sidecar takes down all snapshot capability on the node, including for those workloads.

Is WaitForDaemon meant to gate all snapshot capability on the node, or only CUDA operations? 
If deliberate, please document it at the call site; if not, scope the wait to CUDA-only paths.

agent/internal/cuda/cuda.go
```diff
@@ -36,10 +40,96 @@ var podResourcesSocketPath = "/var/lib/kubelet/pod-resources/kubelet.sock"
...
+	ctx context.Context,
+	processes []snapshotruntime.ProcessDetails,
+	jobFile,
+	storageMode,
+	checkpointDir string,
+	gpuUUIDs []string,
+	transferSettings types.CUDATransferSettings,
+	log logr.Logger,
+) (CheckpointPhaseTimings, error) {
+	if err := acquireCUDAOperation(ctx); err != nil {
```
oleg-kushniriov: Wrap the `acquireCUDAOperation` and `validatedProcessIdentities` error paths in `&checkpointOperationError{targetMayBeMutated: false}`
otherwise an error that never mutated the target is still assumed to have mutated it.

Add a test that calls LockAndCheckpointProcessTreeValidated

agent/internal/cuda/cuda.go
```diff
@@ -328,57 +418,145 @@ func BuildDeviceMap(sourceUUIDs, targetUUIDs []string, log logr.Logger) (string,
...
+		identityValidatingRunner{
+			runner:     commandHelperActionRunner{},
+			procRoot:   snapshotruntime.HostProcPath,
+			identities: identities,
+		},
+		log,
+	)
+}
+
+func restoreAndUnlockProcessTree(
```
oleg-kushniriov, line 588: Is it intended that any unlock RPC failure now kills the placeholder, even when the process may actually be running fine? 
If yes, why was the old getState-based tolerance dropped instead of ported to the new protocol - is the daemon's unlock guaranteed idempotent?
If no, the daemon needs a state-query capability before this ships.

agent/internal/executor/restore.go
```diff
@@ -149,13 +157,99 @@ func Restore(ctx context.Context, rt snapshotruntime.Runtime, log logr.Logger, r
...
+			"bytes", prefetchResult.Bytes,
+			"service_duration", prefetchResult.Duration,
+			"overlapped_with_criu", true,
+		)
+	}
 	if result.CleanupError != nil {
 		cleanupErr = errors.Join(cleanupErr, result.CleanupError)
 	}
+	if len(result.DeferredCUDAProcesses) > 0 {
+		processTable, err := snapshotruntime.ReadProcessTable(snapshotruntime.HostProcPath)
```
oleg-kushniriov: On checkpoint failure, `terminateCUDAProcessesAfterCheckpointFailure` kills the target when its CUDA state is uncertain. Restore has no equivalent: a failure in `ReadProcessTable`, `ResolveHostProcessIdentityFromTable`, `ValidateProcessIdentity`, or `RestoreAndUnlockProcessTreeValidated` itself just returns an error from Restore, leaving the CRIU-resumed, possibly still-CUDA-locked process running with no termination or reconciliation attempt.

Suggestions: 
Add a symmetric cleanup path on the restore side- terminate the deferred/CUDA-locked processes using the same identity-revalidate-then-signal pattern as terminateCUDAProcessesAfterCheckpointFailure - covering both the new intermediate failure points and a failed `RestoreAndUnlockProcessTreeValidated` call, and join the cleanup error into the returned error.

agent/internal/cuda/daemon_client.go
```diff
@@ -0,0 +1,383 @@
...
+	// five minutes longer so the helper, not the caller, resolves an operation
+	// that has already reached CUDA.
+	daemonRPCTimeout = time.Hour + 5*time.Minute
+)
+
+var errDaemonUnavailable = errors.New("CUDA helper daemon unavailable")
+var errDaemonFatal = errors.New("CUDA helper daemon entered fatal state")
+var daemonSocketPath = types.CUDAHelperSocketPath
+
+func daemonRequest(
```
oleg-kushniriov: `daemonRequest` and `RestoreAndUnlockProcessTreeValidated` take `deviceMap`, `storageMode, storageDir/checkpointDir`, and `jobFile` as consecutive untyped string parameters. Swapping any two of them at a call site compiles cleanly and silently sends corrupted paths to a daemon that runs with elevated privileges over CUDA state.

Suggestion: Thread `helperAction` (or an equivalent small struct) all the way down into daemonRequest/runDaemonAction instead of re-expanding it at each layer boundary.

agent/internal/cuda/cuda.go
```diff
@@ -36,10 +40,96 @@ var podResourcesSocketPath = "/var/lib/kubelet/pod-resources/kubelet.sock"
 
 var gpuUUIDPattern = regexp.MustCompile(`^GPU-[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$`)
 
+// The node-local helper deliberately executes one CUDA operation at a time.
+// Hold the slot across each process-tree sequence so checkpoints/restores from
+// different workloads cannot interleave lock, transfer, and unlock requests.
+var cudaOperationSlot = make(chan struct{}, 1)
```
oleg-kushniriov, line 46: cudaOperationSlot is a single package-level channel shared by the whole agent process, not scoped per-GPU or per-daemon-socket. On a multi-GPU node, workloads on different GPUs can't checkpoint/restore concurrently - one slow transfer (up to daemonRPCTimeout = 1h5m) blocks every other workload's checkpoint/restore on that node, with no queue-depth or wait-time metric to see it happening.

Suggestions: If this mirrors a genuine single-threaded constraint in the daemon itself, say so explicitly (worth a line in the PR description/ops runbook - "multi-GPU nodes get no CUDA-checkpoint parallelism today") and add wait-time logging/metrics. Otherwise, scope the slot per-GPU or per-daemon-socket.

agent/internal/cuda/daemon_client.go
```diff
@@ -0,0 +1,383 @@
...
+	"time"
+
+	"github.com/go-logr/logr"
+
+	snapshotruntime "github.com/ai-dynamo/snapshot/agent/internal/runtime"
+	"github.com/ai-dynamo/snapshot/agent/internal/types"
+)
+
+const (
+	daemonProtocolMagic   = uint32(0x50484344)
```
oleg-kushniriov, line 25: The Go daemonRequest encoder and the C++ daemon_protocol.cpp decoder each define the full packet layout independently, and their tests are self-referential - Go tests encode with Go constants, C++ tests decode C++-built packets. A drift within protocol version 6 (a payload field reordered or a limit changed on one side only) would pass both suites and fail, or worse silently misparse, only at runtime.

Consider adding a shared golden-bytes fixture: a checked-in set of canonical request/response packets that the Go tests must produce byte-for-byte and the C++ tests must parse to known field values (and vice versa). That converts any one-sided edit into a test failure.

agent/internal/cuda/daemon_client.go
```diff
@@ -0,0 +1,383 @@
...
+	daemonMaxCgroup       = 4096
+	daemonMaxJobFile      = 4096
+
+	daemonActionHealth     = uint16(0)
+	daemonActionCheckpoint = uint16(1)
+	daemonActionRestore    = uint16(2)
+	daemonActionLock       = uint16(3)
+	daemonActionUnlock     = uint16(4)
+
+	daemonResponseFatal           = uint32(1 << 0)
```
oleg-kushniriov, line 40: On every failed CUDA checkpoint, captureCheckpoint must decide whether to SIGKILL the source process (device state possibly corrupt) or leave the customer's workload running (target never touched) - but the daemon, the only party that knows which one is true, doesn't say, so targetMayBeMutated is inferred from whether the connection succeeded. A daemon-set response flag would make the kill decision exact, and would also let the agent surface the outcome operators most need after a failed checkpoint - "is my source pod still alive?" - as a machine-readable reason instead of joined error text.

Please consider following:

- Add a rejectedBeforeMutation response flag (or reserved status-code range) that the daemon sets for every failure occurring before its first state-changing CUDA call, and drive `checkpointOperationError.targetMayBeMutated` from it instead of from connection-level heuristics.
- Propagate the resulting kill/spare outcome into the capture result and a SnapshotJob condition reason (e.g. `SourceLeftRunning` vs `SourceTerminated`) - today it exists only as joined error text, so an operator has no structured way to learn whether the source survived a failed checkpoint.

agent/internal/cuda/daemon_client.go
```diff
@@ -0,0 +1,383 @@
...
+	log.Info("CUDA custom-storage transfer succeeded", append(values, "output", output)...)
+	return nil
+}
+
+func daemonCapabilities(ctx context.Context) (uint32, error) {
+	packet, err := daemonRequest(0, "", "", "", "", "", nil, types.CUDATransferSettings{}, snapshotruntime.ProcessDetails{})
+	if err != nil {
+		return 0, err
+	}
+	status, flags, _, stderr, _, err := daemonRPC(ctx, daemonSocketPath+".health", packet)
```
oleg-kushniriov, line 375: WaitForDaemon and ValidateCUDAStorageMode both call `daemonCapabilities`, which only probes the `.health `socket, never `daemonSocketPath` itself. If the health listener is up but the operational listener is down or misconfigured (or vice versa), both the startup readiness check and the pre-mutation capability check pass while every real `lock/checkpoint/restore/unlock` RPC still fails.

Better either have the health/capability probe exercise the same socket path used for real operations, or add a code comment explaining why they're intentionally separate (likely to avoid queuing behind an in-flight operation) and what keeps the two listeners' liveness in sync.

agent/internal/cuda/cuda.go
oleg-kushniriov: `FilterProcesses` directly execs the on-disk `cuda-checkpoint-helper` binary (--get-restore-tid) instead of going through the new daemon RPC protocol, bypassing both cudaOperationSlot's serialization and the daemon's capability gating. It's unclear whether this read-only probe can race with an in-flight daemon-driven lock/checkpoint/restore for the same PID.

agent/internal/cuda/cuda.go
```diff
@@ -328,57 +418,145 @@ func BuildDeviceMap(sourceUUIDs, targetUUIDs []string, log logr.Logger) (string,
...
 	timings.TotalDuration = time.Since(start)
 
 	return timings, nil
 }
 
-// RestoreAndUnlockProcessTree restores and unlocks CUDA state for the given PIDs.
-// helperBinaryPath must be the absolute path to cuda-checkpoint-helper: DefaultHelperBinaryPath
-// on the agent, or filepath.Join(bundleDir, HelperBinaryName) inside the placeholder namespace.
-func RestoreAndUnlockProcessTree(ctx context.Context, cudaPIDs []int, deviceMap, helperBinaryPath string, log logr.Logger) (RestorePhaseTimings, error) {
+func customStorageProcessDir(checkpointDir string, processIndex int) string {
```
oleg-kushniriov: customStorageProcessDir assigns process-%04d directories purely by loop index over the PID list, independently at checkpoint time and restore time, with no PID/UUID tag inside the artifact tying a directory back to a process identity. A reordering anywhere in the chain that produces state.CUDAHostPIDs/state.CUDANSPIDs - a sort, a map iteration, a manifest migration - would silently restore one process's GPU memory into a different process, with no error raised.

Suggesting to tag each cuda-custom-storage/process-XXXX artifact with an identifying key (e.g. the checkpoint-time namespace PID) that the daemon or agent can cross-check against the resolved restore-time identity, instead of relying on array-index correlation alone.

agent/internal/executor/restore.go
```diff
@@ -149,13 +157,99 @@ func Restore(ctx context.Context, rt snapshotruntime.Runtime, log logr.Logger, r
...
+			if err := snapshotruntime.ValidateProcessIdentity(snapshotruntime.HostProcPath, process); err != nil {
+				return 0, fmt.Errorf("validate restored CUDA process identity: %w", err)
+			}
+			hostProcesses = append(hostProcesses, process)
+		}
+		cudaJobFile := ""
+		if stagedJobFile, err := cuda.JobFileFromCheckpoint(artifactPath); err != nil {
+			return 0, err
+		} else if stagedJobFile != "" {
+			cudaJobFile, err = cuda.HostJobFilePath(hostProcesses[0].OutermostPID)
```
oleg-kushniriov: In a multi-process restore, every daemon RPC resolves the CUDA job file through /host/proc/<first PID>/root/..., but only the request's own PID is identity-revalidated before each call. If the first process dies mid-restore (a locked process is still SIGKILL-able) and its PID is recycled by another container, the privileged daemon opens and mutates that path inside an unrelated container's rootfs; without PID reuse the milder outcome is a confusing ENOENT attributed to a later PID's unlock.

Sugesstion: Derive the job-file path per-request from `request.PID` - whose identity identityValidatingRunner already validates immediately before each call - or anchor it to the placeholder rootfs the way the checkpoint side anchors to state.RootFS.

agent/internal/cuda/shim.go
```diff
@@ -5,114 +5,134 @@ package cuda
...
+		ctx, request.PID, request.Action, request.DeviceMap, request.StorageMode,
+		request.StorageDir, request.JobFile, request.GPUUUIDs, request.Transfer, request.Identity, log,
 	)
-	return nil
 }
 
-func normalizeProcessGroupKillError(err error) error {
-	if errors.Is(err, syscall.ESRCH) {
-		return os.ErrProcessDone
+func (r identityValidatingRunner) run(
```
oleg-kushniriov, line 124: `identityValidatingRunner.run` has no direct test - only its ValidateProcessIdentity dependency is tested. The "PID not found in identity map" and "mismatch, don't call the driver" branches are unexercised.
No test calls `LockAndCheckpointProcessTreeValidated` with a failing acquireCUDAOperation/validatedProcessIdentities and asserts the resulting classification - this gap is why the pre-mutation misclassification Major finding wasn't caught.
`WaitForDaemon's` retry/timeout loop is untested (only daemonCapabilities is exercised) - no coverage for "succeeds on a later retry" or "gives up after daemonHealthWait and wraps lastErr."

 Add table-driven tests for each.

agent/internal/cuda/daemon_client.go
oleg-kushniriov: `daemonRequest` validates most arguments but misses five invariants the daemon enforces, so a violating request fails at operation time - in the checkpoint-with-deviceMap and case-variant-UUID cases, after the target is already locked, converting a client-side programming error into a targetMayBeMutated failure that kills the workload. All are latent today (call sites happen to construct valid requests), but the encoder is the layer meant to make them unrepresentable.

Suggestion: Mirror the daemon's parse-time invariants in daemonRequest: reject checkpoint+deviceMap, sanitize the three missed health fields, validate transfer settings non-zero/positive before encoding, canonicalize UUIDs before the duplicate check, and reject NUL bytes in all strings.

oleg-kushniriov (commented):
Reviewed the Go side mostly

1. Restore lost safety nets the checkpoint side has - no cleanup path when a restore fails after CRIU has resumed the process, and the old tolerance for a benign unlock error was dropped without a replacement in the new protocol.
2. Timeout and startup architecture - the daemon's watchdog doesn't actually bound the CUDA driver calls the client's 1h5m design assumes it does; the 1h5m caller-budget floor conflicts with the fast-restore goal; and WaitForDaemon couples all snapshot capability on a node (including CPU-only workloads) to the CUDA helper's health.
3. Go<->C++ contract divergences - the two sides enforce different invariants (encoder vs parser validation, per-operation pinned cap only in C++ and post-mutation, response-overflow turning a successful restore into a reported failure), and the restore path anchors its job-file path to the first PID's /proc root without revalidating that anchor for the other PIDs' calls.

agent/internal/cuda/cuda.go
hhzhang16: True, added

agent/internal/cuda/daemon_client.go
hhzhang16: Made it up to 5 minutes to match the startup probe

agent/internal/cuda/prefetch.go
hhzhang16: Fadvise (FADV_WILLNEED) is only an advisory hint and it can return before the pages/file are actually resident. This loop faults the full extent into the page cache during CRIU restore. I added a comment explaining this

agent/cmd/agent/main.go
hhzhang16: This is a valid conern, I can scope the wait to CUDA checkpoint/restore paths

agent/internal/cuda/cuda.go
hhzhang16: Done

agent/internal/cuda/cuda.go
hhzhang16: Yeah this was intentional; since it's a newly restored placeholder, I thought that terminating it would be safer than reporting an ambiguous successful restore

agent/internal/cuda/daemon_client.go
hhzhang16: Done

agent/internal/cuda/daemon_client.go
hhzhang16: They're intentionally separate, added a comment

agent/internal/cuda/cuda.go
hhzhang16: Good catch, thanks!

agent/internal/cuda/cuda.go
hhzhang16: Also: the old `getState` tolerance wasn't ported because the daemon protocol doesn't expose an equivalent yet. At least for now, I'd prefer fail-closed termination, something like state-query recovery can be a follow-up

agent/cmd/agent/main.go
hhzhang16: Fixed

agent/internal/cuda/prefetch.go
hhzhang16: I ended up removing it in #121

agent/internal/executor/restore.go
hhzhang16: Done, with focused tests

hhzhang16:

PR #98, opened 2026-08-23
fix(agent): bundle a statically linked tar for restore

oleg-kushniriov:
### Problem

`ApplyRootfsDiff` extracts the rootfs diff inside the placeholder container by invoking `tar` through PATH. Until now the bundle shipped the agent image's dynamically linked tar, whose library closure is built from the agent base image (Ubuntu 24.04, glibc 2.39). Because tar resolves its ELF loader and glibc from the placeholder image, any placeholder based on an older userspace (e.g. Ubuntu 22.04) fails at restore with `version 'GLIBC_2.38' not found` — or silently falls back to the placeholder's own tar.

### Fix

Build GNU tar 1.35 statically against musl in a dedicated `tar-builder` stage and bundle it, so rootfs-diff extraction no longer depends on any userspace library in the placeholder image:

- New Alpine-based `tar-builder` stage downloads a SHA256-pinned GNU tar source tarball, builds it with `LDFLAGS=-static`, and verifies the result at build time: `readelf` asserts there is no ELF interpreter and no `NEEDED` entries, and a smoke test exercises the exact restore invocation.
- `ApplyRootfsDiff` now takes an explicit tar binary path and rejects a non-absolute one, so extraction can never fall back to the placeholder's tar via PATH; `nsrestore` passes the bundle path.
- Extraction now restores the xattrs the capture side archives (`--xattrs`), scoped to `user.*` and `security.*` — the archive is cut from the overlay upperdir and may carry `trusted.overlay.*` entries that must never be re-applied through the restored container's overlay mount. The smoke test sets a `user.*` xattr and asserts it survives the round trip, failing the build if xattr support was compiled out of the static tar.
- The bundle `ldd` closure check drops tar (static binaries have no shared deps to resolve).
- The glibc-compatibility constraint documented for `BASE_IMAGE` now applies only to CRIU, `ip`, and the CUDA tools — the restore tar is exempt.

### Compliance

- GNU tar source tarball is staged under `/legal/source/gnu-tar/` and documented in `README.source.txt`.
- The GPL license text (`COPYING`) is folded into the generated attribution file.
- The `attr` package added to the `tar-builder` stage is build-time test tooling only (`setfattr`/`getfattr` for the smoke test); it is not redistributed and does not change the image's license surface.

### Testing

- Unit tests updated for the new `ApplyRootfsDiff` signature, plus cases asserting the provided tar path is actually the one invoked and that a relative tar path is rejected.
- Build-time smoke test in the Dockerfile validates static linkage, the restore extraction flags, and xattr round-trip.

## Summary by CodeRabbit

* **Reliability**
  * Restore operations now use a bundled, statically compiled archive tool for consistent extraction across environments.
  * Restore no longer depends on the system’s dynamically linked archive utility.

* **Bug Fixes**
  * Improved handling of extended attributes during root filesystem restoration.
  * Clearer errors are provided when the required restore archive tool is missing or misconfigured.

* **Documentation**
  * Updated compliance and compatibility documentation for the bundled archive tool and restore requirements.

### DIFF
--- a/agent/Dockerfile
+++ b/agent/Dockerfile
@@ -103,6 +103,7 @@ ARG GNU_TAR_VERSION
 ARG GNU_TAR_SHA256
 
 RUN apk add --no-cache \
+    attr \
     build-base \
     ca-certificates \
     curl \
@@ -143,11 +144,14 @@ RUN set -eu; \
     echo preserved > /tmp/tar-smoke/target/existing; \
     echo replaced > /tmp/tar-smoke/source/existing; \
     echo created > /tmp/tar-smoke/source/created; \
-    /out/tar -C /tmp/tar-smoke/source -cf /tmp/tar-smoke/archive.tar .; \
+    setfattr -n user.snapshot-smoke -v smoke /tmp/tar-smoke/source/created; \
+    /out/tar --xattrs -C /tmp/tar-smoke/source -cf /tmp/tar-smoke/archive.tar .; \
     /out/tar --skip-old-files --blocking-factor=2048 \
+      --xattrs --xattrs-include='user.*' --xattrs-include='security.*' \
       -C /tmp/tar-smoke/target -xf /tmp/tar-smoke/archive.tar; \
     grep -qx preserved /tmp/tar-smoke/target/existing; \
-    grep -qx created /tmp/tar-smoke/target/created
+    grep -qx created /tmp/tar-smoke/target/created; \
+    test "$(getfattr --only-values -n user.snapshot-smoke /tmp/tar-smoke/target/created)" = smoke
 
 # =============================================================================
 # Stage: CUDA checkpoint helper builder
--- a/agent/Dockerfile
+++ b/agent/Dockerfile
@@ -111,47 +111,13 @@ RUN apk add --no-cache \
 
 WORKDIR /tmp
 
-RUN set -eu; \
-    curl -fsSLo gnu-tar.tar.xz \
-      "https://ftp.gnu.org/gnu/tar/tar-${GNU_TAR_VERSION}.tar.xz"; \
-    echo "${GNU_TAR_SHA256}  gnu-tar.tar.xz" | sha256sum -c -; \
-    cp gnu-tar.tar.xz /gnu-tar-source.tar.xz; \
-    xz -d gnu-tar.tar.xz; \
-    tar -xf gnu-tar.tar
+COPY scripts/fetch-gnu-tar.sh scripts/build-gnu-tar.sh /tmp/scripts/
+
+RUN sh /tmp/scripts/fetch-gnu-tar.sh "${GNU_TAR_VERSION}" "${GNU_TAR_SHA256}"
 
 WORKDIR /tmp/tar-${GNU_TAR_VERSION}
 
-RUN set -eu; \
-    FORCE_UNSAFE_CONFIGURE=1 LDFLAGS=-static ./configure \
-      --disable-nls \
-      --without-posix-acls \
-      --without-selinux; \
-    make -j"$(nproc)"; \
-    mkdir -p /out; \
-    cp src/tar /out/tar; \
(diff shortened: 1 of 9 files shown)


Ronkahn21 (changes requested):
Tar is GPL adding it to the repo is problematic
I would want to get the legal approval before continue

oleg-kushniriov:
> Tar is GPL adding it to the repo is problematic I would want to get the legal approval before continue

The agent exec's tar as a separate process (exec.Command). It doesn't link against GPL code, so GPLv3's copyleft doesn't extend to your Go code. Shipping a GPL binary next to Apache-2.0 binaries in one image is "mere aggregation" - the same reason every container image can ship bash/coreutils without becoming GPL.

I will see for another solution that does not bundle tar

agent/internal/executor/nsrestore.go
```diff
@@ -127,7 +127,8 @@ func executeRestore(
 	timings = &nsrestorePhaseTimings{}
 
 	overlayStart := time.Now()
-	if err := snapshotruntime.ApplyRootfsDiff(opts.CheckpointPath, "/", log); err != nil {
+	tarBinary := filepath.Join(opts.BundleDir, "tar")
```
leeZardNav: nit: add the "tar" to a const, similar to the `cuda.HelperBinaryName`

agent/Dockerfile
```diff
@@ -87,6 +89,70 @@ RUN CGO_ENABLED=0 GOOS=${TARGETOS} GOARCH=${TARGETARCH} go build -ldflags="-w -s
...
+    curl -fsSLo gnu-tar.tar.xz \
+      "https://ftp.gnu.org/gnu/tar/tar-${GNU_TAR_VERSION}.tar.xz"; \
+    echo "${GNU_TAR_SHA256}  gnu-tar.tar.xz" | sha256sum -c -; \
+    cp gnu-tar.tar.xz /gnu-tar-source.tar.xz; \
+    xz -d gnu-tar.tar.xz; \
+    tar -xf gnu-tar.tar
+
+WORKDIR /tmp/tar-${GNU_TAR_VERSION}
+
+RUN set -eu; \
```
leeZardNav: in other places in the dockerfile we call bash scripts that are in files. it is usally easier to read and test ourself when it's in a separate file
would you consider moving the 2 run blocks to bash files?

agent/internal/runtime/overlay.go
```diff
@@ -182,12 +188,20 @@ func ApplyRootfsDiff(checkpointPath, targetRoot string, log logr.Logger) error {
...
+	//
+	// The capture side archives with --xattrs; restore them, but only the user
+	// and security namespaces. The archive is cut from the overlay upperdir, so
+	// it can carry trusted.overlay.* entries that must never be re-applied
+	// through the restored container's overlay mount. These flags are mirrored
+	// by the tar-builder smoke test in the Dockerfile — keep them in sync.
 	log.Info("Applying rootfs diff", "target", targetRoot, "bytes", info.Size())
-	cmd := exec.Command("tar", "--skip-old-files", "--blocking-factor=2048", "-C", targetRoot, "-xf", localPath)
+	cmd := exec.Command(tarBinary, "--skip-old-files", "--blocking-factor=2048",
+		"--xattrs", "--xattrs-include=user.*", "--xattrs-include=security.*",
```
leeZardNav, line 204: i'm not sure i understand the need for added xattrs... i checked what i know and these additions aren't useful in this case from what i can gather
do you know which of them is used/useful in this case?

oleg-kushniriov: We restore` user.* xattrs` because the capture side archives them from the overlay upperdir with `--xattrs`, and they can carry app-set metadata that needs to survive restore. security.* is included defensively - SELinux/AppArmor labels can end up in the upperdir on some placeholders, and we don't want those silently dropped, even though the smoke test currently only exercises `user.*.`

PR #101, opened 2026-08-23
feat!: remove leave-running and complete SnapshotJob from the capture result

oleg-kushniriov:
### What

A checkpoint now always terminates the target container — leave-running is removed as a mode — and the control plane treats the target's death as the expected outcome of a successful capture instead of a failure. The design's two-stage completion gate is preserved: helper containers (e.g. a GMS saver writing weights concurrently with the dump) must finish before the SnapshotJob completes and cleans up.

Stacked on #80; #81 (e2e) is retargeted onto this branch, and its workloads that wait for the `snapshot-complete` sentinel need updating.

### Why

Running the source after checkpoint was never actually supported: a CUDA-locked source is left frozen and never resumed (see `docs/cuda-source-not-resumed-after-checkpoint.md`), and no CRD exposes a `leaveRunning` field. Meanwhile the old completion contract required the source Job to report `Complete=True`, which is impossible once the dump kills the workload — every successful capture would end as `SnapshotJob Failed/JobFailed`.

### Operator: two-stage completion gate, adapted to always-kill

- `PodSnapshot Ready=True` sets `Captured=True` and `Completed=False/WaitingForPodCompletion`; the SnapshotJob then waits for the source Job to reach a terminal state before completing and deleting the Job — cleanup must not kill helper containers mid-work.
- The killed target makes the Job report `Failed` on every successful capture, so "finished cleanly" is now verified from the pod's container statuses: every non-target container must exit 0 (`Completed=True/JobCompleted`); a helper exiting non-zero fails the run with `JobFailed` while `Captured` stays True and the artifact remains usable.
- An explicit `DeadlineExceeded` still fails the run even after a successful capture — the deadline bounds the whole source lifecycle, helpers included.
- A capture failure keeps its own reason instead of being overridden by a raced Job failure; a terminal source Job with a pending capture waits for the capture result (bounded backstop requeue) instead of failing on event ordering; a deleted Job stays terminal (`JobDeleted`) because helper completion can no longer be verified.
- The PodSnapshot reconciler defers the `SourcePodGone` failure by a grace window (5m from the pod's terminal timestamp) so a killed source cannot race the agent's content `Ready` write; expiry keeps an abandoned capture bounded if the agent never publishes a result.

### Agent: always-kill enforcement (b70efb1)

- `criu.BuildDumpOptions` hardcodes `LeaveRunning=false`; the knob is removed from `CRIUSettings`, `values.yaml`, and the ConfigMap template.
- The `snapshot-complete` release machinery is deleted — the target is dead by design, so `Ready` is written without needing a live PID.
- Recovery ordering fixed for the dead-source-is-success world:
  - source identity (UID) is validated separately from liveness, so a same-named replacement pod can never validate another incarnation's artifact;
  - a committed artifact is recovered to `Ready` before any liveness-derived failure gate (container exit, pod phase, PID resolution) — closing the crash window between artifact commit and the `Ready` write;
  - the pre-bind gate resync defers to an in-flight capture or a committed artifact instead of terminally failing the work order mid-dump.

### Breaking change

Direct PodSnapshot sources are now terminated by the capture. `config.criu.leaveRunning` (chart default was `true`) is removed; agent config parsing is non-strict, so old ConfigMaps still containing the key are ignored, not rejected. Workloads polling for `snapshot-complete` are killed instead of released (`SnapshotCompleteFile` is deprecated).

Ordering note: the operator commit is deliberately first — it is safe with old agents, while the agent commit requires the new operator semantics. A single rolling upgrade has a transient window where a new agent kills a source under an old operator, which would misreport `JobFailed`.

### Testing

- Operator and API suites pass; new coverage for capture-vs-Job arbitration (19-case table), the grace window, and JobDeleted-vs-Ready.
- Full agent suite passes on Linux; new coverage for artifact recovery preceding liveness gates, dead-source-without-artifact still failing, gate in-flight/artifact deference, and no-kill on Failed-before-Ready.
- `helm lint` passes; the rendered chart contains no `leaveRunning`.

## Summary by CodeRabbit

* **New Features**
  * Snapshot completion now requires durable capture, source workload completion, and successful helper-container completion.
  * Added clearer completion, waiting, and failure status reasons.

* **Bug Fixes**
  * Improved recovery when source workloads terminate or disappear after capture.
  * Preserved capture-specific failures and avoided unnecessary retries or process termination.
  * Added safeguards for terminal, missing, and conflicting workload states.

* **Documentation**
  * Clarified that checkpointing terminates the source process.
  * Deprecated the completion sentinel and removed its configuration option.


agent/internal/controller/podsnapshotcontent.go
```diff
@@ -86,7 +86,30 @@ func (w *NodeController) reconcilePodSnapshotContent(ctx context.Context, name s
 		logger.Error(err, "Failed to get source pod", "pod", key.String())
 		return
 	}
-	if reason, msg := classifySourcePod(content, pod); reason != "" {
+	if reason, msg := classifySourcePodIdentity(content, pod); reason != "" {
+		if err := w.setSnapshotContentFailed(ctx, content, reason, errors.New(msg)); err != nil {
+			logger.Error(err, "Failed to write PodSnapshotContent failed status", "content", content.Name)
+		}
+		return
```
hhzhang16: nit: this is repeated from lines 113-115

agent/internal/controller/podsnapshotcontent.go
hhzhang16, line 86: What if there's the case where the checkpoint was already committed but there's no source pod due to some failure? Should this also check for the artifact?

operator/internal/controller/snapshotjob_reconciler.go
```diff
@@ -216,22 +227,104 @@ func (r *SnapshotJobReconciler) reconcilePodSnapshotResources(ctx context.Contex
...
+		}
+		if terminated.ExitCode != 0 {
+			return &snapshotJobFailure{
+				reason: snapshotv1alpha1.ReasonJobFailed,
+				cause: fmt.Errorf("helper container %q exited with code %d after the capture succeeded",
+					status.Name, terminated.ExitCode),
+			}, false, nil
+		}
+	}
+	return nil, false, nil
```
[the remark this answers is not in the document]
hhzhang16: Was this fixed/reintroduced?

hhzhang16 (commented):
Have the documentation been updated with this new change logic?

To be completely honest though I think we should consider separating the required always-terminate fix from the helper policy being added here. Snapshot says it owns capture mechanics, while consumers own failure policy. This PR instead makes every non-target container part of SnapshotJob success: a valid captured artifact becomes an aggregate failure if any other container exits non-zero, and an ordinary long-running sidecar keeps the SnapshotJob waiting until its deadline.

agent/internal/controller/podsnapshotcontent.go
```diff
@@ -86,7 +86,30 @@ func (w *NodeController) reconcilePodSnapshotContent(ctx context.Context, name s
...
+	}
+	if reason, msg := classifySourcePodLiveness(pod); reason != "" {
+		// The dump terminates the source process, so a terminal pod may mean a
+		// capture just succeeded: while the capture goroutine holds the
+		// in-flight guard it owns the outcome, and a committed artifact only
+		// needs its Ready write. Only a dead source with neither is a failure.
+		if contentUID := strings.TrimSpace(string(content.UID)); contentUID != "" {
```
Ronkahn21: redundent check UID will be always field no need to trim

agent/internal/controller/podsnapshotcontent.go
```diff
@@ -174,10 +191,29 @@ func (w *NodeController) reconcileSourcePod(ctx context.Context, pod *corev1.Pod
...
+		err := w.setSnapshotContentFailed(ctx, content, reason, errors.New(msg))
+		w.removeCaptureEligibleLabel(ctx, pod)
+		return err
+	}
+
+	// Artifact recovery must precede every liveness-derived classification below: the dump
+	// terminates the source process, so after a success the pod is failed, the target container
+	// is terminated, and its PID is unresolvable — all expected, none of it a capture failure.
+	// The artifact dir exists only after the executor's atomic rename, so its presence means a
+	// completed dump whose Ready write did not land yet.
```
Ronkahn21: this is not true the procsses is just suspended, it does not same thing as terminated

agent/internal/controller/podsnapshotcontent.go
```diff
@@ -174,10 +191,29 @@ func (w *NodeController) reconcileSourcePod(ctx context.Context, pod *corev1.Pod
...
+	artifactPath, err := nsmount.ResolveArtifactPath(w.config.Storage.BasePath, contentUID, containerName)
```
Ronkahn21, line 249: it do means it need to berestored

agent/internal/criu/dump.go
```diff
@@ -49,6 +49,11 @@ func BuildDumpOptions(
 		Pid:     proto.Int32(int32(state.PID)),
 		Root:    proto.String(state.RootFS),
 		LogFile: proto.String(dumpLogFilename),
+		// The dump always terminates the source process. Leaving it running is
+		// not a supported mode (a CUDA-locked source is never resumed), and the
+		// operator treats the source's death as the expected outcome of a
+		// successful capture.
+		LeaveRunning: proto.Bool(false),
```
Ronkahn21, line 56:
```suggestion
		resumeAfterCapture: proto.Bool(false),
```

operator/internal/controller/snapshotjob_reconciler.go
```diff
@@ -216,22 +227,104 @@ func (r *SnapshotJobReconciler) reconcilePodSnapshotResources(ctx context.Contex
...
 	}
-	observed.failure = snapshotJobTerminalFailure(observed.job, observed.podSnapshot)
-	return observed, ctrl.Result{}, nil
+}
+
+// classifyHelperContainers verifies that every non-target container of the
+// failed source Job exited 0. The target container's exit is ignored — the
+// checkpoint terminates it by design, which is also why the source Job reports
+// Failed on every successful capture.
+func (r *SnapshotJobReconciler) classifyHelperContainers(ctx context.Context, sj *snapshotv1alpha1.SnapshotJob, job *batchv1.Job) (failure *snapshotJobFailure, helpersStillRunning bool, err error) {
```
Ronkahn21, line 298: can you explain this better, do we should care about it

operator/internal/controller/snapshotjob_reconciler.go
```diff
@@ -256,6 +350,29 @@ func deriveSnapshotJobStatus(sj *snapshotv1alpha1.SnapshotJob, observed snapshot
...
+		{snapshotv1alpha1.SnapshotJobConditionCompleted, snapshotv1alpha1.ReasonCaptureInProgress, "waiting for the checkpoint capture to complete"},
+		{snapshotv1alpha1.SnapshotJobConditionFailed, snapshotv1alpha1.ReasonNoFailure, "no failure observed"},
+	}
+	for _, def := range defaults {
+		if meta.FindStatusCondition(next.Status.Conditions, def.conditionType) == nil {
+			setCondition(next, def.conditionType, metav1.ConditionFalse, reconciliationTime, def.reason, def.message)
+		}
+	}
+}
+
```
Ronkahn21, line 412: this function is kind of weird, I understand what it does, and still
by the way condition dont need to have always reason and msg if it does not make sense 
so just False condition could be fine

oleg-kushniriov:
> Have the documentation been updated with this new change logic?
> 
> To be completely honest though I think we should consider separating the required always-terminate fix from the helper policy being added here. Snapshot says it owns capture mechanics, while consumers own failure policy. This PR instead makes every non-target container part of SnapshotJob success: a valid captured artifact becomes an aggregate failure if any other container exits non-zero, and an ordinary long-running sidecar keeps the SnapshotJob waiting until its deadline.

We plan to add this as configuration flag - currently implementing  leaveRunning:true requires fixing another bug. Not urgent for current release

agent/internal/controller/podsnapshotcontent.go
oleg-kushniriov: CRIU with leave_running=false kills the tree after dump

agent/internal/criu/dump.go
oleg-kushniriov: it's the upstream go-criu protobuf field; the suggestion won't compile

operator/internal/controller/snapshotjob_reconciler.go
oleg-kushniriov: metav1.Condition.Reason is required by apimachinery validation

agent/internal/controller/podsnapshotcontent.go
```diff
@@ -100,6 +109,43 @@ func (w *NodeController) reconcilePodSnapshotContent(ctx context.Context, name s
...
+	// alias onto a shared path (same invariant as MissingContentUID).
+	contentUID := string(content.UID)
+	if contentUID == "" {
+		return false
+	}
+	containerName, err := singleTargetContainer(content)
+	if err != nil {
+		return false
+	}
+	if w.checkpointInFlight(contentUID + "/" + containerName) {
```
julienmancuso, line 127: [P1] Honor the shared Lease before declaring the capture failed

This guard is process-local. If two agent instances overlap, holder A can own the Kubernetes Lease and be between terminating the source and atomically renaming the artifact, while B sees neither its own in-flight entry nor the artifact and returns false here. B then writes `SourcePodGone`/`SourcePodNotFound`; because Failed is sticky, A's successful dump cannot mark Ready. The same window exists in `reconcileSourcePod`, where liveness failure is handled before `acquireLease`. Please treat an unexpired Lease held by another agent as in flight, or arbitrate the Lease before any terminal liveness write, and cover the killed-source-before-artifact-commit window with two controller instances.

julienmancuso (changes requested):
Requesting changes for a cross-agent capture arbitration race. Details inline.

agent/internal/controller/podsnapshotcontent.go
oleg-kushniriov: The gate's pod-NotFound branch now checks for an in-flight capture or a committed artifact (recovering it to Ready) before failing with SourcePodNotFound, plus two tests covering both cases.

julienmancuso (approved):

Ronkahn21 (approved):

Issue #27, opened 2026-07-27
De-brand DYN_* environment variables to a vendor-neutral prefix


Ronkahn21:
The snapshot repo is a generic checkpoint/restore stack, but two runtime-contract environment variables still carry the Dynamo `DYN_` brand:

- `DYN_SNAPSHOT_CONTROL_DIR` — the `SnapshotControlDirEnv` contract constant in `api/v1alpha1/constants.go` (shared operator/agent/workload contract)
- `DYN_SNAPSHOT_RESTORE_STANDBY` — `RestoreStandbyModeEnv` in `operator/internal/protocol/restore.go`

Plus comment references in `operator/internal/protocol/{restore,checkpoint,control_volume}.go` and `agent/Dockerfile`.

### Proposal
Rename to a vendor-neutral prefix (suggest `SNAPSHOT_`): `SNAPSHOT_CONTROL_DIR`, `SNAPSHOT_RESTORE_STANDBY`.

### Why it needs care (not a mechanical rename)
This is a breaking, coordinated contract change. Both the writer (operator injects the var into checkpoint/restore pod specs) and the reader (agent) must move together, and any external workload already setting `DYN_SNAPSHOT_*` must switch. Because `SnapshotControlDirEnv` is exported from the `api` module, the rename should land before any external consumer pins an `api` release, so the published constant carries the neutral name and consumers inherit it by import.

### Scope
- [ ] Rename the `api` constant and any operator/agent constants
- [ ] Update every pod-spec injection site and the agent reader together
- [ ] Update comment references and `agent/Dockerfile`
- [ ] Document the contract change in release notes

Context: surfaced while planning the Dynamo -> standalone-snapshot migration. Deliberately kept out of that migration scope and tracked here.

PR #103, opened 2026-08-24
test(e2e): de-brand restore standby env

caozhuozi:
## Summary
- set SNAPSHOT_RESTORE_STANDBY in restore e2e fixtures while keeping the legacy DYN_SNAPSHOT_RESTORE_STANDBY alias
- update the placeholder image comment to reference the neutral restore standby env

Part of #27

## Summary by CodeRabbit

* **Documentation**
  * Updated restore-standby configuration guidance to use the current environment variable.

* **Bug Fixes**
  * Improved restore pod configurations by including the supported restore-standby setting alongside the existing configuration.

### DIFF
--- a/agent/Dockerfile
+++ b/agent/Dockerfile
@@ -341,7 +341,7 @@ ENTRYPOINT ["/usr/local/bin/snapshot-agent"]
 # PATH search — and so must this placeholder image (same BASE_IMAGE).
 #
 # Backend entrypoints run a restore standby path when the operator
-# injects DYN_SNAPSHOT_RESTORE_STANDBY=1, then sleep while the
+# injects SNAPSHOT_RESTORE_STANDBY=1, then sleep while the
 # snapshot-agent restores into the container.
 # =============================================================================
 FROM ${BASE_IMAGE} AS placeholder
--- a/e2e/snapshot_e2e/workloads.py
+++ b/e2e/snapshot_e2e/workloads.py
@@ -144,6 +144,7 @@ def restore_pod(
         }
     }
     spec["containers"][0]["env"] = [
+        {"name": "SNAPSHOT_RESTORE_STANDBY", "value": "1"},
         {"name": "DYN_SNAPSHOT_RESTORE_STANDBY", "value": "1"},
         {"name": "SNAPSHOT_CONTROL_DIR", "value": CONTROL_DIR},
         {"name": RESTORE_TOKEN_ENV, "value": run.restore_token},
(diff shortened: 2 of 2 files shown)


Ronkahn21 (commented):
Hey thanks for the contribution,  fix the conflict

caozhuozi:
@Ronkahn21 conflicts are resolved. Could you please take another look?

julienmancuso:
@caozhuozi — apologies, there's overlap here I should have caught sooner. #131
added `RestoreStandbyModeEnv` and `LegacyRestoreStandbyModeEnv` to
`api/v1alpha1/constants.go` in d051f17, duplicating the constants you'd already
proposed here two days earlier. That was my miss.

On the design question this PR raises — who sets the variable — we ended up
somewhere different after discussion on #131:

Snapshot publishes `SNAPSHOT_RESTORE_STANDBY` and the deprecated `DYN_` alias as
public constants, but the generic restore-Pod helper injects neither. The
reasoning is that Snapshot doesn't consume this variable. Grepping this repo
finds nothing that reads it — only the two E2E fixture call sites and a
Dockerfile comment. It's the application entrypoint that implements standby
semantics, and Snapshot can't verify that any given image does. Injecting it
universally would look like a guarantee we can't actually make.

The requirement Snapshot really has is that the destination container stays
alive and inert until the agent replaces its placeholder process. The env var is
one way to satisfy that; an inert command override is another. So the mechanism
belongs to the integration — Dynamo injects the variable for the entrypoints it
knows implement it.

That means the injection changes in `operator/internal/protocol/restore.go`
don't have a home: #131 deletes that file and its tests.

The rest of this PR, though, is work #131 needs and doesn't have:

- `e2e/snapshot_e2e/workloads.py` — #131 still emits
  `DYN_SNAPSHOT_RESTORE_STANDBY` at both call sites, so it publishes the neutral
  name as canonical while our own reference manifests use the deprecated one
- `agent/Dockerfile` — #131 doesn't touch it at all

Would you be up for rebasing this down to just those two? Drop the constants
(already in #131) and the protocol injection, keep the de-branding. That leaves
a small, clean PR under your name that fixes a real inconsistency in mine, and
#27 closes once both land.

Two notes if you do:

**Set both names in the fixture rather than renaming.** Nothing in this repo
reads the variable, so we can't tell from here which name the `BASE_IMAGE`
backend entrypoint honors. A straight rename bets that it already knows the new
one — and if it doesn't, the container cold-starts instead of entering standby
and restore fails silently. `pytest -m workload` won't catch it, since it never
touches a cluster. Setting both is safe either way, and it's exactly the
migration pattern we're asking integrations to follow.

**I'd keep the `Deprecated:` marker on `LegacySnapshotControlDirEnv`.** #131
fixed the validator in a90d7a3 so the legacy name is tolerated but no longer
required, which makes the marker accurate again.

caozhuozi:
@julienmancuso thanks for the context. I rebased and narrowed this PR down to the two non-overlapping pieces: the e2e restore fixtures and the placeholder image comment.

The fixtures now set both SNAPSHOT_RESTORE_STANDBY=1 and DYN_SNAPSHOT_RESTORE_STANDBY=1, so they don't assume which name the BASE_IMAGE entrypoint already honors. I also dropped the api/operator changes from this PR and left the LegacySnapshotControlDirEnv Deprecated marker untouched.

PR #104, opened 2026-08-24
✅ RUN-41928: Add standalone vLLM checkpoint and restore E2E

leeZardNav:
## Summary
- add a standalone vLLM checkpoint and restore workload without importing or deploying Dynamo
- exercise real generation before capture, then pause and sleep vLLM
- identify artifacts by `PodSnapshotContent` UID and restore through the `PodSnapshot` reference contract
- verify the `snapshot/Restored` condition, wake the engine, resume generation, and prove post-restore inference
- provide a restore-compatible test image and the required tun device and Hugging Face settings

## Validation
- latest HEAD `d50e1bc`: Python compilation and focused pytest collection
- `lee-dyn-4`: NVIDIA A10G, driver 580.82.07
- Snapshot `v0.0.0-gb52c6009` in `snapshot-system`; no Dynamo or Run:AI release
- focused PodSnapshot-reference E2E passed in 481.65s with pre-capture and post-restore inference
- build, check, test, DCO, and CodeRabbit checks pass

## Jira
https://runai.atlassian.net/browse/RUN-41928


e2e/tests/test_vllm_checkpoint.py
```diff
@@ -0,0 +1,211 @@
...
+def test_vllm_checkpoint_captures_quiesced_engine(
+    config: k8s.E2EConfig,
+) -> None:
+    run = VllmCheckpointRun.new()
+    try:
+        k8s.create_pod(vllm_source_pod(config, run))
+        source = snap.wait_for_pod_ready(config.namespace, run.source_pod, timeout=1200)
+        source_node = source.spec.node_name
+
+        snap.create_podsnapshot(
```
shayasoolin, line 56: the PR body says "pause generation, enter vLLM sleep mode". I suggest adding some model prompt requests (i.e. generation) here, just before creating the snapshot. both as a validation (on top of readiness) as well as to mess with the gpu-memory/kv-cache a bit. I would also expect to have such prompts post restore, when the test gets there too. there it will actually be even more important.
I can't say it's a must-have (as vllm init cycle also sends a simple request itself), but intuitively it feels right and provides some more confidence.

leeZardNav:
Closing this PR because we are continuing this work on Oz's `ozb/alpha-docs` branch in [PR #118](https://github.com/ai-dynamo/snapshot/pull/118). We will add any remaining changes needed from this PR to PR #118.

PR #106, opened 2026-08-24
refactor(operator): one signal owner per SnapshotJob condition

oleg-kushniriov:
### Summary

Aligns the SnapshotJob reconciler with the requested reconciliation model: each condition has exactly one signal owner, and once a PodSnapshot exists its status is the only capture signal SnapshotJob consumes.

- **PodSnapshot controls `Captured`.** `deriveCapturedStatus` is now the single writer of the `Captured` condition, fed exclusively by `PodSnapshot.status`. A Failed capture stamps `Captured` with the capture's own reason and message; the aggregate failure no longer rewrites it. When a capture failure coincides with a source Job deadline, `Failed` still reports `DeadlineExceeded` as the aggregate root cause, but `Captured` keeps the capture's result. An already-True `Captured` is never downgraded, so a recorded capture success survives losing its PodSnapshot.
- **Job controls the workload gate only.** `snapshotJobTerminalFailure` derives the aggregate failure alone and never feeds `Captured`. The defensive terminal arbitration after creating a PodSnapshot is removed — a capture created this pass is pending by definition.
- **Direct reads shrink to identity and confirmed-absence checks.** Helper-container verification is now cache-first: terminated container states are immutable, so staleness can only report a reversible `helpersStillRunning` requeue; the API read remains only to confirm a cached pod miss before the sticky unverifiable failure. Because a PodSnapshot terminal status is sticky, stale cached status can only delay progress at this layer — it cannot invent a competing terminal result.

### Validation

- Operator suite passes; new coverage: `Captured` keeps the capture's own reason under a Job deadline, helper verification tolerates an informer-lagged pod, and a True `Captured` is not downgraded by `PodSnapshotDeleted`.
- The remaining `NonCacheReadClient` uses are exactly the sanctioned set: recorded-child cache-miss confirmation (Job and PodSnapshot), capture-absence confirmation before `SourceCompletedWithoutCapture`, adoption identity, and helper-absence confirmation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SnapshotJob status reporting so capture-specific failures remain visible when the source job fails or reaches its deadline.
  * Preserved successful capture status after related resources are removed.
  * Improved verification reliability when resources are temporarily missing or stale in the local cache.
  * Ensured newly initiated captures remain pending until capture status is available.
  * Improved handling of helper-container restarts and authoritative resource verification.

### DIFF
--- a/operator/internal/controller/snapshotjob_reconciler.go
+++ b/operator/internal/controller/snapshotjob_reconciler.go
@@ -36,6 +36,13 @@ import (
 
 // SnapshotJobReconciler reconciles a SnapshotJob.
 //
+// Each condition has one owner: PodSnapshot.status is the only capture signal
+// (Captured), Job.status is the only workload completion/failure signal, and
+// SnapshotJob publishes the aggregate Completed/Failed result. The source Pod
+// is read solely for the identity needed to construct the PodSnapshot and to
+// verify helper containers; SnapshotJob never reinterprets PodSnapshotContent
+// or Pod state into a capture result of its own.
+//
 // Resource helpers create, find, and classify the Job, source Pod, and
 // PodSnapshot without mutating SnapshotJob status. Reconcile then derives the
 // complete status from those observations and persists it once. Completion is
@@ -151,14 +158,17 @@ func (r *SnapshotJobReconciler) reconcileResources(ctx context.Context, sj *snap
 	}
 }
 
-// reconcilePodSnapshotResources drives the capture and arbitrates it against
-// the source Job. Once a PodSnapshot exists, its terminal result decides the
-// SnapshotJob: Ready completes it (regardless of the Job's state), Failed fails
-// it with the capture's own reason, and a pending capture is waited on with a
-// bounded requeue even when the Job is already terminal — the checkpoint
+// reconcilePodSnapshotResources implements the capture-boundary flow: look up
+// the deterministic PodSnapshot; if absent, consult the Job/source Pod only to
+// create it or determine that creation is no longer possible; if present, the
+// capture signal comes exclusively from PodSnapshot.status, and Job.status
+// supplies only the independent workload completion/failure gate. A pending
+// capture is waited on even when the Job is already terminal — the checkpoint
 // terminates the source process, so a Job failure racing a committed artifact
-// is the expected success sequence, not an error. Only before a capture exists
-// does a terminal source Job fail the SnapshotJob directly.
+// is the expected success sequence, not an error. Because a PodSnapshot
+// terminal status is sticky, stale cached status can only delay progress here,
+// never invent a competing result; direct API reads are limited to identity
+// and confirmed-absence checks immediately before an irreversible decision.
 func (r *SnapshotJobReconciler) reconcilePodSnapshotResources(ctx context.Context, sj *snapshotv1alpha1.SnapshotJob, job *batchv1.Job) (snapshotJobObservation, ctrl.Result, error) {
 	observed := snapshotJobObservation{job: job}
 	snap, err := r.findOwnedPodSnapshot(ctx, sj)
@@ -186,14 +196,9 @@ func (r *SnapshotJobReconciler) reconcilePodSnapshotResources(ctx context.Contex
 			return observed, ctrl.Result{}, nil
 		}
 		if terminal.state != sourceJobComplete {
-			observed, result, err := r.createPodSnapshotForSourceJob(ctx, sj, job)
-			if err != nil {
-				return observed, result, err
-			}
-			if observed.failure == nil {
-				observed.failure = snapshotJobTerminalFailure(observed.job, observed.podSnapshot)
-			}
-			return observed, result, nil
+			// A capture created this pass is pending by definition; no terminal
+			// arbitration applies until its status reports a result.
+			return r.createPodSnapshotForSourceJob(ctx, sj, job)
 		}
 		// The source Job is complete, so a capture created now can never succeed.
 		// Confirm the capture's absence authoritatively before the sticky failure.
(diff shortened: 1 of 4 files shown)


hhzhang16 (approved):

operator/internal/controller/snapshotjob_reconciler.go
```diff
@@ -304,15 +309,24 @@ func (r *SnapshotJobReconciler) classifyHelperContainers(ctx context.Context, sj
 	if len(helpers) == 0 && len(sidecars) == 0 {
 		return nil, false, nil // the target is the only container
 	}
-	pod, found, err := findSourcePod(ctx, r.NonCacheReadClient, job)
+	// Cache-first: terminated container states are immutable, so staleness can
+	// only report helpersStillRunning — a reversible requeue. Only the sticky
+	// absence failure below needs the cached miss confirmed against the API.
+	pod, found, err := findSourcePod(ctx, r.Client, job)
```
julienmancuso: [P1] Recheck restartable sidecars before completing

The monotonicity assumption above is not true for native sidecars: an init container with `restartPolicy: Always` can move from `State.Terminated` back to `State.Running` when kubelet restarts it. Informer lag can therefore return an older terminated state while the authoritative Pod has a live sidecar; the code accepts that state and may delete the Job while the sidecar is still working. Since success is irreversible, please use `NonCacheReadClient` for this verification (or authoritatively recheck every success result before returning it). A cached `helpersStillRunning` result remains safe to requeue.

julienmancuso (changes requested):
Requesting changes because the cache-first completion check is not monotonic for restartable native sidecars. Details inline.

julienmancuso (approved):

operator/internal/controller/snapshotjob_reconciler.go
```diff
@@ -186,14 +196,9 @@ func (r *SnapshotJobReconciler) reconcilePodSnapshotResources(ctx context.Contex
...
-			observed, result, err := r.createPodSnapshotForSourceJob(ctx, sj, job)
-			if err != nil {
-				return observed, result, err
-			}
-			if observed.failure == nil {
-				observed.failure = snapshotJobTerminalFailure(observed.job, observed.podSnapshot)
-			}
-			return observed, result, nil
+			// A capture created this pass is pending by definition; no terminal
+			// arbitration applies until its status reports a result.
```
Ronkahn21, line 200: redundent

operator/internal/controller/snapshotjob_reconciler.go
```diff
@@ -304,15 +309,24 @@ func (r *SnapshotJobReconciler) classifyHelperContainers(ctx context.Context, sj
...
 		return nil, false, nil // the target is the only container
 	}
-	pod, found, err := findSourcePod(ctx, r.NonCacheReadClient, job)
+	// Cache-first: terminated container states are immutable, so staleness can
+	// only report helpersStillRunning — a reversible requeue. Only the sticky
+	// absence failure below needs the cached miss confirmed against the API.
+	pod, found, err := findSourcePod(ctx, r.Client, job)
 	if err != nil {
 		return nil, false, fmt.Errorf("find source pod to verify helper containers: %w", err)
 	}
```
Ronkahn21: simpligy it be check if ignoreNotFoundError , make the found bool retunr redundent

operator/internal/controller/snapshotjob_reconciler.go
```diff
@@ -304,15 +309,24 @@ func (r *SnapshotJobReconciler) classifyHelperContainers(ctx context.Context, sj
...
+		pod, found, err = findSourcePod(ctx, r.NonCacheReadClient, job)
+		if err != nil {
+			return nil, false, fmt.Errorf("confirm source pod absence for helper verification: %w", err)
+		}
+		if !found {
+			return &snapshotJobFailure{
+				reason: snapshotv1alpha1.ReasonJobFailed,
+				cause:  errors.New("source pod is gone before its helper containers could be verified"),
+			}, false, nil
+		}
```
Ronkahn21: can we export this to one function that manged the uncahced cahed verions

operator/internal/controller/snapshotjob_reconciler.go
```diff
@@ -456,26 +470,41 @@ func deriveRunningStatus(next *snapshotv1alpha1.SnapshotJob, observed snapshotJo
...
-				snapshotv1alpha1.ReasonCaptureInProgress, "waiting for the node agent to capture the checkpoint")
+				failure.reason, "capture did not complete: "+failure.cause.Error())
 		}
+		return failure
+	}
+
+	next.Status.PodSnapshotName = observed.podSnapshot.Name
+	if next.Status.PodSnapshotUID == "" {
+		next.Status.PodSnapshotUID = observed.podSnapshot.UID
+	}
```
Ronkahn21, line 506: I feel this overkill, as the intend is more importent than the same id 
I mean the intend is to caputre, it does not need to be care about the uid, even if eventually you recrate the podSnasphot twice

operator/internal/controller/snapshotjob_reconciler.go
oleg-kushniriov: findSourcePod is a List, not a Get; "not found" is err == nil, so there's no IsNotFound(err) to collapse onto - the bool is load-bearing.

operator/internal/controller/snapshotjob_reconciler.go
oleg-kushniriov: UID is what lets the reconciler distinguish the actually-captured object from a same-named recreation; dropping it breaks the sticky-failure/irreversible-success invariant.

operator/internal/controller/snapshotjob_reconciler.go
oleg-kushniriov: did you mean the comment text or the remove snapshotJobTerminalFailure call?

Ronkahn21: comment

operator/internal/controller/snapshotjob_reconciler.go
Ronkahn21: I understand but do we care ... 
if we introduce retries do we need UID check  ?

Ronkahn21 (approved):

PR #79, opened 2026-08-18
feat(agent): forward CUDA multicast APIs

galletas1712:
## Summary
- add transparent CUDA multicast wrappers in `multicast.c`
- register multicast entry points across direct binding, `dlsym`, `cuGetProcAddress`, and CUDA runtime resolvers
- select CUDA 13.1 device-explicit bind ABIs when the build headers and requested API version support them
- forward calls and results unchanged; no multicast topology state or replay yet

This is layer 5 of [stack #156](https://github.com/ai-dynamo/snapshot/stack/156).

## Validation
- CPU-only fake CUDA driver/runtime test covers direct and resolver forwarding
- pinned CUDA-devel image build with the GLIBC 2.34 ceiling
- `go test ./api/... ./operator/... ./agent/...`

PR #78, opened 2026-08-18
feat(agent): forward POSIX CUDA VMM APIs

galletas1712:
## Summary
- replace the delivery artifact with transparent POSIX CUDA VMM wrappers in `interpose.c`
- cover direct ELF binding, `dlsym`, `cuGetProcAddress`, and CUDA runtime resolver families
- bootstrap the real `dlsym` with one Makefile-pinned `dlvsym(..., \"GLIBC_2.34\")` call
- reject CUDA linkage and glibc requirements newer than 2.34 at build time
- forward every intercepted call and result unchanged; no checkpoint state or multicast surface yet

This is layer 2 of [stack #156](https://github.com/ai-dynamo/snapshot/stack/156).

## Validation
- CPU-only fake CUDA driver/runtime test covers direct and resolver forwarding
- pinned CUDA-devel image build with `-Wall -Wextra -Werror`
- `go test ./api/... ./operator/... ./agent/...`

PR #110, opened 2026-08-25
feat(agent): scaffold CUDA interposer delivery

galletas1712:
## Summary
- establish the final `cuinterpose` source layout and build targets, with thin inert skeletons in the final filenames
- build and package a valid no-op `libcuinterposer.so` and static no-op `cuinterposer-coordinator`
- add the `nvidia.com/cuda-interposer: enabled` SnapshotJob opt-in; no agent-image field is added to the API
- centralize capture shaping in `api/podcontract`: one `emptyDir` and one init container using the Helm-configured agent image
- copy both `libcuinterposer.so` and `cuda-checkpoint` from that image, mount them only on checkpoint targets, set `LD_PRELOAD`, and wrap the target with the injected `cuda-checkpoint --launch-job`
- record enablement on the generated `PodSnapshot`
- restore the same executable bundle path through the fixed `ns-bind-mount` helper, without a restore init container
- remove the obsolete placeholder image target

This is layer 1 of [stack #156](https://github.com/ai-dynamo/snapshot/stack/156). It intentionally provides final packaging and file seams without CUDA interception or coordinator protocol behavior.

## Validation
- `make check` on #110 and at stack top
- all Go module tests on every intermediate branch
- production CUDA-builder compile on implementation layers
- full top-of-stack agent image build
- POSIX forwarding test on #78
- multicast forwarding test on #79 and final behavior test on #133

## Summary by CodeRabbit

* **New Features**
  * Added optional CUDA interposer support for snapshot capture and restore.
  * Enabled opt-in configuration through pod annotations, with automatic library injection, preload setup, and checkpoint launch wrapping.
  * Added interposer mounting during restore with cleanup and error handling.
  * Added configurable agent image handling across Helm deployments and snapshot jobs.
* **Bug Fixes**
  * Added validation for conflicting volumes, mounts, containers, and environment settings.
  * Improved restore metadata propagation for interposer-enabled workloads.
* **Tests**
  * Added coverage for enablement, validation, capture shaping, mounting, and restore behavior.


agent/Dockerfile
```diff
@@ -309,53 +308,3 @@ RUN mkdir -p /checkpoints /var/run/snapshot
 USER root
 
 ENTRYPOINT ["/usr/local/bin/snapshot-agent"]
-
```
galletas1712, line 312: Do not merge - we need to wait for statically linked tar etc first before we can get rid of placeholder, unless all workloads we support are already on the latest glibc matching 24.04

galletas1712:
Superseded by #212 (packaging), #213 (delivery: podcontract, operator, chart), and #214 (agent orchestration). The placeholder image stage is left untouched in the new stack, per the thread on the Dockerfile; its removal stays with the separate static-tar effort. The re-cut stack (#212 → #220) supersedes this PR. Its body is quoted verbatim in the replacement's Origin section, and each review thread here has a row in the replacement's "Review threads carried" table with what was done about it.

PR #113, opened 2026-08-25
Add parallel multi-container restore

Ronkahn21:
## Summary

- add `nvidia.com/restore-container-map` for cloning one captured container into multiple restore destinations
- validate the complete map before Pod shaping and before agent workers start
- restore destinations concurrently while a single coordinator publishes the Pod-level aggregate result, including `RestorePartiallySucceeded`
- add a CPU-only one-to-two restore E2E scenario

## Validation

- `go test ./api/... ./operator/...`
- Linux container: `go test ./...` from `agent/`
- Linux container: `go test -race ./internal/controller` from `agent/`
- Python compile and E2E test collection (5 tests)

The cluster-backed E2E scenario was added but not executed locally.

## Summary by CodeRabbit

* **New Features**
  * Restore one captured container into multiple destination containers using an optional mapping annotation.
  * Destinations restore concurrently with success, partial-success, pending, and failure outcomes.
  * Added validation for container names, source matching, nonempty mappings, and duplicate destinations.
  * Added destination-specific restore tokens, artifacts, readiness checks, and status reporting.

* **Documentation**
  * Updated API and command-line documentation with multi-container restore configuration and outcome details.

* **Tests**
  * Added lifecycle, validation, retry, and failure-handling coverage for multi-container restores.

### DIFF
--- a/agent/internal/controller/controller.go
+++ b/agent/internal/controller/controller.go
@@ -85,16 +85,34 @@ type NodeController struct {
 }
 
 type restoreArtifact struct {
-	SnapshotName  string
-	ContentUID    string
-	ContainerName string
-	Path          string
+	SnapshotName        string
+	ContentUID          string
+	SourceContainerName string
+	Path                string
 }
 
 type restoreTarget struct {
-	SnapshotName  string
-	ContentUID    string
-	ContainerName string
+	SnapshotName        string
+	ContentUID          string
+	SourceContainerName string
+}
+
+type restorePlan struct {
+	artifact *restoreArtifact
+	mappings []snapshotv1alpha1.RestoreContainerMapping
+}
+
+type restoreResultState int
+
+const (
+	restoreResultPending restoreResultState = iota
+	restoreResultSucceeded
+	restoreResultFailed
+)
+
+type restoreResult struct {
+	destination string
+	state       restoreResultState
 }
 
 type restorePendingError struct {
@@ -110,6 +128,7 @@ type restoreOperation struct {
 	controller  *NodeController
 	pod         *corev1.Pod
 	artifact    *restoreArtifact
+	destination string
 	containerID string
 	startedAt   time.Time
 	log         logr.Logger
@@ -121,8 +140,10 @@ const (
 	restoreContainerResolveTimeout     = 30 * time.Second
 	restoreFailedReason                = "RestoreFailed"
 	restoreInProgressReason            = "RestoreInProgress"
+	restorePartiallySucceededReason    = "RestorePartiallySucceeded"
 	restoreSucceededReason             = "RestoreSucceeded"
 	restoreAlreadyCompletedReason      = "RestoreAlreadyCompleted"
+	restoreAlreadyPartialReason        = "RestoreAlreadyPartiallySucceeded"
 	restoreAlreadyFailedReason         = "RestoreAlreadyFailed"
 	restoreFinalizerUpdateFailedReason = "RestoreFinalizerUpdateFailed"
 	restoreStatusUpdateFailedReason    = "RestoreStatusUpdateFailed"
@@ -410,26 +431,11 @@ func (w *NodeController) processRestoreQueueItem(ctx context.Context, key client
 		return
 	}
 	if isRestoreTerminal(pod) {
-		if isRestoreSucceeded(pod) {
-			emitPodEvent(ctx, w.clientset, w.log, pod, snapshotEventComponent, corev1.EventTypeNormal, restoreAlreadyCompletedReason, "Pod restore already completed; no further action is required")
-		} else {
-			message := findRestoredCondition(pod).Message
-			if message == "" {
-				message = "Pod restore previously failed; create a new restore Pod to retry"
-			}
-			emitPodEvent(ctx, w.clientset, w.log, pod, snapshotEventComponent, corev1.EventTypeWarning, restoreAlreadyFailedReason, message)
-		}
-		requeue = w.removeRestoreFinalizerWithEvent(ctx, pod)
+		requeue = w.handleTerminalRestorePod(ctx, pod)
 		return
 	}
 	if pod.DeletionTimestamp != nil {
 		requeue = w.removeRestoreFinalizerWithEvent(ctx, pod)
-		return
-	}
-	if err := w.addRestoreFinalizer(ctx, pod); err != nil {
-		w.log.Error(err, "Failed to protect restore Pod", "pod", key.String())
-		emitPodEvent(ctx, w.clientset, w.log, pod, snapshotEventComponent, corev1.EventTypeWarning, restoreFinalizerUpdateFailedReason, err.Error())
-		requeue = true
 		return
 	}
 	if !isRestorePodActive(pod) {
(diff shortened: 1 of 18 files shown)


api/v1alpha1/protocol.go
```diff
@@ -10,8 +10,16 @@ import (
 	"k8s.io/apimachinery/pkg/util/validation"
 )
 
+// RestoreContainerMapping maps the one captured source container to a restore
+// destination in the target Pod.
+// +kubebuilder:object:generate=false
+type RestoreContainerMapping struct {
+	Source      string
+	Destination string
+}
```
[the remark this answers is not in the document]
Ronkahn21: this is not a official API

agent/internal/controller/controller.go
```diff
@@ -548,21 +589,41 @@ func validatePodSnapshotContentForRestore(content *snapshotv1alpha1.PodSnapshotC
...
+	if err := validateRestoreMappingsForPod(&pod.Spec, mappings, containerName); err != nil {
+		return nil, nil, err
 	}
-	return &restoreTarget{SnapshotName: snapshot.Name, ContentUID: string(content.UID), ContainerName: containerName}, nil
+	return &restoreTarget{SnapshotName: snapshot.Name, ContentUID: string(content.UID), SourceContainerName: containerName}, mappings, nil
+}
+
+// validateRestoreMappingsForPod validates the mapping contract and ensures
+// every destination exists in the restore Pod spec.
+func validateRestoreMappingsForPod(spec *corev1.PodSpec, mappings []snapshotv1alpha1.RestoreContainerMapping, capturedSource string) error {
```
oleg-kushniriov, line 611: validateRestoreMappingsForPod - validates only mapping exists
while ValidateRestorePodSpec validated also that mount exists on certain container.
Should we validate it too for validateRestoreTarget

api/v1alpha1/constants.go
```diff
@@ -22,11 +22,17 @@ const (
...
 
+	// RestoreContainerMapAnnotation optionally maps the single captured source
+	// container to one or more restore destinations. Its value is a comma-separated
+	// list of source=destination pairs. When absent, restore uses the captured
+	// container name as the destination.
+	RestoreContainerMapAnnotation = "nvidia.com/restore-container-map"
+
 	// RestoredCondition is the Pod status condition owned by the node agent.
-	RestoredCondition = "snapshot/Restored"
+	RestoredCondition = "nvidia.com/Restored"
```
oleg-kushniriov, line 35: Breaking change: 
Consider doing it in a separate PR.
Also do we need to add handling for legacy Condition for another version?

e2e/snapshot_e2e/lifecycle.py
```diff
@@ -289,7 +290,7 @@ def wait_for_restored_condition(
 ) -> client.V1Pod:
     def check() -> client.V1Pod | None:
         pod = k8s.read_pod(namespace, pod_name)
-        restored = pod_condition(pod, "snapshot/Restored")
+        restored = pod_condition(pod, "nvidia.com/Restored")
         if restored and restored.status == status and restored.reason == reason:
             return pod
         if reason != "RestoreFailed" and restored and restored.reason == "RestoreFailed":
```
oleg-kushniriov: `wait_for_restored_condition` fail-fasts only on RestoreFailed, so a `RestorePartiallySucceeded` terminal outcome makes tests (and any consumer with the same switch) wait out the full timeout; the new reason and its create-a-new-Pod retry semantics are also undocumented.
here `isRestoreTerminal()` - `RestorePartiallySucceeded` also defined as terminal

api/v1alpha1/constants.go
Ronkahn21: its was never publsijed

agent/internal/controller/controller.go
```diff
@@ -579,54 +640,118 @@ func (w *NodeController) resolveRestoreArtifact(podKey string, target *restoreTa
...
+	pod *corev1.Pod,
+	artifact *restoreArtifact,
+	destination, podKey string,
+	recovering bool,
+) restoreResult {
+	result := restoreResult{destination: destination, state: restoreResultPending}
+	containerID, _ := w.resolveRestoreContainerID(ctx, pod, destination, podKey)
+	if containerID == "" {
+		return result
 	}
```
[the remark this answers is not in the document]
Ronkahn21: Keeping this pending/requeue behavior intentionally. The same 30-second resolution timeout on main returns requeue via ContainerPending; it is a polling window, not a restore lifecycle deadline. Treating it as terminal would change existing single-container semantics and can incorrectly fail a legitimate slow image pull or runtime startup. A terminal startup policy should be introduced explicitly rather than inferred from this internal polling interval.

agent/internal/controller/controller.go
Ronkahn21: Good question. I am keeping the agent preflight scoped to the mapping contract and destination existence. The supported creation path calls PrepareRestorePodSpec before Pod creation: it resolves every destination before mutating the spec, then EnsureControlVolume shapes each one. The agent does not currently duplicate the operator's restore-shaping validation for the volume, mount, env, probe, or seccomp profile. Copying only the mount check here would create a partial second validator; if we want hand-authored annotated Pods independently validated, the complete validator should first move to a shared package and be used by both paths.

e2e/snapshot_e2e/lifecycle.py
Ronkahn21: Fixed in ffc82e8. The E2E waiter now recognizes all three terminal reasons and fails immediately when the observed terminal reason differs from the expected one. The snapshotctl documentation now defines all-success, partial-success, and all-failed outcomes and states that partial/failed restores are terminal and require a new restore Pod.

api/v1alpha1/constants.go
Ronkahn21: Agreed that this changes the wire value. We are keeping the rename intentionally because the condition has not been released yet, so there are no supported legacy restore Pods to migrate. Supporting both names before the first release would add compatibility code without a compatibility requirement.

agent/internal/controller/controller.go
oleg-kushniriov: If multi-destination restore is deliberately scoped to non-networked/CPU-only workloads for now (matching the PR's CPU-only E2E scenario), should validateRestoreMappingsForPod or preflightRestore reject len(mappings) > 1 when w.config.CRIU.TcpEstablished is true, so the failure is an explicit validation error instead of an undefined runtime conflict?

Suggeston: If the combination is meant to be supported later, gate it behind an explicit check now (reject at preflight with a clear message) until INET remap is made destination-aware; if it's never meant to be supported, add the same guard permanently.

agent/internal/controller/controller.go
```diff
@@ -469,11 +503,18 @@ func (w *NodeController) preflightRestore(ctx context.Context, pod *corev1.Pod)
...
+	target, mappings, err := validateRestoreTarget(pod, snapshot, content)
 	if err != nil {
 		return nil, err
 	}
-	return w.resolveRestoreArtifact(fmt.Sprintf("%s/%s", pod.Namespace, pod.Name), target)
+	artifact, err := w.resolveRestoreArtifact(fmt.Sprintf("%s/%s", pod.Namespace, pod.Name), target)
+	if err != nil {
+		return nil, err
+	}
+	if w.config.CRIU.TcpEstablished && pod.Status.PodIP == "" {
```
oleg-kushniriov, line 514: If multi-destination restore is deliberately scoped to non-networked/CPU-only workloads for now (matching the PR's CPU-only E2E scenario), should preflightRestore reject len(mappings) > 1 when w.config.CRIU.TcpEstablished is true, so the failure is an explicit validation error instead of an undefined runtime conflict?

Suggestion: If the combination is meant to be supported later, gate it behind an explicit check now (reject at preflight with a clear message) until INET remap is made destination-aware; if it's never meant to be supported, add the same guard permanently.

agent/internal/controller/controller.go
oleg-kushniriov, line 611: For annotation-driven restore Pods (not shaped by snapshotctl), nothing validates that each destination mounts snapshot-control with its own subPath - but the sentinel protocol assumes that isolation. With a shared control directory, one destination can clear a sibling's restore-complete sentinel (restored process stuck in quiescence despite RestoreSucceeded) or be falsely marked restored off a sibling's sentinel.

Suggestions: Validate the control-mount shape in agent preflight when `len(mappings) > 1 `- e.g. require each destination's snapshot-control mount to carry a distinct subPath (or reuse the checks from `ValidateRestorePodSpec`, which is currently dead code in production, wiring it into `validateRestoreMappingsForPod` or `preflightRestore`).

api/v1alpha1/protocol.go
```diff
@@ -22,3 +30,57 @@ func GetRestoreFromSnapshotName(annotations map[string]string) (string, error) {
 	}
 	return snapshotName, nil
 }
+
```
oleg-kushniriov, line 33: A capture pod template with a stray nvidia.com/restore-container-map annotation (e.g. copy-pasted from a restore manifest) is currently accepted silently instead of being caught at capture time the way nvidia.com/restore-from is, breaking the "capture pods must have no snapshot-owned annotations" invariant the README documents for the sibling annotation.

Suggestions: Add the same rejection for snapshotv1alpha1.RestoreContainerMapAnnotation in `checkpoint.go` next to the existing RestoreFromAnnotation check.

agent/internal/controller/controller.go
Ronkahn21: its not scoped to non-networking, also GPU is something that required applcaition support so for now we have only CPU test for multi container

agent/internal/controller/controller.go
Ronkahn21: accepted

Ronkahn21: Fixed in 1ceab5e. Agent preflight now requires a snapshot-control emptyDir and, for every multi-restore destination, a mount at /snapshot-control with subPath equal to that destination name. Validation runs before the restore finalizer and workers. A focused test rejects a sibling-shared subPath.

agent/internal/controller/controller.go
Ronkahn21: No infrastructure guard should be added. Multi-destination restore is not scoped to CPU-only or non-networked workloads; the CPU-only E2E reflects the test workload currently available, not the product contract. The criu.tcpEstablished option is enabled by default and indicates CRIU capability, not an actual socket conflict. Likewise, GPU restore support is provided per destination by the infrastructure, while the application and its resource layout are responsible for ensuring duplicated network and GPU state can coexist. Rejecting this configuration in preflight would incorrectly turn an application-compatibility concern into an infrastructure limitation.

api/v1alpha1/protocol.go
Ronkahn21: Fixed in 1ceab5e. NewCheckpointJob now rejects both restore-from and restore-container-map when either annotation is present on the capture template, including an empty value. The checkpoint protocol test covers both keys.

oleg-kushniriov (approved):

PR #117, opened 2026-08-26
fix(operator): wait for CRD establishment in crd-installer

oleg-kushniriov:
### Problem

Installing the operator sometimes left a CRD (observed with `PodSnapshot`) unusable until the operator pod restarted. The `crd-installer` init container exits as soon as the API server accepts the apply, but CRD registration is asynchronous — the manager container could start before the CRD was Established, fail to build its informer, and crash. It only recovered because the container restart happened after registration finished.

### Fix

After applying each CRD, `crdinstaller` now polls `status.conditions[type=Established]` until it is `True` before reporting success (200ms interval, 30s timeout):

- Transient `Get` errors during the wait do not abort it — the timeout is the only terminal condition, and the last error is attached to the timeout message.
- On timeout, the last-observed CRD conditions are included in the error, so a terminal cause (e.g. a `NamesAccepted` conflict) is distinguishable from a slow API server.
- A log line is emitted once when a CRD is found not yet established, so the init container is not silent while waiting.

### Testing

- New unit tests cover the wait, the timeout path, transient-error tolerance, and error-message contents.
- Full `crdinstaller` package suite passes (`go test ./internal/crdinstaller/...`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * CRD installation now waits until each custom resource definition is established before reporting success.
  * Temporary API server errors and interrupted watches are retried.
  * Installation detects CRD deletion and authorization failures.
  * Timeout errors include the latest observed status or error details.
  * Updated permissions support monitoring CRD establishment.
  * Added logging when establishment waiting begins.
  * Corrected GPU restore test expectations for failed restores.

* **Tests**
  * Added coverage for delayed establishment, transient failures, authorization errors, timeouts, deletion, and diagnostic reporting.

### DIFF
--- a/charts/snapshot/templates/operator-rbac.yaml
+++ b/charts/snapshot/templates/operator-rbac.yaml
@@ -51,10 +51,6 @@ rules:
     resources: ["events"]
     verbs: ["create", "patch"]
   {{- if .Values.crdUpgrade.enabled }}
-  # RBAC cannot match resourceNames on create, and list/watch requests are
-  # only authorized against resourceNames on servers with the
-  # AuthorizeWithSelectors feature, so these three are granted unscoped.
-  # The installer watches each applied CRD until it is Established.
   - apiGroups: ["apiextensions.k8s.io"]
     resources: ["customresourcedefinitions"]
     verbs: ["create", "list", "watch"]
--- a/charts/snapshot/templates/operator-rbac.yaml
+++ b/charts/snapshot/templates/operator-rbac.yaml
@@ -51,10 +51,13 @@ rules:
     resources: ["events"]
     verbs: ["create", "patch"]
   {{- if .Values.crdUpgrade.enabled }}
-  # RBAC cannot match resourceNames on create, so only get/patch are scoped.
+  # RBAC cannot match resourceNames on create, and list/watch requests are
+  # only authorized against resourceNames on servers with the
+  # AuthorizeWithSelectors feature, so these three are granted unscoped.
+  # The installer watches each applied CRD until it is Established.
   - apiGroups: ["apiextensions.k8s.io"]
     resources: ["customresourcedefinitions"]
-    verbs: ["create"]
+    verbs: ["create", "list", "watch"]
   - apiGroups: ["apiextensions.k8s.io"]
     resources: ["customresourcedefinitions"]
     resourceNames:
--- a/operator/internal/crdinstaller/installer.go
+++ b/operator/internal/crdinstaller/installer.go
@@ -155,6 +155,9 @@ func waitForEstablished(ctx context.Context, cl Client, log logr.Logger, gvk sch
 	switch {
 	case err == nil:
 		return nil
+	case lastErr != nil && len(lastConditions) > 0:
+		return fmt.Errorf("%w (last error: %w; last observed conditions: %s)",
+			err, lastErr, formatConditions(lastConditions))
 	case lastErr != nil:
 		return fmt.Errorf("%w (last error: %w)", err, lastErr)
 	case len(lastConditions) > 0:
--- a/operator/internal/crdinstaller/installer.go
+++ b/operator/internal/crdinstaller/installer.go
@@ -7,6 +7,7 @@ package crdinstaller
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"time"
@@ -64,7 +65,7 @@ func (r Results) Changed() bool {
 func InstallCRDs(ctx context.Context, cl Client, log logr.Logger, manifests []string) (Results, error) {
 	results := make(Results, 0, len(manifests))
 	for _, manifest := range manifests {
-		res, err := applyCRD(ctx, cl, manifest)
+		res, err := applyCRD(ctx, cl, log, manifest)
 		if err != nil {
 			return results, err
 		}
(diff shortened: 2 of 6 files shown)


Ronkahn21 (commented):
Could we use client-go’s watchtools.UntilWithSync instead of polling every 200 ms? The current loop generates five GET requests per second per CRD and treats every GET error—including terminal errors—as transient until timeout. UntilWithSync checks the initial state, watches subsequent updates, and transparently relists after disconnections or expired resource versions, while preserving the timeout and last-condition diagnostics.

operator/internal/crdinstaller/installer.go
```diff
@@ -105,3 +122,74 @@ func applyCRD(ctx context.Context, cl Client, manifest string) (Result, error) {
...
+		func(event watch.Event) (bool, error) {
+			if event.Type == watch.Deleted {
+				return false, fmt.Errorf("CRD %q was deleted while waiting", name)
+			}
+			crd, ok := event.Object.(*apiextensionsv1.CustomResourceDefinition)
+			if !ok || crd.Name != name {
+				return false, nil
+			}
+			lastConditions = crd.Status.Conditions
+			if crdEstablished(crd.Status.Conditions) {
```
julienmancuso, line 174: Established=True is sufficient for the current upgrade model because the GVR remains unchanged (nvidia.com/v1alpha1/...), the existing endpoint is already registered, while newly created CRDs start unestablished and are correctly handled here.
One limitation worth noting: if a future release introduces a new served version, the existing Established=True condition may be carried forward before that new version is discoverable. Such an upgrade should additionally wait for the desired GVR in API discovery

operator/internal/crdinstaller/installer.go
```diff
@@ -105,3 +122,74 @@ func applyCRD(ctx context.Context, cl Client, manifest string) (Result, error) {
...
+			return crdClient.Watch(ctx, opts)
+		},
+	}
+
+	// On timeout the bare error says nothing about the cause; the last
+	// observation distinguishes a slow apiserver from e.g. a NamesAccepted
+	// conflict that will never resolve.
+	var lastConditions []apiextensionsv1.CustomResourceDefinitionCondition
+	waiting := false
+	_, err := watchtools.UntilWithSync(ctx, lw, &apiextensionsv1.CustomResourceDefinition{}, nil,
```
julienmancuso, line 164: UntilWithSync retries reflector failures internally, but the code only retains observed conditions. Missing list/watch RBAC therefore waits 30 seconds and returns a generic timeout instead of the underlying Forbidden; transport failures behave similarly. This contradicts the PR’s promised “last error” diagnostics. Capture the last reflector error, fail fast on permanent authorization errors, and test the no-event failure path.

charts/snapshot/templates/operator-rbac.yaml
```diff
@@ -51,10 +51,13 @@ rules:
     resources: ["events"]
     verbs: ["create", "patch"]
   {{- if .Values.crdUpgrade.enabled }}
-  # RBAC cannot match resourceNames on create, so only get/patch are scoped.
+  # RBAC cannot match resourceNames on create, and list/watch requests are
+  # only authorized against resourceNames on servers with the
+  # AuthorizeWithSelectors feature, so these three are granted unscoped.
+  # The installer watches each applied CRD until it is Established.
```
Ronkahn21: delete the commets

operator/internal/crdinstaller/installer.go
```diff
@@ -105,3 +123,111 @@ func applyCRD(ctx context.Context, cl Client, manifest string) (Result, error) {
...
+// current state, updates arrive as watch events, and the underlying informer
+// relists after disconnections or expired resource versions.
+func waitForEstablished(ctx context.Context, crdClient apiextensionsv1client.CustomResourceDefinitionInterface, log logr.Logger, name string) error {
+	ctx, cancel := context.WithTimeout(ctx, establishedTimeout)
+	defer cancel()
+
+	// UntilWithSync retries list/watch failures internally and would surface
+	// them only as a bare timeout. Record the most recent failure for the
+	// timeout diagnostics, and abort the wait outright on an authorization
+	// error, which retrying cannot resolve.
```
Ronkahn21: remove

operator/internal/crdinstaller/installer.go
```diff
@@ -105,3 +123,111 @@ func applyCRD(ctx context.Context, cl Client, manifest string) (Result, error) {
...
+	var (
+		reflectorMu  sync.Mutex
```
Ronkahn21, line 137: why do you need mutex ?

operator/internal/crdinstaller/installer.go
```diff
@@ -105,3 +123,111 @@ func applyCRD(ctx context.Context, cl Client, manifest string) (Result, error) {
...
+			observe(err)
+			return list, err
+		},
+		WatchFuncWithContext: func(ctx context.Context, opts metav1.ListOptions) (watch.Interface, error) {
+			opts.FieldSelector = selector
+			w, err := crdClient.Watch(ctx, opts)
+			observe(err)
+			return w, err
+		},
+	}
```
Ronkahn21: this should be outside of this function

operator/internal/crdinstaller/installer.go
oleg-kushniriov: Agreed - `Established=True `is sufficient while v1alpha1 is the only served version. When a future release adds a new served version, this wait will also need to check API discovery for the new GVR. Tracked in DEP-1216.

operator/internal/crdinstaller/installer.go
oleg-kushniriov: The mutex is needed because observe writes reflectorErr/failFast on the reflector goroutine that UntilWithSync spawns internally, while waitForEstablished reads them on its own goroutine after UntilWithSync returns - a cross-goroutine hand-off (plus the check-then-set on failFast when concurrent list/watch callbacks race) that would be a data race without synchronization.

Ronkahn21 (approved):

PR #118, opened 2026-08-26
📝 RUN-41928: Complete vLLM snapshot-ready image guide

leeZardNav:
## Summary

- add downloadable `app.py` and `Dockerfile.vllm` companion files for the vLLM image guide
- provide `curl` commands so users can save both files locally instead of copying large inline examples
- document model selection, the single-image build, and the checkpoint and restore lifecycle

## Validation

- `npx --yes markdownlint-cli2 docs/guides/vllm.md`
- parsed `docs/guides/vllm/app.py` with `ast.parse`
- `git diff --check`


docs/guides/vllm/app.py
```diff
@@ -0,0 +1,84 @@
...
+import asyncio
+import os
+from pathlib import Path
+
+from vllm import SamplingParams
+from vllm.engine.arg_utils import AsyncEngineArgs
+from vllm.usage.usage_lib import UsageContext
+from vllm.v1.engine.async_llm import AsyncLLM
+
+MODEL = "Qwen/Qwen3-0.6B"
```
ozbarshalom: Why is it hard coded? I believe in vLLM it's provided as a variable we don't need to specify it again

docs/guides/vllm/Dockerfile.vllm
```diff
@@ -0,0 +1,29 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+ARG VLLM_RUNTIME_IMAGE=vllm/vllm-openai:v0.27.1
```
ozbarshalom, line 4: 0.28 is available no ?

ozbarshalom (changes requested):

docs/guides/vllm/Dockerfile.vllm
leeZardNav: done in 2c6c303

docs/guides/vllm/app.py
leeZardNav: I kept the model in `app.py` because this example runs a Python application instead of `vllm serve`, so it passes the model directly to `AsyncEngineArgs`. The guide asks users to set that value before building.

ozbarshalom: We should not have a custom image per model, it should be generic

ozbarshalom (changes requested):

docs/guides/vllm/app.py
leeZardNav: what do you mean? this is an example for the users, and they can modify it. what do you mean by generic?

ozbarshalom (approved):

PR #109, opened 2026-08-24
feat(cuda): add a transfer-neutral CustomStorage helper core

hhzhang16:
## Summary

Add the transfer-neutral foundation used by the CUDA checkpoint operation code.

## Details

- bounded transfer configuration and chunk layout validation
- incremental SHA-256 content digests
- shared cancellation for sibling transfer work
- a backend interface with an explicit unavailable implementation
- a CUDA 13.4 header-only build boundary on the existing CUDA 13.0 agent base

It does not add the daemon, storage manifest, production transfer adapter, PageBroker protocol, or Snapshot orchestration.

## Architecture

PageBroker is the canonical owner of node-local CUDA execution and artifact staging. Snapshot owns workload orchestration. Later MRs consume these contracts behind PageBroker's protobuf transaction boundary.

## Validation

- Linux/amd64 `cuda-transfer-contracts-builder`: passed.
- CUDA 13.4 header compile for the unavailable backend: passed.
- Content-digest tests: passed.
- Transfer-configuration tests: passed.
- Cancellation tests: passed.
- `git diff --check`, including new files: passed.

## Summary by CodeRabbit

- **New Features**
  - Added foundational CUDA checkpoint transfer support, including configurable storage layouts, chunked transfers, SHA-256 content verification, transfer metrics, and cancellation controls.
  - Added validation for transfer settings, storage paths, alignment, capacity, and layout consistency.
  - Added graceful handling when a CUDA transfer backend is unavailable.

- **Tests**
  - Added automated coverage for digest generation, transfer configuration, cancellation behavior, and unavailable-backend handling.
  - CI now runs the helper tests when supported and reports missing required test dependencies.

- **Documentation**
  - Added guidance for CUDA checkpoint transfer behavior, configuration, and testing requirements.

PR #120, opened 2026-08-26
docs: document CUDA CustomStorage configuration

hhzhang16:
## Summary

Adds documentation for configuring and validating Snapshot's opt-in CUDA CustomStorage path (the behavior implemented in #109 + #96). The guide is linked from the main README, chart documentation, storage documentation, troubleshooting, and the guide index.

## Review focus

Whether the CUDA capability requirements, Helm values, artifact compatibility, and current limitations are clear and accurate for someone enabling this for the first time.

## Validation

- Cross-checked the values, defaults, capability fields, and failure behavior against #96.
- Helm lint and render passed with the documented `posix` configuration.
- All newly added links and anchors resolve.
- `git diff --check` passes.


README.md
```diff
@@ -71,6 +71,8 @@ Before installing Snapshot, make sure the following are in place:
 - The [Helm](https://helm.sh/docs/intro/install) CLI
 - A cluster that permits privileged pods for the node agent — see [Security](docs/operations/security.md)
 
+The default CUDA checkpoint path works with the general driver prerequisite above. Enabling [CUDA CustomStorage](docs/guides/custom-storage.md) additionally requires a driver that exports the CUDA 13.4 CustomStorage API; CUDA 13.x compatibility alone is not a capability check.
```
ozbarshalom: CUDA 13.4 is not GA

ozbarshalom: https://docs.nvidia.com/cuda/developer-preview/13.4/cuda-toolkit-release-notes/index.html#driver-compatibility which CUDA drivers supports CUDA 13.4?

charts/snapshot/README.md
```diff
@@ -137,6 +150,11 @@ kubectl get pods -n ${NAMESPACE} -l app.kubernetes.io/name=snapshot -o wide
 | `seccomp.deploy` | Deploy the CRIU seccomp profile ConfigMap and init container. Use this field name; `seccomp.enabled` is not a chart value | `true` |
 | `runtime.type` | CRI backend: `containerd` or `crio` | `containerd` |
 | `runtime.socketPath` | CRI socket (empty = default for `runtime.type`) | `""` |
+| `config.cudaCheckpoint.storageMode` | Storage policy for newly created CUDA checkpoints: `legacy` or `posix` | `legacy` |
```
ozbarshalom: move config.cudaCheckpoint -> storage.gpu.engine

storageMode -> backend [criu  | nixl ]  later we will also have <objectStorage>

charts/snapshot/README.md
```diff
@@ -137,6 +150,11 @@ kubectl get pods -n ${NAMESPACE} -l app.kubernetes.io/name=snapshot -o wide
...
+| `config.cudaCheckpoint.transferBufferCount` | Pinned transfer slots per active CUDA device | `4` |
+| `config.cudaCheckpoint.transferChunkBytes` | Bytes per pinned transfer slot | `67108864` |
+| `config.cudaCheckpoint.daemon.maxOperationSeconds` | Cooperative watchdog for one CUDA helper operation | `3600` |
```
ozbarshalom: @dfeigin-nv

docs/guides/custom-storage.md
```diff
@@ -0,0 +1,156 @@
...
+The Snapshot agent image may be built against CUDA 13.0 headers because the helper dynamically resolves the public CUDA 13.4 entry point. The node driver, not the image's toolkit version, determines whether CustomStorage is available at runtime.
+
+## Enable CustomStorage
+
+Set `config.cudaCheckpoint.storageMode` to `posix` in the Helm values used for the Snapshot release:
+
+```yaml
+config:
+  cudaCheckpoint:
+    storageMode: posix
```
ozbarshalom, line 38: I'm not sure the full example is required here only to enable NIXL

docs/guides/custom-storage.md
```diff
@@ -0,0 +1,156 @@
...
+## Configuration values
+
+| Value | Meaning | Default | Valid values |
+|---|---|---:|---|
+| `config.cudaCheckpoint.storageMode` | Storage policy for newly created CUDA checkpoints | `legacy` | `legacy`, `posix` |
+| `config.cudaCheckpoint.transferBufferCount` | Pinned transfer slots per active CUDA device; this controls pipeline depth, not the number of worker threads | `4` | `1`-`8` |
+| `config.cudaCheckpoint.transferChunkBytes` | Bytes in each pinned transfer slot | `67108864` (64 MiB) | 1-256 MiB, 4096-byte aligned |
+| `config.cudaCheckpoint.daemon.maxOperationSeconds` | Cooperative transfer watchdog for one helper operation | `3600` | `1`-`3600` |
+| `config.cudaCheckpoint.daemon.resources` | CPU and memory requests and limits for the helper container | See the example above | Kubernetes resource values |
+| `config.restore.restoreTimeoutSeconds` | Overall restore budget in the Snapshot agent | `8100` | Positive integer |
```
ozbarshalom: why is it in the custom storage guide?

charts/snapshot/README.md
dfeigin-nv: I think "Cooperative watchdog for one CUDA helper operation" is unclear from a user perspective and also I don't get it.

docs/guides/custom-storage.md
```diff
@@ -0,0 +1,156 @@
+# Use CUDA CustomStorage
```
ozbarshalom: custom storage engine

docs/guides/custom-storage.md
```diff
@@ -0,0 +1,156 @@
...
+- A Snapshot agent image built with the NIXL POSIX transfer adapter. A transfer-neutral helper build does not support `posix` mode.
+- A shared RWX checkpoint PVC with enough capacity for the CRIU, root filesystem, and CUDA extent artifacts. Plan for approximately the checkpointed CUDA allocation bytes in addition to the existing process artifacts and temporary staging space.
+- A one-GPU workload. The initial `posix` implementation supports one or more CUDA-owning processes on one GPU. Snapshot rejects multi-GPU CustomStorage checkpoint creation before locking a target.
+
+The Snapshot agent image may be built against CUDA 13.0 headers because the helper dynamically resolves the public CUDA 13.4 entry point. The node driver, not the image's toolkit version, determines whether CustomStorage is available at runtime.
+
+## Enable CustomStorage
```
ozbarshalom: Customers don't "enable custom storage" they can modify the IO engine to use NIXL instead of CRIU.

ozbarshalom (changes requested):

charts/snapshot/README.md
hhzhang16: They are not interchangeable backend values imo -- CRIU is the process checkpoint engine, while NIXL POSIX is the CUDA artifact transfer backend.

README.md
hhzhang16: The CustomStorage path is an explicit opt-in using the CUDA 13.4 Developer Preview API, while the default remains the existing driver-managed path. The guide calls this out

docs/guides/custom-storage.md
hhzhang16: I included the transfer and resource values to show the operational defaults but I can reduce this to the minimal setting so it's not as scary/complicated-looking

docs/guides/custom-storage.md
hhzhang16: I think this is the same layer distinction as one of the above comments. NIXL does not replace CRIU here — CRIU still captures and restores the process state. Selecting `posix` changes the CUDA path from driver-managed checkpoint storage to the CUDA 13.4 CustomStorage API, with NIXL POSIX moving the resulting extents.

README.md
hhzhang16: The helper checks `cuCheckpointOperationComplete` at runtime and rejects `posix` before mutating the workload when it isn’t available

docs/guides/custom-storage.md
hhzhang16: Updated so that the enablement example now contains only the required `storageMode: posix` setting. The tuning and resource options remain in the reference table.

charts/snapshot/README.md
hhzhang16: I’m open to reorganizing the config hierarchy, but the accurate values would be closer to `driver-managed | nixl-posix`, not `criu | nixl`. Changing it would also require updating #96 rather than only renaming it in the docs

charts/snapshot/README.md
hhzhang16: Reworded

hhzhang16: It's the maximum time allowed for one CUDA checkpoint or restore helper request

charts/snapshot/README.md
ozbarshalom: I do want to change the hierarchy, of course not only in the docs but that's my reference as PM for these interface changes.
regarding values: @dfeigin-nv WDYT?

docs/guides/custom-storage.md
ozbarshalom: From a user perspective, customStorage is another storage rather than the existing supported storage which is local PVC.  This is not the case. I do understand that for the CUDA C/R it's a custom storage backend / engine but we should slightly abstract it from Snapshot users. (naming wise) 

@dfeigin-nv  WDYT?

dfeigin-nv: I think the naming should be something clearer than custom storage. Custom Storage is more of an implementation detail for us, not something the user should actively choose

README.md
ozbarshalom: Which CUDA driver supports CUDA 13.4? I believe there is a support matrix of GPU operator + CUDA driver + CUDA version

The missing piece for me is which CUDA driver does customers needs to install

docs/guides/custom-storage.md
dfeigin-nv: only after criu finishes the IO does the driver-managed checkpoint storage start, so legacy | criu are both correct. What is most clear to the user should be used.

hhzhang16:
Closing for now; will update with new docs after all MRs are merged in

PR #121, opened 2026-08-26
perf(snapshot): remove duplicate CustomStorage restore prefetch

hhzhang16:
## Summary

Removes Snapshot's separate full-artifact CustomStorage prefetch so each CUDA extent is read once during restore by the authoritative CUDA helper.

## Why

The existing POSIX restore path reads every CUDA extent twice:

1. the Snapshot agent reads the complete artifact into the page cache while CRIU runs;
2. the CUDA helper reads the complete artifact again into registered pinned buffers and copies it to the GPU.

The first read is best-effort only. It is not authoritative and does not replace the helper's extent validation or transfer. For large checkpoints, the extra read dominates restore wall time even when it overlaps CRIU.

## Main changes

- Remove the agent-side full-artifact prefetch goroutine and cancellation plumbing.
- Remove the now-unused prefetch implementation and unit tests.
- Keep the helper's existing pinned-buffer NIXL/POSIX read as the single authoritative data path.

## TRT-LLM TP4/EP4 + GMS result

The controlled diagnostic used DSV4 Flash NVFP4, TP4/EP4, 1M context, 64 sequences, GMS V1, and Snapshot POSIX CustomStorage on one B200 node. The artifact contained 328,162,344,960 engine CustomStorage bytes, 195,723,001,856 GMS CustomStorage bytes, and about 19.7 GB of non-CUDA/CRIU data.

| Metric | Prefetch enabled | Single authoritative read | Delta |
| --- | ---: | ---: | ---: |
| GMS restore wall | 258.391 s | 147.233 s | -111.158 s (-43.0%) |
| GMS CUDA phase | 142.061 s | 145.136 s | +3.074 s (+2.2%) |
| Engine restore wall | 436.973 s | 252.308 s | -184.665 s (-42.3%) |
| Engine CUDA phase | 243.682 s | 247.008 s | +3.326 s (+1.4%) |
| Pod create-to-Ready | 723.772 s | 422.000 s | -301.772 s (-41.7%) |

The CUDA phases remained essentially flat while the separate 115.5-second GMS and 192.1-second engine prefetch reads disappeared. That isolates the gain to eliminating duplicate storage I/O rather than changing CUDA transfer semantics.

The successful restore also passed restored-owner GMS validation, recreated the four-rank NCCL communicator, produced an exact all-reduce result of 10.0 on every rank, and completed post-restore inference.


hhzhang16:

PR #126, opened 2026-08-27
fix(agent): monitor restored process lifecycle

### LINKED ISSUE #12: restore: restored process is never PID 1 — container lifecycle decoupled from workload health
### Background

When a restore pod starts, the workload container runs in standby mode — its entrypoint becomes PID 1. The node agent enters the container's namespaces via `nsenter` and uses CRIU to restore the checkpoint process tree, but since PID 1 is already occupied, the restored workload lands at PID 2 or higher. PID 1 (the standby process) remains alive for the lifetime of the container.

This creates two distinct failure modes:

---

### Problem A — workload exit is invisible to Kubernetes

Because the restored workload is never PID 1, its exit does not terminate the container:

- If restore fails mid-flight and the agent's SIGKILL of the placeholder ([`controller.go:601`](https://github.com/ai-dynamo/snapshot/blob/main/agent/internal/controller/controller.go#L601)) races or is skipped (e.g., agent OOM-killed), PID 1 keeps running and the container hangs.
- After a successful restore, if the workload process exits for any reason, PID 1 (standby) is still alive — the container remains `Running` with nothing inside it.

**Observed symptom:** restored workload dead, pod `Running`, no restart triggered.

---

### Problem B — PID 1 (standby) death causes spurious container failure

After a successful restore, PID 1 is the dormant standby process — not the restored workload. If the standby crashes for any reason unrelated to the workload (OOM, memory corruption, etc.), the container exits and Kubernetes records the pod as `Failed` even though the workload was healthy.

**Observed symptom:** pod `Failed`, restored workload was running fine at the time.

caozhuozi:
## Summary

Partially fixes #12.

This fixes Problem A from the issue: after restore, the restored workload process
can exit while the placeholder PID 1 keeps the container Running.

The agent now keeps watching the restored workload process after restore
completion. If that process exits or becomes a zombie, the agent terminates the
placeholder PID 1 so Kubernetes can observe the container exit.

The restore-complete sentinel stores the restored PID, so an agent restart after
restore completion can recover the monitor without replaying CRIU.

Problem B is intentionally left for a follow-up, since making placeholder PID 1
transparent when it dies first needs a larger change to the placeholder process
model.

## Summary by CodeRabbit

* **New Features**
  * Restore completion markers now include process information.
  * Restored workloads are monitored after recovery to ensure proper lifecycle handling.
  * Added support for reading and writing custom control-marker data.

* **Bug Fixes**
  * Improved cleanup by terminating the restore placeholder when the restored workload exits.
  * Completed restores can now resume process monitoring using recorded marker data.

### DIFF
--- a/agent/internal/controller/controller.go
+++ b/agent/internal/controller/controller.go
@@ -13,6 +13,8 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"path/filepath"
+	"strconv"
 	"strings"
 	"sync"
 	"syscall"
@@ -65,10 +67,13 @@ type NodeController struct {
 	log                     logr.Logger
 	holderID                string
 	checkpointFn            func(ctx context.Context, params CheckpointParams) error
-	restoreFn               func(context.Context, snapshotruntime.Runtime, logr.Logger, executor.RestoreRequest, executor.RestoreMounter) (int, error)
-	writeControlSentinelFn  func(int, string) error
+	restoreFn               func(context.Context, snapshotruntime.Runtime, logr.Logger, executor.RestoreRequest, executor.RestoreMounter) (*executor.RestoreResult, error)
+	writeControlSentinelFn  func(int, string, []byte) error
 	controlSentinelExistsFn func(int, string) (bool, error)
+	readControlSentinelFn   func(int, string) ([]byte, error)
+	validateProcessStateFn  func(string, int) error
 	sendSignalFn            func(logr.Logger, int, syscall.Signal, string) error
+	restoreMonitorInterval  time.Duration
 	restoreQueue            workqueue.TypedDelayingInterface[client.ObjectKey]
 	restorePodLister        corev1listers.PodLister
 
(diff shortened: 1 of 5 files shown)


Ronkahn21 (changes requested):
Thanks a lot for taking the time to work on this issue and submitting a PR!
However, replacing the kubelet isn't quite the direction we want to take for this project. @shayasoolin is actually already working on a more transparent solution that aligns better with our current roadmap.

We really appreciate your effort and contribution, and we'd love to have you help out on other open issues if you're interested!

shayasoolin:
Hi @caozhuozi, as @Ronkahn21 wrote, we're working on a robust solution for the issue, in which a proxy will be run as the placeholder process (PID 1). This proxy will watch the restored process and act as a passthrough for both exit codes and signals.
We considered an alternative of resolving part of the issue from the node-agent side, but found the proxy solution more complete.
Thank you

PR #129, opened 2026-08-28
feat(api): propagate SnapshotJob PodSnapshot metadata

julienmancuso:
## Summary

- add caller-owned labels and annotations to `SnapshotJob.spec.podSnapshotTemplate`
- validate propagated metadata and reserve SnapshotJob ownership labels before the source Job starts
- copy metadata to the generated `PodSnapshot` while keeping mutable metadata out of identity and crash-window adoption checks

## Validation

- `GOCACHE=/private/tmp/snapshot-pr129-api-go-cache go test ./... -count=1` in `api/`
- `GOCACHE=/private/tmp/snapshot-pr129-operator-go-cache go test ./... -count=1` in `operator/`
- `GOCACHE=/private/tmp/snapshot-pr129-api-go-cache go vet ./...` in `api/`
- `GOCACHE=/private/tmp/snapshot-pr129-operator-go-cache go vet ./...` in `operator/`
- regenerated API and CRDs; output matches aside from the local older `controller-gen` version stamp

`make lint` could not run locally because the installed golangci-lint binary was built with Go 1.24, below this repository's Go 1.26.6 module version.

## Summary by CodeRabbit

* **New Features**
  * SnapshotJob pod templates now support custom labels and annotations.
  * Custom metadata is copied to generated PodSnapshots.
  * Metadata remains isolated when resources are copied or adopted.

* **Bug Fixes**
  * Invalid labels or annotations are rejected before workloads are created.
  * Controller-owned labels cannot be overridden.
  * Existing PodSnapshots with differing or missing metadata are handled safely.


galletas1712 (approved):

operator/internal/controller/snapshotjob_podsnapshot.go
```diff
@@ -283,6 +304,15 @@ func podSnapshotHasExpectedIdentity(actual, desired *snapshotv1alpha1.PodSnapsho
 			return false
 		}
 	}
+	return containsMetadata(actual.Labels, desired.Labels) && containsMetadata(actual.Annotations, desired.Annotations)
```
oleg-kushniriov: The PodSnapshot's labels/annotations are mutable, so requiring them to match the template during crash-window adoption can terminally fail a SnapshotJob whose capture is fine.  for example a mutating webhook rewrote a value between Create and the status write. 

Since ownership labels plus the immutable source-pod UID already prove the object is belonging to SnapshotJob, consider adopting it and patching the template's metadata back on, rather than returning `PodSnapshotNameConflict`.

operator/internal/controller/snapshotjob_podsnapshot.go
```diff
@@ -283,6 +304,15 @@ func podSnapshotHasExpectedIdentity(actual, desired *snapshotv1alpha1.PodSnapsho
...
+}
+
+func containsMetadata(actual, expected map[string]string) bool {
+	for key, value := range expected {
+		if actual[key] != value {
```
oleg-kushniriov: cannot distinguish an empty value from an absent key

operator/internal/controller/snapshotjob_podsnapshot.go
julienmancuso: Good catch. Fixed in 074787e.

Adoption now uses the controller ownership labels plus the immutable source Pod identity (name, UID, and target containers) only. Caller labels and annotations are intentionally not part of identity, so admission-time mutation cannot turn a valid capture into a terminal `PodSnapshotNameConflict`.

I did not patch the metadata back: `podSnapshotTemplate.metadata` has copy-on-create semantics, and rewriting mutable metadata after admission would give the SnapshotJob controller ambiguous field ownership and potentially fight a mutating webhook. Added coverage for both the `AlreadyExists` classification path and the later create/status-write crash window.

operator/internal/controller/snapshotjob_podsnapshot.go
julienmancuso: Fixed in 074787e by removing `containsMetadata` entirely. Mutable labels and annotations no longer participate in adoption identity, so an absent key and an explicitly empty value do not need to be distinguished for this decision.

api/v1alpha1/snapshotjob_types.go
```diff
@@ -134,9 +134,17 @@ type SnapshotJobSpec struct {
...
+// PodSnapshotTemplate mirrors the PodSnapshot fields the user controls. The
+// controller fills in identity, ownership, and spec.source automatically.
+// +kubebuilder:validation:XValidation:rule="!has(self.metadata) || !has(self.metadata.labels) || !('nvidia.com/snapshot-job' in self.metadata.labels)",message="metadata.labels must not set controller-owned label nvidia.com/snapshot-job"
+// +kubebuilder:validation:XValidation:rule="!has(self.metadata) || !has(self.metadata.labels) || !('nvidia.com/snapshot-job-uid' in self.metadata.labels)",message="metadata.labels must not set controller-owned label nvidia.com/snapshot-job-uid"
 type PodSnapshotTemplate struct {
+	// Metadata contains labels and annotations copied to the generated
+	// PodSnapshot. SnapshotJob ownership labels are reserved for the controller
+	// and cannot be supplied here.
+	// +optional
+	Metadata *PodSnapshotTemplateMetadata `json:"metadata,omitempty"`
```
oleg-kushniriov, line 146: Malformed metadata (e.g. a label value over 63 chars) passes admission and only fails at reconcile time as terminal InvalidSpec. Since the spec is immutable, the user can't fix it in place - delete-and-recreate is the only path. 

Consider CEL size guards on podSnapshotTemplate.metadata
- label/annotation key length
-  label value ≤ 63 

so common mistakes are rejected at admission instead.

Also consider adding doc comment warning that invalid metadata terminally fails the job

api/v1alpha1/snapshotjob_types.go
```diff
@@ -157,6 +165,19 @@ type PodSnapshotTemplate struct {
 	TargetContainers []string `json:"targetContainers,omitempty"`
 }
 
+// PodSnapshotTemplateMetadata is the caller-owned metadata propagated to a
+// SnapshotJob's generated PodSnapshot. It intentionally exposes only labels and
+// annotations; identity and ownership remain controller-owned.
+type PodSnapshotTemplateMetadata struct {
+	// Labels are copied to the generated PodSnapshot.
+	// +optional
```
Ronkahn21, line 178: add cel validation here

api/v1alpha1/snapshotjob_types.go
```diff
@@ -157,6 +165,19 @@ type PodSnapshotTemplate struct {
...
+	Labels map[string]string `json:"labels,omitempty"`
+
+	// Annotations are copied to the generated PodSnapshot.
+	// +optional
```
Ronkahn21, line 183: same

api/v1alpha1/snapshotjob_types.go
julienmancuso: Agreed and fixed in 21c6383.

The generated CRD now rejects label and annotation keys that are not valid Kubernetes qualified names and label values longer than 63 characters. The maps have a generous `maxProperties` bound so Kubernetes can statically prove the CEL rules stay within its admission-cost budget. The existing Go validation remains as defense-in-depth for the complete Kubernetes metadata rules, including annotation aggregate size.

I also documented that controller-detected invalid metadata terminally fails the immutable `SnapshotJob` and requires delete/recreate. The CRD was regenerated with the pinned controller-gen, checked with Kubernetes' CRD/CEL validator, and the emitted admission rules are covered by API tests.

api/v1alpha1/snapshotjob_types.go
julienmancuso: Added in 21c6383. The CEL rules live on `PodSnapshotTemplateMetadata` because they validate entries across the labels map: every key must pass Kubernetes `format.qualifiedName()` and every value must be at most 63 characters. The generated schema places those rules directly on `podSnapshotTemplate.metadata`.

api/v1alpha1/snapshotjob_types.go
julienmancuso: Added in 21c6383. Every annotation key must now pass Kubernetes `format.qualifiedName()` at admission. Annotation values remain free-form, as Kubernetes permits; the controller keeps its existing aggregate-size validation as defense-in-depth.

oleg-kushniriov (approved):

Ronkahn21 (commented):
Can you have cel validation to each key and value

api/v1alpha1/crds/embed_test.go
```diff
@@ -134,6 +118,45 @@ func TestSnapshotJobConditionsUseMapListSchema(t *testing.T) {
...
+		"!has(self.labels) || self.labels.all(k, !format.qualifiedName().validate(k).hasValue())",
+		"!has(self.labels) || self.labels.all(k, size(self.labels[k]) <= 63)",
+		"!has(self.annotations) || self.annotations.all(k, !format.qualifiedName().validate(k).hasValue())",
+	}
+	for _, rule := range wantRules {
+		if !containsValidationRule(validations, rule) {
+			t.Errorf("PodSnapshot template metadata validations = %v, want rule %q", validations, rule)
+		}
+	}
+}
```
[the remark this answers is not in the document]
julienmancuso: Good point. Added in e61fd50. The generated-CRD test now asserts both controller-owned label reservation rules at the PodSnapshotTemplate level, in addition to the metadata entry validation. API tests and go vet pass.

Ronkahn21 (approved):

julienmancuso:
CEL now iterates over every metadata entry where Kubernetes defines a per-entry constraint. Label and annotation keys are validated as qualified names, and label-value length is checked at admission. Annotation values are intentionally free-form; Kubernetes only imposes an aggregate 256 KiB annotation-size limit.

I also tested moving full label-value syntax and aggregate annotation-size validation into CEL, but Kubernetes rejects those rules because unbounded map values exceed the static CEL cost budget by more than 100x. Making them admissible would require introducing wrapper value types solely to shape the OpenAPI schema, or manually modifying generated CRDs. The controller therefore retains `ValidateLabels` and `ValidateAnnotations` as the complete validation boundary and reports invalid immutable input as `InvalidSpec`.

PR #97, opened 2026-08-23
feat: refuse incompatible restores before CRIU

leeZardNav:
## Why

A restore onto a machine that cannot run what was captured fails deep inside CRIU, after the
attempt has been claimed and the GPU device map has been built, with an error that names neither
the fact that differs nor the machine it differs on. This adds the comparison that refuses such a
restore up front, records at capture the facts the comparison needs, and reports a refusal in a
way an operator can act on.

Design: #99.

## How to read this

**Commit by commit.** Each of the 30 commits is one idea with its own tests, and the branch is a
ladder: the seam first with an empty policy table, then what a refusal reports, then the switches
that turn it off, then the facts, then one rule per commit, then the condition, then the nightly
scenarios. Nothing on the branch fixes anything else on the branch - every correction lives in the
commit that introduced what it corrects, so a commit can be read once and left behind. Every commit
builds, vets and tests on its own.

| Stage | Commits | What it does |
| --- | --- | --- |
| 1. The seam | 1-4 | The fact model, the mismatch, one reason formatter, and both gates calling a policy table that has no rules yet |
| 2. What a refusal reports | 5-6 | A log line with a stable `reason` field, and a `RestoreIncompatible` pod event |
| 3. Turning it off | 7-9 | The per-restore pod annotation, the per-node `restore.skipCompatCheck`, re-read live per restore and carried to the second gate |
| 4. Recording the facts | 10-16 | GPU model and driver on every discovery path, kernel, architecture, image, image ID and limits, all written into the manifest |
| 5. Gate A rules | 17-21 | Architecture, kernel, image and digest, memory, CPU - everything decidable from the manifest and the pod spec |
| 6. Gate B rules | 22-25 | Mounted paths, GPU model, GPU count, driver - everything that needs the restore node itself |
| 7. The pod condition | 26-27 | The refusal on the `snapshot/Restored` condition, and no re-checking a pod already refused |
| 8. Nightly scenarios | 28-30 | A real artifact's facts against the node and `nvidia-smi`, a real refusal, and both kill switches |

## The two gates

Gate A runs in the pod reconciler before the restore attempt is claimed, on the manifest and the
target pod spec. Gate B runs on the restore node before `BuildDeviceMap`, on what the node
actually offers. A gate B refusal is terminal rather than a CRIU failure: nothing was attempted,
so the placeholder is left alive and the pod is not driven back around the loop.

## Behaviour change

A restore that today fails inside CRIU is now refused before it starts. The refusal is published on
[...]

PR #130, opened 2026-08-28
feat(api): publish restore outcome classifier

julienmancuso:
## Summary

- publish stable `Unknown`, `Pending`, `Succeeded`, `Failed`, and `PartiallySucceeded` restore outcomes
- classify the public `nvidia.com/Restored` Pod condition without defaulting unrecognized statuses or reasons to `Pending`
- make the node agent use the public terminal reason constants and recognize only explicit terminal outcomes

## Validation

- `GOCACHE=/private/tmp/snapshot-outcomes-go-cache go test ./...` in `api/`
- `make linux-test` (all agent packages)
- Linux cross-compilation of `agent/internal/controller`

`make lint` could not run locally because the installed golangci-lint binary was built with Go 1.24, below this repository's Go 1.26.6 module version.

## Summary by CodeRabbit

* **New Features**
  * Added standardized restore status reporting for pending, successful, failed, and partially successful restores.
  * Improved handling of restore outcomes when status conditions are missing, unknown, or incomplete.

* **Bug Fixes**
  * Restore completion detection now consistently reflects the reported restore outcome.
  * Improved reliability when determining whether a restore has reached a terminal state.

### DIFF
--- a/api/v1alpha1/restore_status.go
+++ b/api/v1alpha1/restore_status.go
@@ -1,0 +1,66 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package v1alpha1
+
+import corev1 "k8s.io/api/core/v1"
+
+// RestoreOutcome is the stable, consumer-facing state of a restore request.
+type RestoreOutcome string
+
+const (
+	// RestoreOutcomeUnknown means the Restored condition exists but cannot be
+	// classified safely by this API version.
+	RestoreOutcomeUnknown RestoreOutcome = "Unknown"
+	// RestoreOutcomePending means restore has not reached a terminal outcome.
+	// It includes an absent condition and active restore execution.
+	RestoreOutcomePending RestoreOutcome = "Pending"
+	// RestoreOutcomeSucceeded means every requested destination was restored.
+	RestoreOutcomeSucceeded RestoreOutcome = "Succeeded"
+	// RestoreOutcomeFailed means no requested destination was restored.
+	RestoreOutcomeFailed RestoreOutcome = "Failed"
+	// RestoreOutcomePartiallySucceeded means some, but not all, requested
+	// destinations were restored.
+	RestoreOutcomePartiallySucceeded RestoreOutcome = "PartiallySucceeded"
+)
+
+// Stable reasons used on the nvidia.com/Restored Pod condition. Dependency-wait
+// reasons remain agent-internal; consumers should use ClassifyRestoreOutcome.
+const (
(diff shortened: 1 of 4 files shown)


api/v1alpha1/restore_status.go
```diff
@@ -0,0 +1,61 @@
...
+		}
+		if condition.Status != corev1.ConditionFalse {
+			return RestoreOutcomePending
+		}
+		switch condition.Reason {
+		case RestoreReasonFailed:
+			return RestoreOutcomeFailed
+		case RestoreReasonPartiallySucceeded:
+			return RestoreOutcomePartiallySucceeded
+		default:
```
galletas1712: Right now, any reason this function doesn't recognize is treated as "restore is still in progress." That's fine for "we're waiting for the snapshot."

There's a draft PR (#97) that will add a new *final* reason: `RestoreIncompatible`. That means the node looked at the checkpoint and said no. This code would call that "still in progress," and Dynamo would wait forever instead of starting a new Pod.

You don't have to change this PR for that. When #97 lands, add a case for `RestoreIncompatible`. A test that goes through every `RestoreReason*` and checks the finished ones aren't reported as Pending would keep this from happening again.

galletas1712 (commented):
One thing to watch for when later PRs land — details on the classifier below.

api/v1alpha1/restore_status.go
julienmancuso: Agreed. #130 intentionally classifies only the public reasons that exist today. I am leaving it unchanged as suggested; when #97 lands, `RestoreIncompatible` should be added to the terminal classifier together with an exhaustive test over every exported `RestoreReason*` constant, so a new terminal outcome cannot silently fall back to Pending.

galletas1712 (approved):

api/v1alpha1/restore_status.go
Ronkahn21: I have issue with defaulting to Pending, as it might forgot by other ppl. better to have unkown rather a wrong status. if you really need it you can add it to dynamo and not here

Ronkahn21 (commented):
I have issue to default on restore pending as this is might not true, I raher say I dont know to say it pending

api/v1alpha1/restore_status.go
julienmancuso: Agreed and fixed in 32cb9ce. The public classifier now returns a distinct `RestoreOutcomeUnknown` for any unrecognized condition status or reason, so it no longer asserts `Pending` when this API version cannot know. Missing conditions and explicit `RestoreInProgress` remain Pending. The agent's terminal check now matches only Succeeded, Failed, and PartiallySucceeded; Unknown is not implicitly treated as terminal. Tests cover both an unrecognized dependency reason and the future-terminal `RestoreIncompatible` example.

Ronkahn21 (approved):

PR #131, opened 2026-08-28
feat(api): publish restore Pod contract

julienmancuso:
## Summary

- publish a pure, atomic restore Pod builder and minimum-protocol validator from the self-contained `api/restorepod` package
- require an explicit captured source container while defaulting empty mappings to exactly `source -> source`
- keep startup gating and seccomp shaping optional producer policy; the node agent validates only the minimum restore protocol
- migrate `snapshotctl`, node-agent validation, shared protocol constants, tests, and documentation to the public contract
- keep workload-specific standby behavior consumer-owned

## Release note

The restore Pod Go contract now lives at `github.com/ai-dynamo/snapshot/api/restorepod`. The Snapshot node agent validates annotations, mappings, and snapshot-control resources for one- and multi-container restores, but does not prescribe startup probes or seccomp policy. Producers can opt into the canonical restore-completion startup gate and the Snapshot Helm chart's localhost seccomp profile.

## Validation

- `go test ./...` in `api/`
- `go test ./...` in `operator/`
- `make linux-test` (all agent packages)
- `uv run pytest -m workload` in `e2e/` (5 passed)
- generated API/CRDs with repository-pinned `controller-gen v0.19.0` (no drift)
- `git diff --check`

## Summary by CodeRabbit

- **New Features**
  - Added a standardized Restore Pod contract for building and validating restore workloads.
  - Added source-to-destination container mappings, including multi-container restore fan-out.
  - Added optional startup completion gates and consistent control-volume, environment, probe, and security settings.
- **Bug Fixes**
  - Restore validation now rejects incomplete, conflicting, or malformed Pod configurations earlier.
  - Restore workloads preserve namespaces and use consistent restart behavior without replacing container commands.
- **Documentation**
  - Added comprehensive Restore Pod contract guidance and updated integration requirements, including inert restore container entrypoints.


api/v1alpha1/restore_pod_test.go
```diff
@@ -0,0 +1,346 @@
...
+				RestorePodOptions{SeccompProfile: DefaultSeccompLocalhostProfile},
+			); err == nil {
+				t.Fatal("BuildRestorePod() unexpectedly succeeded")
+			}
+			if !reflect.DeepEqual(pod, before) {
+				t.Fatal("BuildRestorePod() mutated input after failure")
+			}
+		})
+	}
+}
```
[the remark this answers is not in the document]
julienmancuso: Fixed in a90d7a3. The atomic conflict table now includes duplicate destination mappings and verifies that BuildRestorePod rejects them without mutating the input Pod.

api/v1alpha1/restore_pod.go
```diff
@@ -0,0 +1,500 @@
...
+		found, err := hasValidControlEnv(container, name)
+		if err != nil {
+			return err
+		}
+		if !found {
+			return fmt.Errorf("container %q is missing %s environment variable", container.Name, name)
+		}
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
julienmancuso: Fixed in a90d7a3. SNAPSHOT_CONTROL_DIR is required; DYN_SNAPSHOT_CONTROL_DIR is optional and is validated for duplicates or conflicting values when present. Regression coverage verifies both omission and conflict.

api/v1alpha1/restore_pod.go
```diff
@@ -0,0 +1,500 @@
...
+		}
+		probe = probe.DeepCopy()
+	}
+	probe.InitialDelaySeconds = 0
+	probe.PeriodSeconds = 1
+	probe.FailureThreshold = restoreStartupFailureThreshold
+	probe.SuccessThreshold = 1
+	container.StartupProbe = probe
+	return nil
+}
```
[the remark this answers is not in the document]
julienmancuso: Fixed in a90d7a3. The builder no longer substitutes startup, liveness, or readiness handlers. It installs the exact restore-complete exec gate, rejects a nonmatching preexisting startup probe, and the validator enforces both the authoritative handler and canonical timing. Liveness and readiness probes remain unchanged; tests and contract documentation were updated.

docs/restore-pod-contract.md
```diff
@@ -0,0 +1,112 @@
...
+`SNAPSHOT_CONTROL_DIR` environment variable points to that mount. The builder
+also injects the deprecated `DYN_SNAPSHOT_CONTROL_DIR` alias during the
+migration window; hand-authored Pods may omit the alias.
+
+The `restore-complete` sentinel probe shown above is the authoritative startup
+gate. The builder keeps workload liveness and readiness probes unchanged, but
+rejects a preexisting startup probe that does not already match the restore
+gate because Kubernetes cannot compose two startup probes. The extended
+failure threshold prevents kubelet liveness checks from killing the
+placeholder while restore is in progress.
```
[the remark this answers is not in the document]
julienmancuso: Fixed in ae75225. The contract now states that the 1,800-by-one-second budget belongs to the startup probe, that liveness and readiness remain paused until it succeeds, and that exceeding the budget causes kubelet to restart the placeholder according to the Pod restart policy.

api/v1alpha1/restore_pod.go
```diff
@@ -0,0 +1,472 @@
...
+		if env.Value != SnapshotControlMountPath || env.ValueFrom != nil {
+			return false, fmt.Errorf("container %q has conflicting %s environment variable", container.Name, name)
+		}
+	}
+	return found, nil
+}
+
+func ensureRestoreStartupProbe(container *corev1.Container) error {
+	if container.StartupProbe != nil {
+		return validateRestoreStartupProbe(container)
```
galletas1712: If the container already has a startup probe, this returns an error. Every Dynamo worker already has one (an HTTP check on `/live`). So this helper can't build a restore Pod for the Dynamo integration this PR is for.

The old code this PR deletes did the opposite: it kept the existing probe and just checked it more often.

Kubernetes only allows one startup probe. The restore check ("does the restore-complete file exist yet?") should take that slot. The app's `/live` URL doesn't work until restore has finished anyway. Leave the liveness and readiness probes alone — this PR already does that.

Please overwrite a startup probe that isn't the restore-complete check. Don't ask callers to delete the probe before calling this. That's the detail this helper is supposed to handle.

e2e/snapshot_e2e/workloads.py
```diff
@@ -146,12 +146,13 @@ def restore_pod(
     spec["containers"][0]["env"] = [
         {"name": "DYN_SNAPSHOT_RESTORE_STANDBY", "value": "1"},
         {"name": "SNAPSHOT_CONTROL_DIR", "value": CONTROL_DIR},
+        {"name": "DYN_SNAPSHOT_CONTROL_DIR", "value": CONTROL_DIR},
         {"name": RESTORE_TOKEN_ENV, "value": run.restore_token},
     ]
     spec["containers"][0]["startupProbe"] = {
         "exec": {"command": ["/bin/bash", "-lc", f"test -f {RESTORE_DONE}"]},
```
galletas1712: The agent now runs the full restore-Pod check on every restore, including simple one-container ones that used to skip it.

These test Pods still use `test -f` to see if restore finished. The new rules require `cat /snapshot-control/restore-complete`. Same idea, different command, so the agent will reject them.

One-container test Pods will fail a second way: they mount `snapshot-control` with no `subPath`, and the checker now requires `subPath` to be the container's name. The multi-container tests already set that.

The `pytest -m workload` tests don't use a cluster, so they can't catch this. Please update these test Pods (the command, and `subPath: main` on the one-container mount). Making the agent's check looser, as in the other comment, would also fix it. Running the real cluster tests before merge would help.

api/v1alpha1/restore_pod.go
```diff
@@ -0,0 +1,472 @@
...
+	return nil
+}
+
+func validateRestoreStartupProbe(container *corev1.Container) error {
+	probe := container.StartupProbe
+	if probe == nil {
+		return fmt.Errorf("container %q is missing the restore startup gate", container.Name)
+	}
+	expectedCommand := []string{"cat", path.Join(SnapshotControlMountPath, RestoreCompleteFile)}
+	if probe.Exec == nil || !reflect.DeepEqual(probe.Exec.Command, expectedCommand) ||
```
galletas1712: The builder and the agent share the same "this probe must match exactly" check. That's a good check for "did we build the Pod we meant to." It's a bad check for "can we restore a Pod that someone else wrote."

A Pod that uses `test -f` instead of `cat` is still waiting on the same file, but the agent would refuse it. And if we later change the 1800-second limit, Pods built by an older version would start failing on the node.

Please keep the exact check in the builder. For the agent, only check what restore actually needs:

- `snapshot-control` is an emptyDir
- each restore container mounts it at `/snapshot-control` with `subPath` set to that container's name
- `SNAPSHOT_CONTROL_DIR` points there
- a startup probe waits for the restore-complete file

Don't require the exact command string or the exact timeout numbers.

docs/restore-pod-contract.md
```diff
@@ -0,0 +1,114 @@
...
+Kubernetes pauses liveness and readiness probes until the startup gate
+succeeds; if restoration exceeds that failure budget, kubelet restarts the
+placeholder according to the Pod's restart policy.
+
+`RestorePodOptions.SeccompProfile` controls Snapshot's pod-level localhost
+profile. An empty value leaves seccomp unmanaged. A destination container must
+not override a requested profile with a conflicting container-level profile.
+
+Snapshot does not modify container commands and does not inject
+`DYN_SNAPSHOT_RESTORE_STANDBY` or any other workload-specific standby setting.
```
galletas1712: This PR and #103 disagree, and they touch the same file: #103 still edits `operator/internal/protocol/restore.go`, which this PR deletes. If this merges first, the usual git fix is to delete the file, which would throw away #103's work without anyone meaning to.

#103 wants Snapshot to own a neutral name (`SNAPSHOT_RESTORE_STANDBY`) and still set it on restore Pods. This PR stops setting it and says the app owner should.

Easy split: keep the names as public constants (so Dynamo has something to set), and keep this PR's choice not to set them here. Snapshot publishes the name; Dynamo sets it. Please agree on that before either PR merges.

galletas1712 (changes requested):
Two things need to change before this can merge:

1. If a container already has a startup probe, replace it with the restore-complete check. Don't return an error. Dynamo workers always have a startup probe, so `BuildRestorePod` can't actually be used on them today.

2. The e2e test Pods don't match what the agent now requires, so tests against a real cluster will fail.

There's also a clash with #103 on the standby environment variable — please agree on that before either PR merges.

One release-note line would help: the agent now checks one-container restore Pods the same way it checks multi-container ones. Pods that used to work may now be rejected.

api/v1alpha1/restore_pod.go
julienmancuso: Fixed in d051f17 without duplicating the structural rules. The builder now uses a private canonical-output validator, while public `ValidateRestorePod` (used by the node agent) enforces the shared stable runtime contract. Runtime validation accepts supported direct `cat` and `test -f` completion gates, including standard absolute executable paths, and deliberately does not pin probe timing. API and agent tests cover the version-tolerant path.

docs/restore-pod-contract.md
julienmancuso: Resolved in d051f17 with the proposed ownership split. Snapshot now exports `RestoreStandbyModeEnv` and deprecated `LegacyRestoreStandbyModeEnv` as public API names, but the generic builder injects neither. Dynamo or another application owner sets the convention its entrypoint supports. This also makes the #103 rebase boundary explicit: retain the public constants, but do not restore generic operator injection or the deleted protocol file.

api/v1alpha1/restore_pod.go
julienmancuso: Fixed in d051f17. `BuildRestorePod` now replaces any destination startup probe with Snapshot's canonical `cat /snapshot-control/restore-complete` gate while preserving liveness and readiness probes. The test starts with an existing HTTP `/live` startup probe and verifies both replacement in the returned Pod and immutability of the input Pod.

e2e/snapshot_e2e/workloads.py
julienmancuso: Fixed in d051f17. The one-container manifest now mounts `snapshot-control` with `subPath: main`, and both one- and multi-container restore manifests use the canonical `cat /snapshot-control/restore-complete` gate. I also added a cluster-free manifest-contract test covering both shapes; all five workload tests pass. I did not run the GPU-backed cluster E2E locally.

api/v1alpha1/restore_pod.go
```diff
@@ -0,0 +1,509 @@
...
+// is never mutated, including when validation fails. Workload-specific standby
+// behavior and container commands remain the caller's responsibility. Mapping
+// sources must come from the referenced PodSnapshot; this pure builder performs
+// no API read to derive them.
+func BuildRestorePod(
+	pod *corev1.Pod,
+	snapshotName string,
+	mappings []RestoreContainerMapping,
+	options RestorePodOptions,
+) (*corev1.Pod, error) {
```
Ronkahn21: this is  should not live under api, I would arge we need to make sure its belong here
my first instinct its not as not specific to snapshot how the pod looks like

api/v1alpha1/restore_pod.go
```diff
@@ -0,0 +1,509 @@
...
+			return fmt.Errorf("restore pod has no destination container named %q", mapping.Destination)
+		}
+		if err := validateControlMount(container); err != nil {
+			return err
+		}
+		if err := validateControlEnvironment(container); err != nil {
+			return err
+		}
+		if err := validateStartupProbe(container); err != nil {
+			return err
```
Ronkahn21: this is not a requirement, it should be optional

api/v1alpha1/restore_pod.go
```diff
@@ -0,0 +1,509 @@
...
+	return validatePodSeccompProfile(&pod.Spec, options.SeccompProfile)
+}
+
+func validateRestorePodRequest(snapshotName string, mappings []RestoreContainerMapping) (string, string, error) {
+	snapshotName, err := validateRestoreFromSnapshotName(snapshotName)
+	if err != nil {
+		return "", "", err
+	}
+	if len(mappings) == 0 {
+		return "", "", fmt.Errorf("restore container mapping must contain at least one destination")
```
Ronkahn21: support empty mapping of soruce and target are the same this is also optimal

api/podcontract/restore_pod.go
```diff
@@ -0,0 +1,509 @@
...
+func canonicalRestoreStartupProbe() *corev1.Probe {
+	return &corev1.Probe{
+		ProbeHandler: corev1.ProbeHandler{Exec: &corev1.ExecAction{
+			Command: []string{"cat", path.Join(SnapshotControlMountPath, RestoreCompleteFile)},
+		}},
+		TimeoutSeconds:   1,
+		PeriodSeconds:    1,
+		FailureThreshold: restoreStartupFailureThreshold,
+		SuccessThreshold: 1,
+	}
```
Ronkahn21, line 449: this is optional

Ronkahn21 (changes requested):

api/v1alpha1/restore_pod.go
julienmancuso: @Ronkahn21 wdyt about about this boundary:

- Snapshot documents the normative minimum restore contract.
- Snapshot exposes a public restorepod helper for required mechanics:
- annotations, mappings, control volume, mounts, subpaths, and validation.
- Startup gate and seccomp shaping are explicit options, not universal requirements.
- Dynamo opts into the startup gate and adds its standby behavior.
- The helper may live in a dedicated public package rather than directly alongside CRD types.

galletas1712: @Ronkahn21 the seccomp profile to disable io_uring is an example of something that can be classified as "pod shaping" but is snapshot-specific (CRIU does not support it). Anyone who is using snapshot would need it.

In general, the pod needs to be obey some requirements in order to be compatible snapshot and we have helpers for that. Previously it was kind of duplicated between Dynamo and Snapshot and now I think we are trying to consolidate it in one place. Not sure how best to do this though.

Ronkahn21: I understand why, I accept it, lets keep it here, but lets move it from api v1aplhpa
its maybe contract but not API package should be the minimal import we could go for as users might just want to import this

julienmancuso: we could put it under `api/restorepod` pkg, wdyt ?

api/v1alpha1/restore_pod.go
julienmancuso: Implemented in d9ae80c. The startup gate is opt-in through `Options.EnableStartupGate`. Zero-value options preserve an existing startup probe, the public minimum validator and node agent do not inspect probes, and `snapshotctl` explicitly opts into the canonical gate.

api/v1alpha1/restore_pod.go
julienmancuso: Follow-up in d9ae80c: the separation is now stricter. The builder retains a private exact check for an opted-in canonical startup gate, while public `restorepod.Validate` and the node agent ignore startup probes entirely and enforce only the minimum restore protocol.

api/podcontract/restore_pod.go
julienmancuso: Implemented in d9ae80c. Seccomp shaping remains optional: an empty `Options.SeccompProfile` leaves the Pod security context untouched, and minimum runtime validation does not inspect seccomp. When a profile is supplied, the builder injects and validates it.

api/v1alpha1/restore_pod.go
julienmancuso: Implemented in d9ae80c. `Request` carries the captured `SourceContainer`; empty `Mappings` normalize to exactly one identity mapping (`source -> source`). It never infers mappings for every Pod container.

api/v1alpha1/restore_pod.go
julienmancuso: Implemented in d9ae80c. The helper, request/options, mapping type and parser, restore annotations, control-volume vocabulary, and standby names now live in `api/restorepod`. The package directly imports only Kubernetes core API/apimachinery and has no `v1alpha1` dependency; all in-repo consumers were migrated.

api/podcontract/restore_pod.go
```diff
@@ -0,0 +1,488 @@
...
+
+const restoreStartupFailureThreshold int32 = 1800 // 30 minutes at 1s cadence.
+
+// Request identifies the PodSnapshot source and its restore destinations.
+// Empty Mappings default to restoring SourceContainer into a same-named
+// destination container.
+type Request struct {
+	SnapshotName    string
+	SourceContainer string
+	Mappings        []ContainerMapping
```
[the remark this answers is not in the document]
julienmancuso: These are plain Go request/configuration types, not Kubernetes API or CRD fields, so kubebuilder optional/default markers do not apply. SnapshotName, SourceContainer, and mapping source/destination are semantically required; optional zero-value behavior for Mappings and Options is documented and enforced by the builder. controller-gen produces no schema for this package. No change planned.

api/v1alpha1/constants.go
```diff
@@ -3,11 +3,10 @@
...
 
-// Snapshot control-plane contract: the labels, annotations, and control-volume
-// vocabulary the operator stamps and the node agent reads (and that external
-// consumers such as the restore webhook and the workload depend on). This is the
-// single versioned home for these constants so both sides pin them from the api
-// module.
+import "github.com/ai-dynamo/snapshot/api/restorepod"
+
+// Snapshot control-plane labels and status vocabulary shared by the operator
+// and node agent. Restore Pod protocol constants live in api/restorepod.
```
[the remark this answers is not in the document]
julienmancuso: This is an intentional breaking Go API change; compatibility aliases will not be added. This PR establishes api/restorepod as the sole public home for the restore-Pod protocol. Retaining aliases under api/v1alpha1 would preserve the package coupling and duplicate API surface that this refactor removes. The API module is pre-1.0, the release note identifies the new import path, and the known downstream Dynamo consumer will migrate in lockstep.

api/v1alpha1/constants.go
```diff
@@ -94,5 +48,5 @@ const (
 
 	// CUDAJobFilePath is the full stable path to the persisted CUDA checkpoint
 	// job file.
-	CUDAJobFilePath = SnapshotControlMountPath + "/" + CUDAJobFileName
+	CUDAJobFilePath = restorepod.SnapshotControlMountPath + "/" + CUDAJobFileName
```
Ronkahn21: could move this const also the new package, maybe need better name for the package (not podRestore)

agent/internal/controller/controller.go
oleg-kushniriov, line 1194: The terminal set (Succeeded/Failed/PartiallySucceeded) is knowledge the API owns, but every consumer has to hand-roll it as a switch, which silently breaks if a terminal outcome is ever added. Exporting `func (o RestoreOutcome) Terminal() bool `next to the enum keeps that decision in the package that maintains the enum.
Consider this
```
func (o RestoreOutcome) Terminal() bool {
    switch o {
    case RestoreOutcomeSucceeded, RestoreOutcomeFailed, RestoreOutcomePartiallySucceeded:
        return true
    default:
        return false
    }
}
```



```suggestion
func isRestoreTerminal(pod *corev1.Pod) bool {
    return snapshotv1alpha1.ClassifyRestoreOutcome(pod.Status.Conditions).Terminal()
}
```

operator/internal/protocol/restore.go
```diff
@@ -1,223 +0,0 @@
...
-			foundRestoreStandbyModeEnv := false
-			for i := range container.Env {
-				if container.Env[i].Name == RestoreStandbyModeEnv {
-					container.Env[i].Value = "1"
-					container.Env[i].ValueFrom = nil
-					foundRestoreStandbyModeEnv = true
-					break
-				}
-			}
-			if !foundRestoreStandbyModeEnv {
```
oleg-kushniriov, line 104: we previously set DYN_SNAPSHOT_RESTORE_STANDBY to be 1
We are not anymore, is it intentional?

api/podcontract/restore_pod.go
```diff
@@ -0,0 +1,506 @@
...
+	if err != nil {
+		return err
+	}
+	if !found {
+		container.Env = append(container.Env, corev1.EnvVar{Name: name, Value: SnapshotControlMountPath})
+	}
+	return nil
+}
+
+func validateControlEnvironment(container *corev1.Container) error {
```
oleg-kushniriov, line 404: If the env var is intended as a forcing function so every restored workload can discover the control dir portably, keeping it in the minimum is defensible - say so in `docs/restore-pod-contract.md`.

if not, should it move to Build's canonical validation and out of `validateMinimumRestorePod`, alongside the startup gate and seccomp?

oleg-kushniriov (approved):

api/v1alpha1/constants.go
julienmancuso: Agreed—addressed in eb74b9a. I renamed the package to `api/podcontract` and made it self-contained for Snapshot-managed Pods: the shared control-volume vocabulary, capture sentinel and CUDA job-file constants, restore builder/validator, and restored-condition outcome API now live there. `api/v1alpha1` is left with CRD/control-plane vocabulary only.

operator/internal/protocol/restore.go
julienmancuso: Yes, this is intentional. `DYN_SNAPSHOT_RESTORE_STANDBY` is Dynamo workload behavior rather than a generic Snapshot restore requirement, so the public builder no longer injects it or changes command/args. The producer must provide an inert restore entrypoint or set the standby variable its workload supports; `podcontract` exports both the canonical and deprecated names for that purpose. This is an intentional breaking change for manifests that relied on `snapshotctl` injecting the Dynamo-specific value, and the restore-contract documentation calls out the new producer responsibility.

agent/internal/controller/controller.go
julienmancuso: Implemented in 95d112b. `RestoreOutcome.Terminal()` now owns the terminal-outcome set, with coverage for every current outcome and an unknown future value. The node agent delegates directly to that method.

api/podcontract/restore_pod.go
julienmancuso: Documented in 95d112b. The canonical `SNAPSHOT_CONTROL_DIR` variable is intentionally part of minimum validation: it gives every Snapshot-managed workload and tool a portable way to discover the control directory instead of hard-coding the mount path. The builder injects it; hand-authored Pods must provide it. The deprecated alias remains optional.

api/podcontract/protocol.go
```diff
@@ -0,0 +1,165 @@
...
+	// RestoreCompleteFile is written by the Snapshot agent when restore has
+	// completed and the workload may resume.
+	RestoreCompleteFile = "restore-complete"
+)
+
+// ContainerMapping maps the one captured source container to a restore
+// destination in the target Pod.
+type ContainerMapping struct {
+	Source      string
+	Destination string
```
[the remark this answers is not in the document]
julienmancuso: These fields belong to a plain Go protocol type, not a Kubernetes API/CRD schema, so kubebuilder markers do not apply. Both fields are semantically required and validated by `ValidateContainerMappings`; `// +optional` would be incorrect, and there is no meaningful default. `controller-gen` does not generate a schema for `api/podcontract`. No change planned.

api/podcontract/restore_status.go
```diff
@@ -1,7 +1,7 @@
 // SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-package v1alpha1
+package podcontract
```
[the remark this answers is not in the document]
julienmancuso: Documented in 409a972. This remains an intentional breaking Go API migration with no compatibility aliases. `docs/restore-pod-contract.md` now explicitly identifies both former package locations, the new `api/podcontract` import path, the moved annotations/mappings/status/outcome APIs, and that CRD types remain in `api/v1alpha1`. The PR release note also identifies `api/podcontract` as the new contract home.

api/podcontract/restore_pod.go
```diff
@@ -0,0 +1,506 @@
...
+func ensurePodSeccompProfile(spec *corev1.PodSpec, expected string) error {
+	if expected == "" {
+		return nil
+	}
+	if spec.SecurityContext == nil {
+		spec.SecurityContext = &corev1.PodSecurityContext{}
+	}
+	if spec.SecurityContext.SeccompProfile == nil {
+		spec.SecurityContext.SeccompProfile = localhostSeccompProfile(expected)
+		return nil
```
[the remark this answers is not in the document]
julienmancuso: Implemented in 409a972. `Options.SeccompProfile` is now applied and canonically validated on mapped restore destination containers only. Existing Pod-level policy and unrelated sidecars are preserved. Tests cover single-destination and fan-out shaping, idempotency, Pod-level `RuntimeDefault` preservation, sidecar preservation, conflicting destination profiles, and the unmanaged zero-value path.

api/podcontract/restore_pod.go
```diff
@@ -0,0 +1,492 @@
...
+func formatContainerMappings(mappings []ContainerMapping) (string, bool) {
+	if len(mappings) == 1 && mappings[0].Source == mappings[0].Destination {
+		return "", false
+	}
+	formatted := make([]string, 0, len(mappings))
+	for _, mapping := range mappings {
+		formatted = append(formatted, mapping.Source+"="+mapping.Destination)
+	}
+	return strings.Join(formatted, ","), true
+}
```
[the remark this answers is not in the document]
julienmancuso: Implemented in 62a9bb4. `formatContainerMappings` now sorts formatted entries before joining, so logically equivalent mapping sets emit a stable annotation regardless of request order. The fan-out idempotency test now supplies reversed mappings and verifies canonical output. `go test ./...` and `make -C api lint` pass locally.

Ronkahn21 (approved):

galletas1712 (approved):

api/podcontract/restore_pod.go
```diff
@@ -0,0 +1,494 @@
...
+	"fmt"
+	"path"
+	"reflect"
+	"sort"
+	"strings"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+const restoreStartupFailureThreshold int32 = 1800 // 30 minutes at 1s cadence.
```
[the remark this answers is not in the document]
julienmancuso: The 30-minute ceiling and kubelet restart behavior are already documented in `docs/restore-pod-contract.md`. The fixed `1800` threshold intentionally preserves the existing `operator/internal/protocol` behavior during this lightweight API migration; making it configurable would be a new policy/API improvement and is deferred to the broader startup-gate follow-up. The `cat` image dependency is real, but addressing it properly belongs with that same follow-up rather than expanding this migration MR. No change planned here.

PR #138, opened 2026-08-30
feat: clean up PodSnapshotContent artifacts

Ronkahn21:
## Summary

- add the dedicated SnapshotContentReconciler and seed/enforce a metadata-only artifact cleanup finalizer
- remove complete content UID roots from the shared PVC and reclaim orphan roots with an authoritative paged API scan
- gate new capture/restore work on cached deletion observations and UID-guard late capture status writes
- mount artifact storage in the operator and expose scanner configuration, defaulting the scan interval to 10m and orphan grace to 5m
- add API, operator, agent, Helm, and end-to-end cleanup coverage

## Preserved behavior

- capture Lease acquisition, renewal, release, holder identity, tests, dependencies, and agent Lease RBAC remain in place
- the operator manager leader-election Lease remains unchanged
- the agent DaemonSet rollout strategy remains unchanged; this PR does not set maxSurge
- direct content deletion does not introduce ContentDeleted parent status handling

## Validation

- go test ./... in api and operator
- make linux-test
- golangci-lint for api, operator, and Linux-targeted agent packages
- make generate and CRD mirror verification
- make verify-license-headers
- helm unittest charts/snapshot
- helm lint and default/OpenShift template rendering
- uv cluster-free E2E workload tests: 4 passed
- Python syntax validation for the new live E2E cases

Live Kubernetes/GPU E2E requires the CI target cluster and was not run locally. This PR remains draft pending Diffx review of the implementation diff.

## Summary by CodeRabbit

- **New Features**
  - Added automatic cleanup of snapshot artifacts when snapshot content is deleted.
  - Added periodic reclamation of orphaned artifact storage.
  - Added configurable cleanup interval, batch size, and retry attempts.
  - Added OpenShift permissions and deployment support for artifact cleanup.
- **Bug Fixes**
  - Restores now fail early when referenced snapshots are being deleted.
  - Improved validation prevents unsafe artifact paths and containment escapes.
- **Validation**
  - PodSnapshot resources now require exactly one pod container.


charts/snapshot/templates/operator-deployment.yaml
```diff
@@ -45,3 +49,18 @@ spec:
         - name: manager
           image: "{{ .Values.image.operator.repository }}:{{ .Values.image.operator.tag | default .Chart.AppVersion }}"
           imagePullPolicy: {{ .Values.image.operator.pullPolicy }}
+          securityContext:
```
oleg-kushniriov, line 53: Operator unconditionally runs as root, overriding the image's deliberate nonroot user.
runAsUser: 0 on the manager container reverts the image's intentional USER 65532 hardening for every install and makes the operator pod inadmissible under the restricted Pod Security Standard, all to support one filesystem-cleanup path.

If root is genuinely required because CRIU artifacts are root-owned on an RWX PVC, consider scoping it: run cleanup in a separate minimal container/deployment, or keep the manager at 65532 and add only what deletion needs. If broad root was chosen deliberately for simplicity, document that tradeoff in values.yaml and the chart README.

operator/internal/controller/snapshotcontent_reconciler.go
```diff
@@ -0,0 +1,73 @@
...
+	controllerutil.RemoveFinalizer(content, PodSnapshotContentArtifactCleanupFinalizer)
+	return ctrl.Result{}, kubeClient.Patch(ctx, content, client.MergeFromWithOptions(before, client.MergeFromWithOptimisticLock{}))
+}
+
+func removeArtifactRoot(basePath, contentUID string) error {
+	root, err := artifact.ResolveContentRoot(basePath, contentUID)
+	if err != nil {
+		return err
+	}
+	if err := artifact.ValidateDirectory(root); err != nil && !errors.Is(err, os.ErrNotExist) {
```
oleg-kushniriov, line 86: When `artifacts/<uid>` is a symlink or regular file, `removeArtifactRoot` errors forever and the finalizer is never removed, so `PodSnapshotContent` deletion - and transitively the parent PodSnapshot and its namespace - is blocked until a human strips the finalizer, with only an operator log line as evidence.
(When called from reconcileSnapshotContent)

agent/internal/controller/controller.go
```diff
@@ -539,6 +539,9 @@ func (w *NodeController) getPodSnapshotFromPod(ctx context.Context, pod *corev1.
 
 func validatePodSnapshotForRestore(snapshot *snapshotv1alpha1.PodSnapshot) error {
 	key := client.ObjectKeyFromObject(snapshot)
+	if snapshot.DeletionTimestamp != nil {
```
oleg-kushniriov, line 542: (1) capture gating - only restore got the deletion gate; reconcilePodSnapshotContent never checks content.DeletionTimestamp, so an agent can start a capture on a content mid-cleanup, wasting the checkpoint work and leaving a root for the scanner to collect; 

(2) UID-guarded status writes - setSnapshotContentSucceeded/markCheckpointReady still patch purely by name with nothing verifying the content is the same object the capture started against; 

(3) the 5m orphan grace - `ArtifactCleanupConfig` has no grace field; the scanner deletes an unrecognized root on the first scan that sees it.

Are all those point by design?

operator/internal/controller/artifact_scanner.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+func (s *artifactOrphanScanner) listExistingUIDsOnce(ctx context.Context) (map[types.UID]struct{}, error) {
+	uids := make(map[types.UID]struct{})
+	continueToken := ""
+	snapshotResourceVersion := ""
+	for {
+		list := &metav1.PartialObjectMetadataList{}
+		list.SetGroupVersionKind(snapshotv1alpha1.GroupVersion.WithKind("PodSnapshotContentList"))
+		options := &client.ListOptions{Raw: &metav1.ListOptions{Limit: podSnapshotContentMetadataListPageLimit, Continue: continueToken, ResourceVersion: ""}}
```
oleg-kushniriov:
```suggestion
		options := &client.ListOptions{
    Limit:    podSnapshotContentMetadataListPageLimit,
    Continue: continueToken,
    Raw:      &metav1.ListOptions{ResourceVersion: ""},
}
```

The values from `client.ListOptions` always override Limit and Continue fields of Raw

agent/internal/controller/controller.go
```diff
@@ -575,6 +578,9 @@ func (w *NodeController) getPodSnapshotContentFromSnapshot(ctx context.Context,
 }
 
 func validatePodSnapshotContentForRestore(content *snapshotv1alpha1.PodSnapshotContent) error {
+	if content.DeletionTimestamp != nil {
```
oleg-kushniriov, line 581: Because `handleRestorePreflightError` sends non-pending errors to `failRestorePod` regardless of `restoreInProgress`, the new `DeletionTimestamp` checks abort in-progress restore recovery even while the artifacts are still on disk waiting for the content finalizer to run.

If aborting recovery on a deleting dependency is intended (the artifacts are doomed anyway), a comment on the checks would make that explicit.
if not, gate the deletion checks on !restoreInProgress(pod) so only new restores fail early, matching the PR description's "fail early" wording.

operator/internal/controller/snapshotcontent_reconciler.go
```diff
@@ -0,0 +1,73 @@
...
+			return ctrl.Result{}, nil
+		}
+		before := content.DeepCopy()
+		controllerutil.AddFinalizer(content, PodSnapshotContentArtifactCleanupFinalizer)
+		return ctrl.Result{}, kubeClient.Patch(ctx, content, client.MergeFromWithOptions(before, client.MergeFromWithOptimisticLock{}))
+	}
+	if !controllerutil.ContainsFinalizer(content, PodSnapshotContentArtifactCleanupFinalizer) {
+		return ctrl.Result{}, nil
+	}
+	if err := removeArtifactRoot(basePath, string(content.UID)); err != nil {
```
oleg-kushniriov, line 58: when removeArtifactRoot invoked - we may be running in-flight restore on another node - the pod restore will fail. Do we tolerate this behaiviour?

charts/snapshot/values.yaml
```diff
@@ -16,6 +16,12 @@ image:
 
 replicaCount: 1
 
+operator:
+  artifactCleanup:
```
oleg-kushniriov, line 20: Consider adding documentation for new fields

operator/internal/controller/artifact_scanner.go
```diff
@@ -0,0 +1,178 @@
...
+			continue
+		}
+		candidates[name] = struct{}{}
+	}
+	return candidates, nil
+}
+
+func (s *artifactOrphanScanner) listExistingUIDs(ctx context.Context) (map[types.UID]struct{}, error) {
+	var lastErr error
+	for attempt := 1; attempt <= s.config.ListAttempts; attempt++ {
```
oleg-kushniriov, line 135: consider adding a short sleep between attempts

agent/pkg/artifact/artifact.go
```diff
@@ -0,0 +1,99 @@
...
+package artifact
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+const DirectoryName = "artifacts"
```
oleg-kushniriov: private?

charts/snapshot/templates/operator-deployment.yaml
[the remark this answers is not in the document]
Ronkahn21: Not changing this in the cleanup feature. Resource requests and limits are chart-wide sizing policy; the scanner's bounded work does not justify introducing new arbitrary manager defaults.

charts/snapshot/templates/operator-deployment.yaml
```diff
@@ -45,3 +49,18 @@ spec:
...
+            runAsUser: 0
+            runAsGroup: 0
```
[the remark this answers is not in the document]
Ronkahn21: Addressed in dfab56b. The manager must remain UID/GID 0 to remove root-owned CRIU artifacts on the shared PVC, but it now disables privilege escalation, drops all capabilities, and uses RuntimeDefault seccomp.

charts/snapshot/templates/operator-deployment.yaml
```diff
@@ -45,3 +49,18 @@ spec:
...
+          args:
+            - {{ printf "--snapshot-storage-base-path=%v" .Values.storage.pvc.basePath | quote }}
+            - {{ printf "--artifact-cleanup-scan-interval=%v" .Values.operator.artifactCleanup.scanInterval | quote }}
+            - {{ printf "--artifact-cleanup-batch-size=%v" .Values.operator.artifactCleanup.batchSize | quote }}
+            - {{ printf "--artifact-cleanup-list-attempts=%v" .Values.operator.artifactCleanup.listAttempts | quote }}
```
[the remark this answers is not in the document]
Ronkahn21: Not applicable to the agreed design: orphan grace was intentionally removed. The scanner deletes immediately after a complete authoritative absence check, and the E2E uses the configured scan interval.

operator/internal/controller/artifact_scanner.go
```diff
@@ -0,0 +1,178 @@
...
+	processed := 0
+	for uid := range candidates {
+		if _, protected := existing[types.UID(uid)]; protected {
+			continue
+		}
+		if processed == s.config.BatchSize {
+			break
+		}
+		processed++
+		if err := removeArtifactRoot(s.config.BasePath, uid); err != nil {
```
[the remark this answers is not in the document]
Ronkahn21: By design, there is no orphan grace period. Immediate deletion after a complete authoritative metadata list is the agreed convergence behavior.

charts/snapshot/templates/operator-deployment.yaml
Ronkahn21: Addressed in dfab56b. Root remains required for root-owned CRIU artifacts, so this does not split cleanup into a separate deployment. The template now documents the reason and applies compatible hardening: no privilege escalation, no capabilities, and RuntimeDefault seccomp.

operator/internal/controller/snapshotcontent_reconciler.go
Ronkahn21: Addressed in dfab56b. Unsafe UID roots remain fail-closed, but the controller now emits a Warning ArtifactCleanupBlocked Event on the deleting PodSnapshotContent with the remediation path, while retaining the finalizer.

agent/internal/controller/controller.go
Ronkahn21: Capture deletion gating is now addressed in dfab56b: deleting content is skipped before source-Pod promotion, content selection, Lease acquisition, and checkpoint launch. UID-aware late status writes remain intentionally out of scope to preserve the requested minimal agent flow. Orphan grace was explicitly removed by design.

operator/internal/controller/artifact_scanner.go
Ronkahn21: Keeping Raw ListOptions intentionally. controller-runtime documents that Limit and Continue are ignored when Raw is set; Raw carries Limit, Continue, and resourceVersion="" together for this authoritative paginated list.

agent/internal/controller/controller.go
Ronkahn21: Intentional for this release. We do not add a separate mid-flight lifecycle path: cleanup may race restore, and restore may finish or fail normally. The change only prevents a restore from starting when its initial dependency observation is deleting.

operator/internal/controller/snapshotcontent_reconciler.go
Ronkahn21: Yes, this is tolerated by design. Cleanup does not wait for or coordinate with restore; an overlapping restore may succeed or fail normally as its artifacts disappear.

charts/snapshot/values.yaml
Ronkahn21: Addressed in dfab56b with concise inline documentation for scan interval, batch size, and list-attempt settings.

operator/internal/controller/artifact_scanner.go
Ronkahn21: Keeping retries immediate and bounded. Sleeping would only block the leader-elected scanner; the scheduled scan interval already provides spaced retry after the three in-run attempts are exhausted.

agent/pkg/artifact/artifact.go
Ronkahn21: Addressed in dfab56b: DirectoryName is now package-private; callers use the artifact resolution helpers.

operator/internal/controller/artifact_scanner.go
oleg-kushniriov: The values of Limit and Continue are overridden by values in client.ListOptions which are not initialized - meaning you get Limit - 0, Continue- "".

agent/pkg/artifact/artifact.go
```diff
@@ -0,0 +1,99 @@
...
+func ResolveContentRoot(basePath, contentUID string) (string, error) {
+	artifactsRoot, err := ResolveRoot(basePath)
+	if err != nil {
+		return "", err
+	}
+	if err := ValidatePathElement("PodSnapshotContent UID", contentUID); err != nil {
+		return "", err
+	}
+	root := filepath.Join(artifactsRoot, contentUID)
+	if err := ValidateContainment(artifactsRoot, root); err != nil {
```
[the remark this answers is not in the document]
Ronkahn21: Addressed in f6a356a. removeArtifactRoot now validates the artifacts ancestor with Lstat before resolving or recursively removing a content root. The regression test makes artifacts a symlink to an external directory and proves the target remains untouched, the finalizer is retained, and ArtifactCleanupBlocked is emitted. Concurrent replacement after validation remains outside this release's descriptor-anchored traversal scope.

operator/internal/controller/artifact_scanner.go
Ronkahn21: Correct — my earlier reply inverted controller-runtime's flattening behavior. Addressed in f6a356a: Limit and Continue are now set on client.ListOptions, while Raw carries only resourceVersion="". Tests assert the first page limit and each continuation token from the direct fields.

oleg-kushniriov (approved):

PR #99, opened 2026-08-23
📝 add restore compatibility checker design

leeZardNav:
## Summary
- add the proposed restore compatibility checker design to the snapshot repository
- incorporate the review decisions for Option C, the `devices` selector, Pod conditions, and explicit record/publish/compare contracts
- defer snapshot-agent compatibility to a follow-up with an explicit format-breaking version map

## Validation
- `git diff --check`
- docs-only change; no runtime tests required

## Summary by CodeRabbit

* **Documentation**
  * Added a proposed design for checking snapshot restore compatibility before restoration proceeds.
  * Documents validation of host, pod, GPU, and volume conditions across two restore checkpoints.
  * Defines failure handling, events, status reporting, logging, retry prevention, security, reliability, and observability considerations.
  * Covers testing, rollout, migration, performance, risks, alternatives, and deferred checks.
  * Documents optional per-pod and per-node controls for bypassing validation.

PR #140, opened 2026-08-31
feat: refuse incompatible restores before CRIU

### LINKED ISSUE #36: Validate restore target pod spec, gpu arch/model, container name
## Summary

Restore should detect when the target restore pod uses a different container
image than the pod/container that was checkpointed, and warn or fail before
attempting CRIU restore.

During manual Snapshot testing, several restore failures were ultimately caused
by using different images for the source checkpoint pod and the restore target
pod. The resulting errors looked like low-level CRIU, tar, mount, or device
problems, but they disappeared once the source and restore pod images matched.

## Why this matters

Snapshot restore assumes the target container rootfs and runtime environment are
compatible with the checkpointed process. If the restore target image differs
from the source image, the restore may fail in confusing ways, or worse, restore
against a subtly incompatible filesystem layout.

This is easy for users to get wrong, especially when using a placeholder image
for restore. Today the failure mode points users toward CRIU/mount debugging
instead of the simpler root cause: the restore target is not using the same
image/runtime rootfs as the checkpoint source.

## Observed symptoms

Examples seen while testing:

```text
INFO nsrestore runtime/overlay.go:174 Applying rootfs diff {"target": "/"}
tar: ./var/run/secrets: Cannot mkdir: Invalid cross-device link
tar: ./var/run/secrets/kubernetes.io: Cannot mkdir: Invalid cross-device link
tar: Exiting with failure status due to previous errors
ERROR nsrestore nsrestore/main.go:50 restore failed {"error": "rootfs diff failed: tar extract failed: exit status 2"}
```

Other failures also appeared as CRIU mount/device restore problems until the
source and restore images were aligned.

## Proposal

Record the source container image identity at checkpoint time and validate it
before restore starts.

The most useful comparison would be the resolved image ID/digest from container
status, not only `spec.containers[*].image`, because tags can move. The human
error message should still include both the source and target image strings.

This should reduce time spent debugging misleading low-level restore errors and
make the restore contract clearer.

leeZardNav:
## Summary

- Record the host, pod, external mount path, and GPU environment with each checkpoint and compare it before restore work begins.
- Refuse incompatible restores with a stable reason on `snapshot/Restored`, in events, and in logs.
- Support per-pod and node-level compatibility bypasses.
- Integrate with parallel multi-container restore: Gate A checks every destination, while Gate B reports through the coordinator so workers do not race on Pod status.

Closes #36. Part of #231.

Design: #99.

## How to read this

The 29 commits form a ladder; each commit introduces one idea with its own tests.

| Stage | Commits | What it does |
| --- | --- | --- |
| 1. The seam | 1-4 | Environment model, stable mismatch reason, and the two policy gates with an empty rule table |
| 2. Refusal logging | 5 | One log message with a stable `reason` field |
| 3. Turning it off | 6-8 | Per-pod annotation and node setting |
| 4. Recording the environment | 9-15 | GPU, kernel, architecture, image, image ID, and resource limits in the manifest |
| 5. Gate A rules | 16-20 | Architecture, kernel, image and digest, memory, and CPU checks for every restore destination |
| 6. Gate B rules | 21-24 | External mount destination-path existence, GPU model, GPU count, and driver checks using node-local values |
| 7. Reporting | 25-26 | Coordinator-owned refusal conditions/events and terminal refusal handling |
| 8. E2E scenarios | 27-29 | Real artifact values, a memory refusal, and the annotation bypass |

## Behavior

Gate A runs before the restore is claimed. Gate B runs after a destination placeholder is resolved but before its device map or CRIU work. The `mount` category checks whether each CRIU-recorded external mount destination path exists in the placeholder rootfs; it does not verify that the destination is a mount point or backed by the same volume. A single-destination refusal reports:

```
Type               Status  Reason                Message
snapshot/Restored  False   RestoreIncompatible   memory-limit: source 4Gi, target 1Gi
```

For a multi-destination restore, all incompatible destinations are aggregated under `RestoreIncompatible`; mixed worker outcomes use the existing partial/failure result handling. Checkpoints missing either side of a comparison remain restorable.

The bypasses are off by default: `nvidia.com/snapshot-skip-compat-check: "true"` skips one Pod, and `config.restore.skipCompatCheck` skips checks node-wide for every restore that agent handles.

## Validation

- [x] `make test` for api, agent, and operator on every commit
- [x] `go vet ./...` for api, agent, and operator on every commit
- [x] `GOOS=linux GOARCH=amd64 make check` on the tip
- [x] Linux agent and operator builds on the tip
- [x] `helm unittest charts/snapshot` — 2 suites, 4 tests
- [x] E2E collection — 24 tests; workload helper tests — 5 passed
- [ ] Re-run the GPU E2E scenarios on the recreated tip

## Summary by CodeRabbit

- **New Features**
  - Restore compatibility checks now compare host architecture, kernel, images, resource limits, mounts, GPUs, and drivers.
  - Incompatible restores are refused with clear conditions, events, and mismatch details, and can be retried after correction.
  - Checkpoint metadata records host, image, runtime image ID, resource, and GPU information.
  - GPU discovery includes model and driver details with improved fallback handling.
  - Compatibility checks can be bypassed through configuration or restore-pod annotation.
- **Documentation**
  - Documented the compatibility-check bypass annotation and its boolean parsing.
- **Bug Fixes**
  - Runtime image-ID and kernel-version read failures no longer unnecessarily stop processing.

Supersedes #135 after removing the internal task reference from the head branch.


agent/internal/executor/restore.go
```diff
@@ -291,10 +333,27 @@ func inspectRestore(
...
+//
+// Only a path that is definitely absent is left out. Any other stat failure is
+// this agent failing to look rather than the pod missing a volume, and reporting
+// it as missing would refuse a restore that would have worked.
+func existingMounts(targetRoot string, destinations []string) []string {
+	existing := make([]string, 0, len(destinations))
+	for _, destination := range destinations {
+		if _, err := os.Stat(filepath.Join(targetRoot, destination)); !os.IsNotExist(err) {
+			existing = append(existing, destination)
+		}
```
hhzhang16: You say this is looking for mounts but it seems to only be looking for path existence

api/compat/checks.go
```diff
@@ -0,0 +1,320 @@
...
+var gpuModelCheck = check{
+	name: CheckGPUModel,
+	gate: GateInspect,
+	compare: func(source, target Facts) []Mismatch {
+		sourceModels, sourceOK := gpuModels(source.GPUDevices)
+		targetModels, targetOK := gpuModels(target.GPUDevices)
+		if !sourceOK || !targetOK || sourceModels == targetModels {
+			return nil
+		}
+		return []Mismatch{{Source: sourceModels, Target: targetModels}}
```
hhzhang16: Does device order matter here? e.g. for heterogenous GPU allocations

leeZardNav: that's a good question. i remember during my investigation that i found the order does matter, but your comments makes me think i should re validate it
answer soon

leeZardNav: Order does not matter for this compatibility check, it only verifies the counted inventory of GPU models. 
however, there can be an issue in CUDA if someone send different ordered map to restore.
I opened [RUN-42577](https://runai.atlassian.net/browse/RUN-42577) to make that ordered following this new check, so thank you. but in here for compatibility, no need to check the order from what i can find.
lemme know if you feel differently

agent/internal/executor/restore.go
leeZardNav: Correct - the probe checks whether CRIU’s recorded mount destination paths exist, not mount identity. I renamed the internal helper and fact in 7ded0e5
 mount remains the public category because these paths come from CRIU’s external mount map.

agent/internal/cuda/cuda.go
```diff
@@ -159,43 +204,85 @@ func discoverGPUUUIDs(
...
 	if len(gpuUUIDs) > 0 {
-		return gpuUUIDs, nil
+		// This path has its GPUs already and needs nvidia-smi only to describe
+		// them, so a failure here costs facts, not the checkpoint.
+		visible, err := discoverVisibleGPUs(ctx, hostProcPath, pid)
+		if err != nil {
+			log.V(1).Info("Failed to describe PodResources GPUs; recording their UUIDs alone",
+				"pid", pid,
+				"error", err,
+			)
```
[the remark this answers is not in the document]
leeZardNav: Leaving this at V(1): the PodResources path still records UUIDs and intentionally treats unavailable optional facts as unknown; this is an observability preference rather than a restore correctness issue.

agent/internal/executor/restore.go
```diff
@@ -248,20 +254,37 @@ func inspectRestore(
 	}
 	log.V(1).Info("Resolved placeholder container", "pid", placeholderPID)
 
+	targetImageID := ""
+	if manifest.K8s.ImageID != "" {
+		if req.ContainerID == "" {
+			return nil, 0, fmt.Errorf("container ID is required to compare the runtime image ID")
+		}
```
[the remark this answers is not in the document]
leeZardNav: The controller resolves a non-empty container ID before calling Restore, and executor is an internal package; the pod-lookup fallback is not used by the production path, so I’m leaving it unchanged here.

agent/internal/executor/restore.go
```diff
@@ -291,10 +333,27 @@ func inspectRestore(
...
+// this agent failing to look rather than the pod missing a volume, and reporting
+// it as missing would refuse a restore that would have worked.
+func existingMounts(targetRoot string, destinations []string) []string {
+	existing := make([]string, 0, len(destinations))
+	for _, destination := range destinations {
+		if _, err := os.Stat(filepath.Join(targetRoot, destination)); !os.IsNotExist(err) {
+			existing = append(existing, destination)
+		}
+	}
+	return existing
```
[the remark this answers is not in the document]
leeZardNav: os.Stat returns a PathError that os.IsNotExist handles here, so this would be an idiom-only change with no behavior difference; leaving it unchanged.

api/compat/checks.go
```diff
@@ -0,0 +1,320 @@
...
+
+// mustMatch reports a mismatch unless the two values are identical. A value
+// absent on either side is unknown, and an unknown fact never refuses a restore:
+// a checkpoint captured before it was ever recorded has to stay restorable.
+func mustMatch(source, target string) []Mismatch {
+	if source == "" || target == "" || source == target {
+		return nil
+	}
+	return []Mismatch{{Source: source, Target: target}}
+}
```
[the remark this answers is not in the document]
leeZardNav: The production inputs are already normalized at collection by ReadKernelVersion, nvidia-smi parsing, runtime.GOARCH, and imageDigest, so this only affects malformed hand-written manifests; leaving it unchanged.

api/compat/compat.go
```diff
@@ -0,0 +1,154 @@
...
+	DriverVersion string
+	GPUDevices    []GPUDevice
+
+	// ExternalizedMounts holds the mount destinations CRIU externalized at
+	// capture.
+	ExternalizedMounts []string
+
+	// ExistingMountPaths holds the destinations that resolve on this machine.
+	// The agent resolves them before comparing, so a comparison never touches disk.
+	ExistingMountPaths []string
```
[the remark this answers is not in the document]
leeZardNav: api/compat/compat.go is added by this unmerged PR and ExistingMounts has never existed on main, so there is no released field to preserve.

leeZardNav:
Regarding CodeRabbit’s out-of-diff RestoreIncompatible E2E comment: valid diagnostic improvement, but it only shortens a timeout when a test expected another outcome and does not affect restore behavior, so I’m leaving it out of this PR.

leeZardNav:
Regarding CodeRabbit’s out-of-diff ContainerMapping marker comment: ContainerMapping is an in-memory protocol type, not a CRD field, and has no JSON schema, so kubebuilder optional/default markers do not apply.

api/compat/checks.go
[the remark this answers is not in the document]
leeZardNav: Both remaining paths normalize before persistence: [`ReadKernelVersion`](https://github.com/ai-dynamo/snapshot/blob/7ded0e53c1a867e3ff4ae67904f87be34bd904c1/agent/internal/runtime/process.go#L22-L34) trims the kernel value and rejects an empty result, while [`parseNvidiaSmiGPUFacts`](https://github.com/ai-dynamo/snapshot/blob/7ded0e53c1a867e3ff4ae67904f87be34bd904c1/agent/internal/cuda/cuda.go#L128-L149) trims the UUID, product name, and driver version. Production kernel and driver facts therefore cannot retain surrounding whitespace, so this remains limited to malformed hand-written manifests.

api/compat/checks.go
hhzhang16: Can you make the RUN-42577 fix a follow-up MR and link it here?

agent/internal/executor/restore.go
hhzhang16: If “destination path exists” is the compatibility contract, can you update the PR’s claims that it verifies the volume is mounted?

api/compat/checks.go
leeZardNav: sure, once we'll get to it i will. we haven't prioritize it yet, so i don't have a definitive ETA, but i'll add a comment to the task to update here once it does

agent/internal/executor/restore.go
leeZardNav: Updated

agent/cmd/agent/main.go
```diff
@@ -38,6 +38,13 @@ func main() {
 	if err := cfg.Validate(); err != nil {
 		fatal(agentLog, err, "Invalid configuration")
 	}
+	// A host fact that cannot be read is unknown, never fatal: the node keeps
```
Ronkahn21: please remvoe the word fact, it not true, its does not reflect the infomation

agent/cmd/agent/config.go
```diff
@@ -33,6 +35,30 @@ func LoadConfig(path string) (*types.AgentConfig, error) {
...
 }
 
+// NewSkipCompatCheckFn returns a per-restore read of the node-wide switch off
+// the mounted ConfigMap, so an admin who flips it does not have to roll the
+// DaemonSet to be heard. Kubernetes projects ConfigMap updates into the mount
+// on its own; nothing here watches or polls.
+//
+// A read that fails keeps the last value it did get: a restore is never failed,
+// and never quietly checked differently, because a config read went wrong.
+func NewSkipCompatCheckFn(path string, initial bool, log logr.Logger) func() bool {
```
Ronkahn21: keep the confiuraion load at start up, as the all the other confiugraion is right now 
keep the same style as current config

agent/cmd/agent/main.go
```diff
@@ -59,7 +66,8 @@ func main() {
 	)
 
 	// The node controller handles both restore and capture paths.
-	nodeController, err := controller.NewNodeController(cfg, rt, rootLog.WithName("controller"))
+	nodeController, err := controller.NewNodeController(cfg, rt, rootLog.WithName("controller"),
+		NewSkipCompatCheckFn(ConfigMapPath, cfg.Restore.SkipCompatCheck, agentLog))
```
Ronkahn21: also agentLog as this per worlload best to pass the log in the contetxt containing the pod req information 
best practice for controller

agent/internal/controller/compat.go
```diff
@@ -0,0 +1,143 @@
...
+		"reason", reason,
+	)
+}
+
+// reopenedAfterRefusal reports a pod that the gates turned down and that has
+// since asked for them to be skipped. Nothing else reopens a terminal restore,
+// which is what makes the skip request an escape hatch and not a retry.
+func (w *NodeController) reopenedAfterRefusal(pod *corev1.Pod) bool {
+	condition := findRestoredCondition(pod)
+	if condition == nil || condition.Status != corev1.ConditionFalse || condition.Reason != podcontract.RestoreReasonIncompatible {
```
Ronkahn21: can we export to helper function

agent/internal/controller/compat.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+	return w.skipCompatCheckRequested(pod)
+}
+
+// podFacts reads what one container of a pod runs as and is allowed. It serves
+// both sides of a comparison: what a capture records about the source pod, and
+// what a restore target offers.
+//
+// A container that is not in the pod leaves its facts unknown.
+func podFacts(pod *corev1.Pod, containerName string) compat.Facts {
```
Ronkahn21: again lets remove the facts suffix we need better name

agent/internal/controller/compat.go
```diff
@@ -0,0 +1,143 @@
...
+	if pod == nil {
+		return compat.Facts{}
+	}
```
Ronkahn21: if there is no way to pass nil here, remove the check

agent/internal/controller/compat.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+// limitString keeps an unset limit unset. A missing quantity formats as "0",
+// which would otherwise read as a container limited to nothing.
+func limitString(limits corev1.ResourceList, name corev1.ResourceName) string {
+	quantity, ok := limits[name]
+	if !ok {
+		return ""
+	}
+	return quantity.String()
```
Ronkahn21, line 92: what is the string for non exists ?

agent/internal/controller/compat.go
```diff
@@ -0,0 +1,143 @@
...
+// A nil error means the restore may be attempted.
+func (w *NodeController) preflightCompatibility(
+	pod *corev1.Pod,
+	artifact *restoreArtifact,
+	mappings []podcontract.ContainerMapping,
+) error {
+	log := w.log.WithValues("pod", fmt.Sprintf("%s/%s", pod.Namespace, pod.Name), "container", artifact.SourceContainerName)
+	if artifact.SkipCompatCheck {
+		log.Info("Restore compatibility check skipped by request", "gate", string(compat.GatePreflight))
+		return nil
```
Ronkahn21: this I am not understanding how come the artifact is the one is auth for skiping compatibility check

agent/internal/controller/compat.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	manifest, err := types.ReadManifest(artifact.Path)
```
Ronkahn21, line 114: it seem types in not the package this should sitting

agent/internal/controller/compat.go
```diff
@@ -0,0 +1,143 @@
...
+	if err != nil {
+		// An unreadable manifest is not an incompatibility. The restore path
+		// reads it again and reports the real error from there, so refusing here
+		// would relabel a broken artifact as an incompatible one.
+		log.V(1).Info("Skipping restore compatibility gate; checkpoint manifest is unreadable",
+			"artifact_path", artifact.Path,
+			"error", err.Error(),
+		)
+		return nil
```
Ronkahn21, line 128: I dont like we silencing if we failed to read, at least lets add event here, or retrun error that we ingonore in caller function

agent/internal/controller/controller.go
```diff
@@ -72,6 +73,12 @@ type NodeController struct {
 	sendSignalFn            func(logr.Logger, int, syscall.Signal, string) error
 	restoreQueue            workqueue.TypedDelayingInterface[client.ObjectKey]
 	restorePodLister        corev1listers.PodLister
+	compareFn               func(compat.Gate, compat.Facts, compat.Facts) []compat.Mismatch
+
+	// skipCompatCheckFn is read once per restore rather than at startup, so the
+	// node-wide switch can be flipped without a DaemonSet rollout. Injected so
+	// the controller never learns where the config file lives.
+	skipCompatCheckFn func() bool
```
Ronkahn21: not needed, as we allow annotation.
also it is not consistent with all the other config

agent/internal/controller/controller.go
```diff
@@ -90,6 +97,9 @@ type restoreArtifact struct {
 	ContentUID          string
 	SourceContainerName string
 	Path                string
+	// SkipCompatCheck is decided once in preflight and carried from there, so
+	// the gate inside the restore reaches the same answer as the one before it.
+	SkipCompatCheck bool
```
Ronkahn21: can we remove all the comments that does not add value

agent/internal/controller/controller.go
```diff
@@ -423,11 +444,14 @@ func (w *NodeController) processRestoreQueueItem(ctx context.Context, key client
 		return
 	}
 	pod = pod.DeepCopy()
-	if w.restoreHandled(pod) {
+	if w.reopenedAfterRefusal(pod) {
```
Ronkahn21: can you change the name of the function, its does not explain what does it mean
reopen

agent/internal/controller/controller.go
```diff
@@ -508,6 +532,12 @@ func (w *NodeController) preflightRestore(ctx context.Context, pod *corev1.Pod)
 	if err != nil {
 		return nil, err
 	}
+	// Gate A: the earliest point the checkpoint's own record of what it was
+	// captured on is readable, and still before any of the restore is attempted.
+	artifact.SkipCompatCheck = w.skipCompatCheckRequested(pod)
```
Ronkahn21: this does not sound true, the artifact dose not the authority the pod. setting it here will confuse other

agent/internal/executor/checkpoint.go
```diff
@@ -188,6 +194,10 @@ func inspectContainer(ctx context.Context, rt snapshotruntime.Runtime, log logr.
 	if err != nil {
 		return nil, 0, fmt.Errorf("failed to resolve container: %w", err)
 	}
+	imageID, err := rt.ResolveContainerImageID(ctx, containerID)
```
oleg-kushniriov, line 200: image ID only feeds the image-digest compat rule, which already treats a blank value as unknown.
Is it intentional to make it a fail hard?

A transient CRI error should cost the fact, not the checkpoint, the same way the kernel version and nvidia-smi facts are handled.

agent/internal/executor/restore.go
```diff
@@ -310,20 +316,37 @@ func inspectRestore(
 	}
 	log.V(1).Info("Resolved placeholder container", "pid", placeholderPID)
 
+	targetImageID := ""
+	if manifest.K8s.ImageID != "" {
```
oleg-kushniriov: inspectRestore fails the restore when ResolveContainerImageID errors, regardless of req.SkipCompatCheck, so the bypass does not actually bypass the new code path and a CRI hiccup becomes a RestoreFailed that has nothing to do with CRIU. The lookup should be skipped when the gate is skipped and should degrade to an unknown fact on error.

consider moving this code in `!req.SkipCompatCheck` gate (or into inspectCompatibility)

api/compat/checks.go
```diff
@@ -0,0 +1,325 @@
...
+const (
+	minKernelMajor = 5
+	minKernelMinor = 13
+)
+
+var kernelVersionCheck = check{
+	name: CheckKernelVersion,
+	gate: GatePreflight,
+	compare: func(source, target Facts) []Mismatch {
+		return mustMatch(source.KernelVersion, target.KernelVersion)
```
oleg-kushniriov, line 46: In the life of a node, cloud providers roll out patched images constantly. A hard check of the full kernel string will fail on any cross-node restore during such rollout, and also on the same node after it reboots into the new kernel. The restore then fails with `kernel-version` reason.

Is this what we really want, or can we relax the validation to major/minor version?

Also, today the only way around it is `skipCompatCheck`, which turns off all rules, including CPU arch and memory. If we keep exact match, maybe the bypass should allow to skip only this one rule.

api/podcontract/protocol.go
```diff
@@ -22,6 +23,9 @@ const (
 	// uses the captured container name as the destination.
 	RestoreContainerMapAnnotation = "nvidia.com/restore-container-map"
 
+	// SkipCompatCheckAnnotation disables compatibility checks for one restore.
+	SkipCompatCheckAnnotation = "nvidia.com/snapshot-skip-compat-check"
```
oleg-kushniriov, line 27: Add this annotation to api.md

agent/internal/controller/controller.go
```diff
@@ -670,33 +700,50 @@ func (w *NodeController) restorePodContainers(ctx context.Context, pod *corev1.P
 // recordRestoreResults publishes the aggregate Pod outcome after every worker
 // in the current pass has returned.
 func (w *NodeController) recordRestoreResults(ctx context.Context, pod *corev1.Pod, artifact *restoreArtifact, results []restoreResult) bool {
-	byState := make(map[restoreResultState][]string, 3)
+	byState := make(map[restoreResultState][]string, 4)
```
Ronkahn21: why 4?

agent/internal/controller/controller.go
```diff
@@ -670,33 +700,50 @@ func (w *NodeController) restorePodContainers(ctx context.Context, pod *corev1.P
...
+	if len(failed) == 0 {
+		return w.finishRestore(ctx, pod, corev1.ConditionFalse, podcontract.RestoreReasonIncompatible, strings.Join(incompatibilityReasons, "; ")) != nil
+	}
+	if len(incompatible) != 0 {
+		message := fmt.Sprintf("Restore failed for %d destination container(s) and refused %d incompatible destination(s) from PodSnapshot %s", len(failed), len(incompatible), artifact.SnapshotName)
+		return w.finishRestore(ctx, pod, corev1.ConditionFalse, podcontract.RestoreReasonFailed, message) != nil
+	}
 	message := fmt.Sprintf("Restore failed for all %d destination container(s) from PodSnapshot %s: %s", len(failed), artifact.SnapshotName, strings.Join(failed, ", "))
 	return w.finishRestore(ctx, pod, corev1.ConditionFalse, podcontract.RestoreReasonFailed, message) != nil
 }
```
Ronkahn21, line 791: can we simlify this function its look too hevy now seapaecliy all the if ...

agent/internal/controller/controller.go
```diff
@@ -726,6 +773,13 @@ func (w *NodeController) restoreDestination(
 	emitPodEvent(ctx, w.clientset, log, pod, snapshotEventComponent, corev1.EventTypeNormal, restoreRequestedReason, fmt.Sprintf("Restore requested from PodSnapshot %s for destination %s", artifact.SnapshotName, destination))
 
 	if err := w.runRestore(ctx, pod, artifact, destination, containerID, startedAt, recovering); err != nil {
+		var incompatible *compat.IncompatibleError
+		if errors.As(err, &incompatible) {
+			result.state = restoreResultIncompatible
+			result.reason = compat.Reasons(incompatible.Mismatches)
+			w.logRestoreRefusal(pod, incompatible, result.reason)
+			return result
```
Ronkahn21, line 841: some humen message could be nice

agent/internal/controller/controller.go
```diff
@@ -807,6 +861,14 @@ func (w *NodeController) runRestore(ctx context.Context, pod *corev1.Pod, artifa
 
 	placeholderHostPID, err := op.executeRestore(restoreCtx)
 	if err != nil {
+		var incompatible *compat.IncompatibleError
+		if errors.As(err, &incompatible) {
+			// Nothing was attempted, so there is no half-restored process to
+			// clean up. The placeholder is deliberately left running: killing it
+			// would restart the container straight back into the same answer.
+			return incompatible
```
Ronkahn21, line 928: remove

agent/internal/cuda/cuda.go
```diff
@@ -87,11 +94,15 @@ func GetPodGPUUUIDs(ctx context.Context, podName, podNamespace, containerName st
...
-// PodResources API does not report GPU devices (e.g. when GPUs are allocated
-// via DRA instead of the NVIDIA device plugin).
-func GetGPUUUIDsViaNvidiaSmi(ctx context.Context, hostProcPath string, pid int) ([]string, error) {
+// DiscoverVisibleGPUFacts describes the GPUs a container can see, by running
+// nvidia-smi inside its mount and PID namespaces. The model and the driver
+// version come from the same call as the UUIDs: nothing else on the restore path
+// gets to look at the source node's GPUs, so what is not read here cannot be
+// compared later.
+func DiscoverVisibleGPUFacts(ctx context.Context, hostProcPath string, pid int) (compat.GPUFacts, error) {
+	ctx, cancel := context.WithTimeout(ctx, nvidiaSMITimeout)
```
Ronkahn21: can we just move the timeout to be argument

agent/internal/controller/compat.go
```diff
@@ -0,0 +1,143 @@
...
+
+	sourceFacts := manifest.CompatFacts()
+	for _, mapping := range mappings {
+		mismatches := w.compareFn(
+			compat.GatePreflight,
+			sourceFacts,
+			w.preflightTargetFacts(pod, mapping.Destination),
+		)
+		if len(mismatches) != 0 {
+			return compat.NewIncompatibleError(compat.GatePreflight, mismatches)
```
oleg-kushniriov, line 139: `preflightCompatibility` returns on the first destination with mismatches, and the reason does not say which destination failed. Gate B in `recordRestoreResults` collects all destinations and prefixes each reason with the destination name.

Can we do the same here: collect across all destinations instead of returning on the first, and name the destination when there is more than one? Then both gates report the same shape.

agent/internal/cuda/cuda.go
```diff
@@ -87,11 +94,15 @@ func GetPodGPUUUIDs(ctx context.Context, podName, podNamespace, containerName st
...
-// container's mount and PID namespaces. This is the fallback path when the kubelet
-// PodResources API does not report GPU devices (e.g. when GPUs are allocated
-// via DRA instead of the NVIDIA device plugin).
-func GetGPUUUIDsViaNvidiaSmi(ctx context.Context, hostProcPath string, pid int) ([]string, error) {
+// DiscoverVisibleGPUFacts describes the GPUs a container can see, by running
+// nvidia-smi inside its mount and PID namespaces. The model and the driver
+// version come from the same call as the UUIDs: nothing else on the restore path
+// gets to look at the source node's GPUs, so what is not read here cannot be
+// compared later.
+func DiscoverVisibleGPUFacts(ctx context.Context, hostProcPath string, pid int) (compat.GPUFacts, error) {
```
Ronkahn21: for the DRA path what are doing

Ronkahn21 (changes requested):
1/2 review

agent/cmd/agent/config.go
leeZardNav: this is an issue, as our kill switch won't work if we won't refetch it, meaning users that want to use he kill switch will need to reset the agent. are you sure this is what we want? 
this isn't just style at this point, and i would argue this might been an oversight in prev repo, though maybe this has downside i'm not aware of?

agent/internal/executor/checkpoint.go
leeZardNav: Done in d6bb5fd.

agent/internal/executor/restore.go
leeZardNav: Done in 5055761.

agent/internal/controller/compat.go
leeZardNav: Done in b73019d.

agent/internal/controller/compat.go
leeZardNav: Done in f9727bb.

api/podcontract/protocol.go
leeZardNav: Done in 5908680.

agent/internal/controller/compat.go
leeZardNav: Done in b76a645.

agent/internal/controller/controller.go
leeZardNav: Done in b76a645.

agent/internal/controller/compat.go
leeZardNav: Done in 049f9c9.

agent/internal/controller/controller.go
leeZardNav: Done in 049f9c9.

agent/internal/controller/controller.go
leeZardNav: Done in 658b089, then simplified further in 6e983d7 after another read.

The chain of `if len(...) != 0` is now one `switch` in `verdict`, which takes a tally and a snapshot name and returns the status, reason and message without touching the pod, so every outcome is one case and the whole policy is readable top to bottom. `recordRestoreResults` is down to a tally, a verdict and one of two writes.

6e983d7 removed `reportRestoreProgress`, which was taking the context, the pod, the snapshot name and the tally just to build one message and write one condition. `verdict` now concludes a pass whether or not it is over, so the in-progress message is built by the same pure function as the terminal ones and the table test covers it too.

agent/internal/controller/controller.go
leeZardNav: Done in d1ff9ab.

agent/internal/cuda/cuda.go
leeZardNav: Done in 11e720a, and the wrapper it left behind is gone in c0b2e9e: the discovery seam takes the timeout too, so DiscoverVisibleGPUs is passed as itself and nvidiaSMITimeout is named once, where the pass starts.

agent/cmd/agent/main.go
leeZardNav: Done in 5509dd6 and e5286df.

agent/internal/controller/controller.go
leeZardNav: Swept in 7e044dd: the comments that only restated the next line are gone, and the longer ones are cut to the clause that carries the reason. b76a645 had only removed this one, as a side effect of moving the field.

agent/internal/controller/controller.go
leeZardNav: Trimmed in 5509dd6. I kept one line of it, that leaving the placeholder running is deliberate, because killing it restarts the container straight back into the same refusal and that is the part the code cannot show. Happy to drop that too if you would rather.

agent/internal/controller/compat.go
leeZardNav: Done in e5286df.

agent/internal/cuda/cuda.go
leeZardNav: Done in 4258205.

api/compat/checks.go
leeZardNav: Good question. We went strict here because of upstream reports rather than a guess: in [criu#2636](https://github.com/checkpoint-restore/criu/issues/2636) a restore that had worked for over a year broke on a kernel change alone, and the maintainer traced it to a single upstream commit. Since a mismatch fails deep inside CRIU with an error that is hard to attribute, we preferred to refuse it up front. The reasoning is in the comment on the kernel-version check.

On the per-check bypass, I agree it would be more useful than the all-or-nothing skip. We have talked about it and it is a good candidate for a follow-up task rather than this one.

agent/internal/controller/compat.go
leeZardNav: it's more a necessity then a statement

Gate B gets the collecting for free. Each destination is restored by its own worker, so every outcome already comes back as a value into one results array, and that array has to be folded into a single Pod condition anyway for succeeded, failed and pending. An incompatible destination rides along with them, and the destination name is prefixed only when there is more than one result.

Gate A has nothing to ride along in. It runs once, before any worker exists, and reports by returning an error, which is the first one by definition. Matching the shape means it stops returning on the first mismatching mapping and accumulates across all of them instead. Doable, but the gain is naming every mismatching destination instead of the first, and on a single-destination restore the two already read the same. Worth it?

leeZardNav:
Review round addressed in d6bb5fd through e5286df, one commit per comment, each replied to on its own thread. `gofmt` clean and the api, agent and operator suites green at every commit.

Two threads are waiting on you rather than done: whether Gate A should collect every destination's mismatches the way Gate B does, and whether the last line of the placeholder comment in `runRestore` should go too.

Stacked [#141](https://github.com/ai-dynamo/snapshot/pull/141) is rebased on this and moved with the rename.

agent/internal/controller/compat.go
leeZardNav: it's not a new function. it was there from prev commit

agent/internal/controller/compat.go
leeZardNav: Renamed to restoreRefused in 79747da, since the reason restated what a refusal already means.

agent/internal/controller/compat.go
leeZardNav: The empty string. `limitString` returns `""` for a limit that is not set, so it never formats as `"0"` and read as a container limited to nothing, and both `mustMatch` and `atLeastSource` treat `""` as unknown, so an unset limit on either side never refuses a restore.

agent/internal/controller/controller.go
leeZardNav: It was the map's size hint, one bucket per `restoreResultState`, and this PR added the fourth: pending, succeeded, failed, incompatible.

That map is gone in 658b089 though, replaced by a struct with a named field per state, so there is no hint left to justify. It also turned out to be the faster shape: a `switch` on the state beats hashing each insert and then looking the four slices back out.

agent/cmd/agent/main.go
leeZardNav: Agreed for the controller paths that handle a workload, but this is not one of them. `NewSkipCompatCheckFn` returns a `func() bool` that takes no context and never sees a pod: the only thing it logs is a failure to re-read the node's own ConfigMap, so there is no pod request to carry here.

Converting the paths that do have one is worth doing, as its own change rather than inside this PR.

agent/internal/controller/controller.go
leeZardNav: Same matter as on `config.go:45`, where I put the other half of it to you: if the switch is only read at startup then using it means restarting the agent, which defeats the point of a kill switch. Probably better settled in that one thread than in two.

On not needing it given the annotation: the annotation covers one pod a user knows is safe, while the node switch is for an operator whose node is refusing restores wrongly and who cannot go and annotate every pod. Both feed the same check, so dropping either one does not simplify the path.

agent/cmd/agent/config.go
Ronkahn21: Yes, I am sure. as we this eventual consistent it adds a layer for edge cases for stall cache and nodes running under different configuration. Users are allow to say I ok with skiping the restore so I feel this is enough

leeZardNav: Done in 1ecd896

agent/internal/controller/controller.go
leeZardNav: Done in 1ecd896 — the injected read is gone. The node setting itself stays, loaded at startup like every other key, per your answer on `config.go:45`.

charts/snapshot/values.yaml
```diff
@@ -212,6 +212,13 @@ config:
   restore:
     # Maximum seconds to allow a restore attempt before the agent marks it failed
     restoreTimeoutSeconds: 7200
+    # Skip the restore compatibility check on every restore this agent handles.
+    # The gate refuses a restore whose checkpoint was captured somewhere the
+    # target node cannot reproduce; turning it off means those restores are
+    # attempted and fail inside CRIU instead. Per-node escape hatch; a single
+    # restore can opt out with the nvidia.com/snapshot-skip-compat-check
+    # annotation on the restore pod.
+    skipCompatCheck: false
```
oleg-kushniriov, line 225: config.restore.skipCompatCheck is loaded only at agent startup, but the DaemonSet has no ConfigMap checksum annotation, so helm upgrade flipping it leaves every running agent on the old value with no signal that a restart is needed.

agent/internal/executor/restore.go
```diff
@@ -333,14 +360,32 @@ func inspectRestore(
 			placeholderPID,
 			log,
 		)
+		discoverDuration = time.Since(discoverStart)
 		if err != nil {
 			return nil, 0, fmt.Errorf("failed to get target GPU UUIDs: %w", err)
```
oleg-kushniriov, line 365: cuda.DiscoverGPUs fails hard before inspectCompatibility runs when the placeholder has no nvidia-smi, so a GPU checkpoint restored into a non-GPU pod still ends as RestoreFailed rather than the gpu-count refusal CheckGPUCount's comment promises.

agent/internal/types/manifest.go
```diff
@@ -206,3 +256,65 @@ func validateArtifactManifest(artifact ArtifactManifest) error {
...
+}
+
+// externalizedMounts returns the destinations CRIU externalized at capture, in a
+// stable order so a refusal always names them the same way.
+func (m *CheckpointManifest) externalizedMounts() []string {
+	if len(m.CRIUDump.ExtMnt) == 0 {
+		return nil
+	}
+	destinations := make([]string, 0, len(m.CRIUDump.ExtMnt))
+	for destination := range m.CRIUDump.ExtMnt {
```
oleg-kushniriov, line 315: externalizedMounts includes masked-path entries whose ExtMnt value is /dev/null; those are not volumes and CRIU restores them from the value side, so the mount rule should skip them the way criuHandledMounts skips / and /dev/shm.

docs/reference/api.md
```diff
@@ -98,6 +98,7 @@ The caller sets these annotations on the new pod to trigger a restore:
 |------------|-------------|
 | `nvidia.com/restore-from` | Names the `PodSnapshot`, in the pod's namespace, to restore into the pod. |
 | `nvidia.com/restore-container-map` | Optional. Comma-separated `source=destination` pairs mapping the single captured container to one or more restore containers. When absent, the captured container name is the destination. |
+| `nvidia.com/snapshot-skip-compat-check` | Optional. Set to `"true"` to attempt the restore without the compatibility checks, which otherwise refuse a checkpoint this node cannot run. The value must parse as a boolean, so any other spelling, including `"yes"`, leaves the checks in place. Treat it as a debugging escape hatch: a restore that should have been refused instead fails somewhere inside CRIU. |
```
oleg-kushniriov: Adding nvidia.com/snapshot-skip-compat-check: "true" to an already-refused pod reopens the restore in place (skipRequestedAfterRefusal), but the api.md row for the annotation does not say so, and the agent's other terminal messages say to create a new pod.

charts/snapshot/values.yaml
leeZardNav: yes. see https://github.com/ai-dynamo/snapshot/pull/140#discussion_r3947814097

agent/internal/types/manifest.go
leeZardNav: there are many such paths. but finding them 1 by 1 is a long task and is in the deferred task. for now we do the basic apple to apple comparison on broad spectrom

docs/reference/api.md
leeZardNav: done in a2399996

agent/internal/executor/restore.go
leeZardNav: Real, but pre-existing: at the merge base `restore.go` already hard-returned `failed to get target GPU UUIDs` on a discovery error, so this ended as RestoreFailed before the checks existed. The comment was the wrong half - corrected in `afba7d6`.

e2e/snapshot_e2e/lifecycle.py
```diff
@@ -5,10 +5,12 @@
 
 from __future__ import annotations
 
+import json
 import shlex
 import time
 from typing import Any, Callable
```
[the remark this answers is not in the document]
leeZardNav: Unchanged by this PR - `from typing import Any, Callable` is identical on the base branch and only entered the diff context because lines above it moved. No ruff configuration in the repo and CI does not run it, and every other module in the suite imports from `typing` - nothing uses `collections.abc` yet. Better as a repo-wide sweep than one odd file out.

e2e/tests/test_snapshot_lifecycle.py
```diff
@@ -239,6 +319,98 @@ def test_failed_restore_gpu_checkpoint_into_non_gpu_target(
...
+        # The placeholder is still the placeholder: a refusal costs no CRIU work,
+        # so the workload never sees restore-complete.
+        assert not snap.file_present(config.namespace, run.restore_pod, snap.RESTORE_DONE)
+    except Exception:
+        snap.debug_dump(config, run)
+        raise
+
+
+# Small enough to be refused, large enough to restore into once the checks are
+# off: with the annotation on, the restore this test starts actually runs.
```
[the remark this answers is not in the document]
leeZardNav: That E2E existed and came out on purpose earlier in this PR. `test_node_switch_lets_a_refused_restore_through_without_a_rollout` asserted that flipping the ConfigMap took effect with no agent rollout - same pod UID, same restart count - and `1ecd896` made the agent read the value once at startup, so the test was pinning behavior that no longer holds.

With a startup-only switch an honest E2E has to bring the agent up with the value already set, which is an install-time chart-value variant of the suite rather than one test: restarting the DaemonSet mid-run changes node-wide state for every test sharing the cluster. Unit-covered in `TestSkipCompatCheckTurnsOffTheGates`.

agent/internal/controller/compat.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	sourceEnv := manifest.CompatEnvironment()
+	for _, mapping := range mappings {
+		mismatches := w.compareFn(
+			compat.GatePreflight,
+			sourceEnv,
+			w.preflightTargetEnvironment(pod, mapping.Destination),
+		)
+		if len(mismatches) != 0 {
```
oleg-kushniriov, line 138: preflightCompatibility's early-return-on-first-mismatch contradicts the PR's own documented Gate A behavior ("checks every destination") and gives multi-container restores strictly worse refusal reporting than Gate B, where an operator can't tell which destination failed or whether others are also broken.

agent/internal/controller/controller.go
```diff
@@ -423,11 +432,14 @@ func (w *NodeController) processRestoreQueueItem(ctx context.Context, key client
...
+	if w.skipRequestedAfterRefusal(pod) {
+		// The skip request is the way back for a pod the gates turned down, so
+		// it has to clear the in-process marker as well as the condition below.
+		w.handledRestores.Delete(string(pod.UID))
+	} else if w.restoreHandled(pod) {
 		requeue = w.removeRestoreFinalizerWithEvent(ctx, pod)
 		return
-	}
-	if isRestoreTerminal(pod) {
+	} else if isRestoreTerminal(pod) {
```
oleg-kushniriov, line 442: `handleTerminalRestorePod` never calls `markRestoreHandled`, so any restore pod refused by the compat gate before an agent restart re-triggers event emission on every subsequent resync indefinitely, and the event it emits carries a generic `RestoreAlreadyFailed` reason instead of the pod condition's actual `RestoreIncompatible` reason - undermining the stable, machine-readable reason this PR is meant to provide for restore refusals.

agent/internal/controller/controller.go
oleg-kushniriov, line 459: handleTerminalRestorePod's default branch collapses a compat-gate refusal and a genuine CRIU failure into the same event reason, defeating the stable, machine-readable reason this PR is meant to provide for restore refusals observed after an agent restart.

agent/internal/controller/compat.go
leeZardNav: Fair point on the mismatch: the doc says "checks every destination" and the loop stops at the first one, so those two don't line up. Here's why the loop works this way.

Three of Gate A's five checks — `cpu-arch`, `kernel-version`, `kernel-minimum` — read the node, not the destination container. When the node fails one of them, every destination fails it in exactly the same way. Collecting those would print the same sentence once per container, and putting a destination name on each copy would suggest the container is at fault for something that belongs to the node. Only `cpu-limit` and `memory-limit` read the container at this gate.

The outcome is also the same either way. A refusal is pod-level and terminal, so one incompatible destination refuses the whole pod whether we stop at the first mismatch or walk the rest first. Stopping early gives the operator the same answer with less to read.

The doc wording is the part that's genuinely off, so I'm happy to fix that line to match the code if you'd prefer.

agent/internal/controller/controller.go
leeZardNav: done in 653a1dc4

agent/internal/controller/controller.go
leeZardNav: This one is intentional, so let me explain the reasoning and where the refusal reason actually lives.

The event reasons in this switch follow `RestoreOutcome`, which is the classifier the API tells consumers to read, and `ClassifyRestoreOutcome` maps both `RestoreFailed` and `RestoreIncompatible` onto the same `RestoreOutcomeFailed`. Since `isRestoreTerminal` gates on `Terminal()`, only three outcomes can reach this function, and the switch has one branch for each. A fourth reason here would be the only event reason with no `RestoreOutcome` behind it.

The refusal reason isn't lost. The condition keeps `Reason: RestoreIncompatible` permanently, and the original refusal event carries that same reason because `finishRestore` emits using the condition's own. This replay then carries `condition.Message`, which for a refusal is the full `Refused restore; this node cannot run the checkpoint: <mismatches>`.

The `RestoreAlready*` events are narrower than the name suggests: they mean "this already concluded, expect nothing more." They're a replay notice, not the record of why it ended that way — that record is the condition.

oleg-kushniriov (approved):

Ronkahn21 (approved):

PR #141, opened 2026-08-31
feat: publish what a checkpoint was captured on

### LINKED ISSUE #233: Show what a checkpoint was captured with - mounts, hardware, software - on the PodSnapshotContent status
## Summary

Surface the environment a checkpoint was captured in on
`PodSnapshotContent.status`, so it is visible through the Kubernetes API.

## Motivation

`PodSnapshotContentStatus` (`api/v1alpha1/podsnapshotcontent_types.go:69`)
currently holds exactly one field: `Conditions`. An operator looking at a
`PodSnapshotContent` can see *whether* it is Ready, and nothing about *what it is* —
not the GPU model, not the driver version, not the mounts it expects, not the image
it came from.

Some of this already exists on disk. `CheckpointManifest`
(`agent/internal/types/manifest.go:21`) records the CRIU mount plan, overlay state,
source node, and source GPU UUIDs — but it is a `manifest.yaml` inside the artifact
directory, readable only by the node agent that has the storage mounted. It is not
in the API, so no controller, no CLI, and no user can see it.

This blocks two things at once: a human cannot answer "will this restore on that
node?" without shelling into storage, and the compatibility rules in #232 have no
API-visible data to compare against. Per the project's own design principle —
everything Snapshot manages is observable through the Kubernetes API — this is a gap.

## Goals

1. Add a provenance section to `PodSnapshotContentStatus` covering three families:
   **mounts** (the CRIU external-mount plan, bind-mount destinations, overlay
   upperdir), **hardware** (GPU model and count, MIG profile, driver version,
   CPU feature summary, memory footprint), and **software** (container image and
   resolved digest, CRIU version, agent version, kernel and distro).
2. Populate it from the node agent at bind time, from data it already collects
   during checkpoint inspection (`CheckpointContainerSnapshot`,
   `agent/internal/types/inspect.go:22`).
3. Version the provenance shape, so an artifact captured by an older agent is
   readable rather than ambiguous. `CheckpointArtifactVersionAnnotation` already
   exists for the artifact itself — decide whether provenance rides that version
   or carries its own.
4. Expose the highest-signal fields as `kubectl` print columns.

**Non-goals:** enforcing any of it (that is #232); changing the on-disk manifest
format beyond what is needed to fill the new fields.

## Acceptance criteria

- `kubectl get podsnapcontent <name> -o yaml` shows mounts, hardware, and software
  provenance for a freshly captured snapshot.
- Fields the agent could not determine are distinguishable from fields it
  determined to be empty.
- The status subresource stays within reasonable size limits for a pod with many
  mounts — bound or summarize the mount list rather than inlining it unbounded.
- An artifact captured before this change surfaces as "provenance unavailable"
  rather than as a partially-populated struct that looks authoritative.

## Design note

Status is the right home rather than spec: this is observed state written by the
agent, and `PodSnapshotContentSpec` is immutable and operator-populated. Keep the
field set additive — #232 will want to add comparisons over time, and a closed
struct will fight that.

leeZardNav:
## Summary

- `PodSnapshotContent` gains an optional `status.source` publishing what a checkpoint was captured on: the node, the container, and the devices.
- The block is projected from the artifact's `manifest.yaml` and written in the same status patch that marks the capture `Ready`, so a content that is Ready already carries its recorded source.
- Device details are grouped under one property per vendor, so each payload stays typed and validated and a second vendor can join without a discriminator or a CEL rule.
- Nothing reads the block back. The restore compatibility gates compare the manifest, so this cannot become a second source of truth.

Closes #233.

Design: [#99](https://github.com/ai-dynamo/snapshot/pull/99), section 7.1. Builds on [#140](https://github.com/ai-dynamo/snapshot/pull/140), now merged, which records what this publishes.

## What it looks like

```yaml
status:
  conditions:
    - type: Ready
      status: "True"
      reason: Captured
  source:
    devices:
      nvidia:
        driverVersion: 580.82.07
        instances:
          - productName: NVIDIA A100-SXM4-80GB
          - productName: NVIDIA A100-SXM4-80GB
    node:
      name: gpu-node-3
      architecture: amd64
      kernelVersion: 5.15.0-1071-aws
    pod:
      image: nvcr.io/nvidia/ai-dynamo/vllm-runtime:0.6.1
      imageDigest: sha256:9f2c...
      memory: 64Gi
```

`instances[].productName` and `driverVersion` are the names the NVIDIA DRA driver publishes on a `ResourceSlice`, so the same GPU reads the same way in both places. The GPU count is the length of the instance list rather than a field of its own, so the two cannot disagree.

## How to read this

The 7 commits form a ladder; each introduces one idea with its own tests.

| Commits | What it does |
| ------- | ------------ |
| 1-3 | The published shape, one block per commit: node, then pod, then vendor-keyed devices. Each regenerates both CRD copies and the deepcopy |
| 4 | The projection from manifest to status as a pure function, reusing `compat.ImageDigest`. Nothing calls it yet |
| 5 | Wires it into the Ready write, at all three call sites |
| 6 | A test per Ready path — fresh dump, artifact recovery, and a work order reconciled after the source pod is gone — plus the conflict retry and the do-not-erase guard |
| 7 | E2E assertion against the node object and nvidia-smi rather than against the manifest |

## Decisions worth a look

**Absent, not empty.** A value the agent could not read is left out, and a block with nothing in it is omitted entirely. An artifact captured before a value was ever recorded still publishes; a reader can tell unknown from a value that happens to be blank.

**The published digest is the compared digest.** `compat.ImageDigest` is exported rather than reimplemented, so the reduction that decides whether two images are the same content is the one this publishes. Runtimes wrap a digest differently and the artifact keeps whichever form it was handed, so a second implementation could show a value the gate never compares.

**One read, not one per retry.** The Ready write retries on optimistic-lock conflict, so the manifest is read once before the loop.

**Losing the source never costs the capture.** By the time this runs the artifact is committed, so an unreadable manifest logs and Ready is written anyway.

**No `node.agentVersion` yet.** The design lists it, but [#140](https://github.com/ai-dynamo/snapshot/pull/140) dropped agent-version recording along with the major/minor comparison, so the field would only ever be empty. It arrives with the deferred agent-version compatibility mapping, which is also what will give it a meaning.

**No new RBAC.** The agent already holds `podsnapshotcontents/status` `update` and `patch`. `PodSnapshot` keeps mirroring only Ready and Failed.

**Single container.** Capture is still one container per content (`PodReference.Containers` is `MinItems=1`/`MaxItems=1`), so one `source` block per content is correct. Multi-container restore does not change that; if multi-container capture arrives, this block is where it grows a key.

## Verification

- Every one of the 7 commits builds, vets and tests on its own (`go vet ./...` and `go test ./...` for the agent under Linux, `go test ./...` for the api module, at each rung).
- Full suites green for api, agent and operator; `make verify-crds` clean, so the chart CRD copy matches the generated one; license headers, fmt, tidy and helm lint clean; govulncheck clean.

### Live validation on real GPU hardware

Re-validated end to end at head `91d500a8` on a freshly provisioned single-T4 cluster, installing this
branch's own published images so the agent and operator under test are built from this head.

| | |
| --- | --- |
| Images | `ghcr.io/ai-dynamo/snapshot/{operator,agent}:v0.0.0-lirane-checkpoint-source-facts-g91d500a8` |
| GPU | Tesla T4, driver `595.71.05` |
| GPU Operator | `v26.3.3` |
| Kubernetes | `v1.33.5`, containerd `2.3.5`, Ubuntu 24.04 |

| Suite | Result |
| --- | --- |
| `pytest e2e/tests -m environment` | 1 passed |
| `pytest e2e/tests/test_snapshot_lifecycle.py` | 10 passed (5m40s) |
| `pytest e2e/tests/test_snapshotjob.py` | 13 passed (4m26s) |

The lifecycle file covers capture, real restore, both refusal gates and the skip-annotation bypass, so
the compatibility work this builds on is exercised in the same run as the publishing added here.

What a captured content published, beside the ground truth read independently from the cluster rather
than from the manifest:

```yaml
devices:
  nvidia:
    driverVersion: 595.71.05
    instances:
    - productName: Tesla T4
node:
  architecture: amd64
  kernelVersion: 7.0.0-1012-aws
  name: ip-172-20-10-235
pod:
  image: ghcr.io/ai-dynamo/snapshot/agent:v0.0.0-lirane-checkpoint-source-facts-g91d500a8
  imageDigest: sha256:563009dbd6dca8836462ec233e4e028bd824136d7995434efdf879a5cd9aa966
  memory: 4Gi
```

```
Node object ip-172-20-10-235:  architecture=amd64  kernelVersion=7.0.0-1012-aws
Pod spec:                      memory limit=4Gi    cpu limit=<never set>
nvidia-smi in the container:   name="Tesla T4"     driver="595.71.05"
```

Two details worth noting there. The product name publishes as `Tesla T4`, the name `nvidia-smi` and the
DRA driver use, not the `Tesla-T4` form the node label carries. And the CPU limit this workload never
set is absent rather than present and empty, which is what comparing the `pod` block whole pins down.

## Summary by CodeRabbit

- **New Features**
  - Pod snapshot status now includes checkpoint source metadata from the committed capture when available.
  - Published metadata covers capture-node details, pod and container information, resource limits, image digests, visible devices, and NVIDIA GPU details.
  - Optional values are omitted, while unknown device fields are preserved for compatibility.

- **Bug Fixes**
  - Snapshots can still reach Ready status when source metadata is unavailable or the manifest cannot be read.
  - Existing published source data is preserved when later captures lack source details.

Supersedes #63 after removing the internal task reference from the head branch.


leeZardNav:
## Live validation on real GPU hardware

Rebased onto `main` now that the stack below this landed, and re-validated end to end on a freshly
provisioned single-T4 cluster. The branch's own published artifacts were installed, so the agent and
operator under test are built from this head.

| | |
| --- | --- |
| Head | `91d500a8` |
| Images | `ghcr.io/ai-dynamo/snapshot/{operator,agent}:v0.0.0-lirane-checkpoint-source-facts-g91d500a8` |
| GPU | Tesla T4, driver `595.71.05` |
| GPU Operator | `v26.3.3` |
| Kubernetes | `v1.33.5`, containerd `2.3.5`, Ubuntu 24.04 |

### Suites

| Suite | Result |
| --- | --- |
| `pytest e2e/tests -m environment` | 1 passed |
| `pytest e2e/tests/test_snapshot_lifecycle.py` | 10 passed (5m40s) |
| `pytest e2e/tests/test_snapshotjob.py` | 13 passed (4m26s) |

The lifecycle file covers capture, real restore, both refusal gates and the skip-annotation bypass, so
the compatibility work this is stacked on is exercised in the same run as the publishing added here.

### The published block, against the ground truth it is compared to

`status.source` on a content whose `Ready` condition is `Captured`:

```yaml
devices:
  nvidia:
    driverVersion: 595.71.05
    instances:
    - productName: Tesla T4
node:
  architecture: amd64
  kernelVersion: 7.0.0-1012-aws
  name: ip-172-20-10-235
pod:
  image: ghcr.io/ai-dynamo/snapshot/agent:v0.0.0-lirane-checkpoint-source-facts-g91d500a8
  imageDigest: sha256:563009dbd6dca8836462ec233e4e028bd824136d7995434efdf879a5cd9aa966
  memory: 4Gi
```

Read independently from the cluster, not from the manifest:

```
Node object ip-172-20-10-235:  architecture=amd64  kernelVersion=7.0.0-1012-aws
Pod spec:                      memory limit=4Gi    cpu limit=<never set>
nvidia-smi in the container:   name="Tesla T4"     driver="595.71.05"
```

Two things worth noting in that output. The product name is published as `Tesla T4`, matching what
`nvidia-smi` and the DRA driver report, not the `Tesla-T4` form the node label carries. And the CPU
limit this workload never set is absent from the published block rather than present and empty,
which is the behaviour the assertion pins by comparing the `pod` block whole.

agent/internal/controller/checkpointsource.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+	return checkpointSourceFromManifest(manifest), nil
+}
+
+// checkpointSourceFromManifest projects a checkpoint manifest onto the status
+// block that publishes it. The compared values come from CompatEnvironment, the same
+// view the restore gates read, so what a user sees cannot drift from what a
+// restore is actually checked against.
+//
+// A block with nothing to say is left out entirely, and a manifest with nothing
```
danbar2, line 29: last 2 lines are redundent

api/v1alpha1/checkpoint_source_types.go
```diff
@@ -0,0 +1,97 @@
...
+	// +kubebuilder:pruning:PreserveUnknownFields
+	Devices *CheckpointSourceDevices `json:"devices,omitempty"`
+
+	// Node records the machine the checkpoint was captured on.
+	// +optional
+	Node *CheckpointSourceNode `json:"node,omitempty"`
+
+	// Pod records the container the checkpoint was captured from.
+	// +optional
+	Pod *CheckpointSourcePod `json:"pod,omitempty"`
```
danbar2, line 26: For future compatibility, I suggest to hold a slice for both Pods and Nodes

leeZardNav: do you mind expend on this point? i don't see when will we need multi pod or nodes, but maybe i'm missing something. we do have multi devices as they are part of what restore needs to have exactly the same for success restore

agent/internal/controller/checkpointsource.go
leeZardNav: done in 1c8140d0d8

api/v1alpha1/checkpoint_source_types.go
danbar2: My thought is that we aim to support multi node soon, but on the other hand we still dont know how the api will look and how we will store the artifact.
Maybe its too soon for now, you can ignore

danbar2 (dismissed):

api/v1alpha1/podsnapshotcontent_types.go
```diff
@@ -71,6 +71,11 @@ type PodSnapshotContentStatus struct {
 	// Standard types are Ready and Failed.
 	// +optional
 	Conditions []metav1.Condition `json:"conditions,omitempty"`
+
+	// Source publishes what the checkpoint was captured on, written when the
+	// capture goes Ready.
+	// +optional
+	Source *CheckpointSource `json:"source,omitempty"`
```
oleg-kushniriov, line 78: docs/reference/api.md still documents the content status as conditions only (Ready and Failed) and lists PodSnapshotContent as consumed by the operator only, and CHANGELOG.md has no [Unreleased] entry for the new status.source CRD field. Since this block exists specifically for a human reader, the API reference and the changelog are where they will look for it.

Suggestions

In docs/reference/api.md, extend the PodSnapshotContent section with a short status.source table (node.{name,architecture,kernelVersion}, pod.{image,imageDigest,memory,cpu}, devices.nvidia.{driverVersion,instances[].productName}), note that every field is optional and absent when unknown, and that it is informational only (the restore gates compare the manifest). Adjust the "Created by / read by" wording so the content is no longer described as operator-only.
Add an ### Added entry under [Unreleased] in CHANGELOG.md naming the new `PodSnapshotContent.status.source` field and that it is populated only for captures that go Ready after the upgrade.

oleg-kushniriov (dismissed):
one comment

api/v1alpha1/podsnapshotcontent_types.go
leeZardNav: done in 8208bb3187

oleg-kushniriov (approved):

PR #146, opened 2026-08-31
feat(chart): add runtime.storageDir for non-default containerd layouts

### LINKED ISSUE #143: Chart has no values for non-default containerd paths (RKE2/k3s)
## Environment

- Kubernetes: RKE2 (containerd socket at `/run/k3s/containerd`, snapshotter storage at `/var/lib/rancher/rke2/agent/containerd`)
- `runtimeClassName: nvidia`, single GPU node

## Summary

The agent DaemonSet's default host-path mounts assume the upstream containerd layout (`/run/containerd` for the socket, and a default runtime-storage path). RKE2 (and k3s) relocate both under `/var/lib/rancher/rke2/agent/containerd/...` / `/run/k3s/containerd`. The chart exposes no values for either path, so on RKE2:

1. With only the default mounts, the agent cannot resolve the container at all (`ContainerNotResolved`).
2. After patching just the containerd **socket** path, capture still fails to produce `rootfs-diff.tar` (see companion issue: "PodSnapshot reports Captured=True when rootfs-diff capture fails"), because the agent tars the overlay `upperDir` using the path recorded in containerd's own metadata (`manifest.yaml`'s `overlay.upperDir`), which is an RKE2 host path (`/var/lib/rancher/rke2/agent/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/<id>/fs`). That means the **storage/runtime-root mount's container-side mountPath must equal the host path**, not just be reachable — the agent reads the path out of containerd metadata and expects to find it at the identical path inside its own mount namespace.

## Steps to reproduce

1. `helm install` the chart's default values against an RKE2 (or k3s) cluster.
2. Observe the agent fails to resolve running containers on that node (`ContainerNotResolved`).
3. Patch only the containerd socket volume/mount to `/run/k3s/containerd` and retry — container resolution succeeds, but a subsequent capture cannot produce a correct rootfs diff because the runtime-storage mount is still at the upstream default path/mountPath rather than mirroring `/var/lib/rancher/rke2/agent/containerd`.

## Expected

Chart values for both the containerd socket directory and the containerd/snapshotter storage directory, with the storage volume mounted **inside the agent container at the same absolute path it has on the host** (since the path is read verbatim out of containerd metadata, not treated as a configurable prefix).

## Actual

No chart values exist for either path; only a direct DaemonSet volume/volumeMount patch works.

## Workaround (verified)

```
kubectl --context <your-context> -n <snapshot-namespace> patch daemonset snapshot-agent --type=json -p '[
  {"op":"replace","path":"/spec/template/spec/containers/0/volumeMounts/<idx-containerd-sock>/mountPath","value":"/run/k3s/containerd"},
  {"op":"replace","path":"/spec/template/spec/volumes/<idx-containerd-sock>/hostPath/path","value":"/run/k3s/containerd"},
  {"op":"replace","path":"/spec/template/spec/containers/0/volumeMounts/<idx-containerd-root>/mountPath","value":"/var/lib/rancher/rke2/agent/containerd"},
  {"op":"replace","path":"/spec/template/spec/volumes/<idx-containerd-root>/hostPath/path","value":"/var/lib/rancher/rke2/agent/containerd"}
]'
```
(Indices depend on the chart's volume ordering — the point is: both the socket path and the storage-root path need to be relocated together, and the storage-root mountPath must equal its hostPath, not just be present under some other in-container prefix.)

## Ask

- Add chart values, e.g. `agent.containerdSocketPath` and `agent.containerdRootPath` (or `runtime.socketDir` / `runtime.storageDir`), defaulting to the current upstream paths.
- Document/enforce (in the chart's template) that the storage-root value is mounted at the identical path inside the container, since it's a metadata-derived path lookup, not a relocatable prefix.
- Call out RKE2 (`/run/k3s/containerd`, `/var/lib/rancher/rke2/agent/containerd`) and k3s equivalents explicitly in the README's supported-environments section.

yuanjiewei:
`runtime.socketPath` is a chart value but the containerd snapshotter **storage** directory was not — `snapshot.runtimeStorageDir` hardcoded `/var/lib/containerd`.

RKE2 and k3s relocate that directory under `/var/lib/rancher/<distro>/agent/containerd`. On such a cluster the agent cannot resolve containers with chart defaults; after patching only the socket, resolution works but capture cannot produce a correct `rootfs-diff.tar` — the agent tars the overlay `upperDir` read verbatim out of containerd's metadata, so that host path must resolve **unchanged inside the agent's mount namespace**. The only workaround was a direct DaemonSet JSON patch of both the volume and the volumeMount (#143).

**Change:** chart-only.

- `runtime.storageDir`, empty by default, so current per-runtime defaults are unchanged.
- `snapshot.runtimeStorageDir` honours it, following the existing `snapshot.runtimeSocket` pattern.
- Rejects a non-absolute value at template time: it lands in both `hostPath.path` and the agent's `volumeMount.mountPath`, and Kubernetes rejects a relative `mountPath` at Pod admission — an error that names the mount rather than the value behind it.
- README section and a values-table row.

No template rewiring: `daemonset.yaml` already renders both the `hostPath.path` and the container `mountPath` from `$storageDir`. Deriving both from one value keeps them equal by construction, which is the point — the path is metadata-derived, not a relocatable prefix.

```yaml
# rendered with runtime.storageDir=/var/lib/rancher/rke2/agent/containerd
      mountPath: "/var/lib/rancher/rke2/agent/containerd"   # container
        path: "/var/lib/rancher/rke2/agent/containerd"      # host
```

**Tests:** `helm unittest charts/snapshot` → 10 passed, 4 suites. `helm lint` passes. Both new cases red-proofed — removing the override branch fails the override case; disabling the absolute-path guard fails exactly the rejection case (1 failed, 9 passed), each on its own assertion.

Observed on RKE2 + A10, driver 580.173.02, against v0.1.0-alpha.1.

Fixes #143

### DIFF
--- a/charts/snapshot/templates/_helpers.tpl
+++ b/charts/snapshot/templates/_helpers.tpl
@@ -99,7 +99,9 @@ Host directory holding per-container storage (overlay upperdirs the agent
 reads for rootfs-diff capture, and CRI-O config.json fallback).
 */}}
 {{- define "snapshot.runtimeStorageDir" -}}
-{{- if eq .Values.runtime.type "crio" -}}/var/lib/containers{{- else -}}/var/lib/containerd{{- end -}}
+{{- if .Values.runtime.storageDir -}}
+{{- .Values.runtime.storageDir -}}
+{{- else if eq .Values.runtime.type "crio" -}}/var/lib/containers{{- else -}}/var/lib/containerd{{- end -}}
 {{- end }}
 
 {{- define "snapshot.pageBrokerControlPath" -}}/pagebroker/control{{- end -}}
--- a/charts/snapshot/templates/_helpers.tpl
+++ b/charts/snapshot/templates/_helpers.tpl
@@ -78,6 +78,9 @@ Fail fast on unsupported runtime.type values. Called once from daemonset.yaml.
 {{- if not (has .Values.runtime.type (list "containerd" "crio")) }}
 {{- fail (printf "runtime.type must be 'containerd' or 'crio', got %q" .Values.runtime.type) }}
 {{- end }}
+{{- if and .Values.runtime.storageDir (not (hasPrefix "/" .Values.runtime.storageDir)) }}
+{{- fail (printf "runtime.storageDir must be an absolute path, got %q" .Values.runtime.storageDir) }}
+{{- end }}
 {{- end }}
 
 {{/*
--- a/charts/snapshot/templates/_helpers.tpl
+++ b/charts/snapshot/templates/_helpers.tpl
@@ -99,9 +99,13 @@ Host directory holding per-container storage (overlay upperdirs the agent
 reads for rootfs-diff capture, and CRI-O config.json fallback).
 */}}
 {{- define "snapshot.runtimeStorageDir" -}}
-{{- if .Values.runtime.storageDir -}}
-{{- .Values.runtime.storageDir -}}
-{{- else if eq .Values.runtime.type "crio" -}}/var/lib/containers{{- else -}}/var/lib/containerd{{- end -}}
+{{- if .Values.runtime.storageDir }}
+{{- .Values.runtime.storageDir }}
+{{- else if eq .Values.runtime.type "crio" }}
+{{- "/var/lib/containers" }}
+{{- else }}
+{{- "/var/lib/containerd" }}
+{{- end }}
 {{- end }}
 
 {{- define "snapshot.pageBrokerControlPath" -}}/pagebroker/control{{- end -}}
--- a/charts/snapshot/values.yaml
+++ b/charts/snapshot/values.yaml
@@ -58,14 +58,11 @@ crdUpgrade:
 runtime:
   type: containerd  # containerd | crio
   socketPath: ""
-  # Host directory holding per-container storage (the overlay upperdirs the
-  # agent tars for rootfs-diff capture, and the CRI-O config.json fallback).
-  # The agent reads these paths verbatim out of runtime metadata, so the
-  # volume is mounted at this same absolute path inside the agent container.
+  # Host directory holding per-container storage. Mounted into the agent at
+  # this same absolute path, because runtime metadata reports it verbatim.
   # Defaults: /var/lib/containerd (containerd), /var/lib/containers (crio).
-  # RKE2/k3s: /var/lib/rancher/rke2/agent/containerd
-  #           (k3s: /var/lib/rancher/k3s/agent/containerd), with
-  #           socketPath: /run/k3s/containerd/containerd.sock
+  # RKE2: /var/lib/rancher/rke2/agent/containerd, with
+  #       socketPath: /run/k3s/containerd/containerd.sock
   storageDir: ""
 
 # Gates the OCP RBAC template and the DS required-scc annotation. Keep false
--- a/charts/snapshot/values.yaml
+++ b/charts/snapshot/values.yaml
@@ -58,11 +58,9 @@ crdUpgrade:
 runtime:
   type: containerd  # containerd | crio
   socketPath: ""
-  # Host directory holding per-container storage. Mounted into the agent at
-  # this same absolute path, because runtime metadata reports it verbatim.
+  # Host per-container storage, mounted into the agent at the same absolute path.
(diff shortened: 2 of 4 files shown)


charts/snapshot/templates/_helpers.tpl
```diff
@@ -99,5 +99,7 @@ Host directory holding per-container storage (overlay upperdirs the agent
 reads for rootfs-diff capture, and CRI-O config.json fallback).
 */}}
 {{- define "snapshot.runtimeStorageDir" -}}
-{{- if eq .Values.runtime.type "crio" -}}/var/lib/containers{{- else -}}/var/lib/containerd{{- end -}}
+{{- if .Values.runtime.storageDir -}}
```
dfeigin-nv: Do you need the trailing dashes? have the style align with `snapshot.runtimeSocket`

dfeigin-nv (approved):
One styling comment. Otherwise LGTM

charts/snapshot/values.yaml
```diff
@@ -36,11 +36,20 @@ crdUpgrade:
...
   type: containerd  # containerd | crio
   socketPath: ""
+  # Host directory holding per-container storage (the overlay upperdirs the
+  # agent tars for rootfs-diff capture, and the CRI-O config.json fallback).
+  # The agent reads these paths verbatim out of runtime metadata, so the
+  # volume is mounted at this same absolute path inside the agent container.
+  # Defaults: /var/lib/containerd (containerd), /var/lib/containers (crio).
+  # RKE2/k3s: /var/lib/rancher/rke2/agent/containerd
+  #           (k3s: /var/lib/rancher/k3s/agent/containerd), with
+  #           socketPath: /run/k3s/containerd/containerd.sock
```
Ronkahn21: can you concise the documentation

charts/snapshot/templates/_helpers.tpl
yuanjiewei: Good catch — dropped the dashes and split it across lines so it reads exactly like `snapshot.runtimeSocket` right above. e55d93e.

charts/snapshot/values.yaml
yuanjiewei: Trimmed to the two things an operator actually acts on: the mount path equals the host path, and what to set on RKE2. Dropped the rest. e55d93e.

yuanjiewei: Cut it further in 852470f — three lines now: mount path equals host path, the two defaults, RKE2 on one line.

yuanjiewei:
@dfeigin-nv sorry for the churn — the last push was only to add the missing `Signed-off-by` trailers (DCO was red on my two follow-up commits). No content change: still 4 files, +75/-3, same as when you approved. The force-push dismissed your approval, so it needs another look when you get a chance.

Both review points are addressed: the helper now reads like `snapshot.runtimeSocket`, and the `storageDir` comment is down to three lines.

charts/snapshot/values.yaml
yuanjiewei: CI is green now if you want to take another look — the comment is three lines as of 852470f.

charts/snapshot/templates/_helpers.tpl
```diff
@@ -99,7 +99,13 @@ Host directory holding per-container storage (overlay upperdirs the agent
 reads for rootfs-diff capture, and CRI-O config.json fallback).
 */}}
 {{- define "snapshot.runtimeStorageDir" -}}
-{{- if eq .Values.runtime.type "crio" -}}/var/lib/containers{{- else -}}/var/lib/containerd{{- end -}}
+{{- if .Values.runtime.storageDir }}
+{{- .Values.runtime.storageDir }}
```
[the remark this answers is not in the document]
yuanjiewei: Good catch — fixed in 072ce3a. `snapshot.validateRuntime` now rejects a non-absolute `runtime.storageDir`, with a `failedTemplate` case in `runtime_storage_test.yaml`. Verified it can fail: disabling the guard fails exactly that case (1 failed, 9 passed).

Ronkahn21 (approved):

Ronkahn21:
@yuanjiewei Thanks

PR #147, opened 2026-08-31
fix(agent): fail the checkpoint when overlay rootfs-diff capture fails

### LINKED ISSUE #142: PodSnapshot reports Captured=True when rootfs-diff capture fails
## Environment

- Kubernetes: RKE2 (containerd at `/run/k3s/containerd`, storage at `/var/lib/rancher/rke2/agent/containerd`)
- `runtimeClassName: nvidia`, single GPU node

## Summary

The agent can emit `ERROR` log lines during the overlay/rootfs-diff step of a capture, and the `PodSnapshot`/`PodSnapshotContent` pair still reaches `Ready=True reason=Captured status=True message="Checkpoint captured and verified"` with no `Failed` condition and no warning event. The resulting checkpoint directory on the PVC is missing `rootfs-diff.tar`, which the project's own e2e test asserts must exist for a successful capture.

## Steps to reproduce

1. Install the chart (operator + agent DaemonSet) against an RKE2 node, capture a GPU workload pod via `PodSnapshot`.
2. Watch the agent pod logs during capture. Errors are logged from the overlay-diff path — around `checkpoint.go:287` (tar of overlay `upperDir` failed) and `checkpoint.go:290` (whiteout `lstat` failed) — while the capture continues.
3. After the `PodSnapshot` reaches `Ready=True/Captured`, list the checkpoint's artifact directory on the storage PVC (`/checkpoints/<checkpoint-id>/versions/<n>/`).

## Expected

Either the CR should not report `Ready=True/Captured` when the rootfs-diff step fails, or a distinct condition/event should surface the partial failure so a caller can tell the artifact is incomplete before attempting restore.

## Actual

- `PodSnapshot` status:
  ```
  conditions:
  - lastTransitionTime: "2026-08-31T05:55:41Z"
    message: "Checkpoint captured and verified"
    reason: Captured
    status: "True"
    type: Ready
  - reason: Captured
    status: "False"
    type: Failed
  ```
- `PodSnapshotContent` mirrors the same `Ready=True/Captured`.
- The artifact directory contains a full CRIU image set (`inventory.img`, `pages-N.img`, `pstree.img`, `mm-1.img`, `fs-1.img`, `files.img`, `criu.conf`, `dump.log`, `manifest.yaml`, `pagemap-shmem-*.img`, `tmpfs-dev-*.tar.gz.img`) but **no `rootfs-diff.tar`**.
- `manifest.yaml`'s `overlay:` block correctly records the container's real overlay `upperDir` path, so the agent did resolve the rootfs-diff inputs — it just did not produce the artifact, and nothing in status/events reflects that.
- `dump.log` itself is clean (`"Dumping finished successfully"`, no error/fail lines) — the failure is isolated to the overlay/rootfs-diff step, not the CRIU dump.
- The project's own e2e test `test_successful_snapshot_captures_cpu_gpu_and_fs` asserts `"./rootfs-diff.tar" in artifact_listing` for a successful capture — this run's artifact would fail that assertion despite the CR reporting success.
- Retrieving the agent's own application logs for the exact capture window after the DaemonSet pod rolled was not possible (logs are not retained across pod restarts) — so the ERROR lines are best captured from a live tail during a reproduction rather than after the fact.

## Ask

- Fail the `PodSnapshot`/`PodSnapshotContent` (or add a distinct condition, e.g. `RootfsDiffFailed`) when the overlay tar/whiteout step errors, instead of folding it into an overall `Ready=True/Captured`.
- At minimum, surface a warning `Event` on the CR when the rootfs-diff artifact write fails, so the omission isn't silent. A caller currently has no signal, from the CR alone, that the artifact is incomplete — the only way to know is to list the PVC directly.

yuanjiewei:
The overlay rootfs-diff step logged its failures and dropped them, so a capture that failed to produce `rootfs-diff.tar` still promoted its staged directory and landed `Ready=True reason=Captured`. Nothing in the CR said otherwise; the only way to find out was to list the PVC.

That artifact is not a degraded checkpoint — it restores with every container-created file silently absent, and the caller cannot tell it from a complete one. The project already treats the file as required: `test_successful_snapshot_captures_cpu_gpu_and_fs` asserts it is in the listing, so a run could report `Captured` and fail that assertion.

**Change:** both errors propagate out of `captureCheckpoint`. The staged directory is never promoted, so the existing `CheckpointFailed` path fires. The step is extracted into `captureOverlay` only so a unit test can reach it without a CRIU dump.

The old comment justified the best-effort handling with "the base container image provides the filesystem". That holds only for a container that wrote nothing, which is not knowable at that point.

**Tests:** `go test ./internal/executor/` on linux (go 1.26) → `ok 0.136s`. Adds `TestCaptureOverlayFailsCheckpointOnRootfsDiffError`.

Observed on RKE2 + A10, driver 580.173.02, against v0.1.0-alpha.1. Evidence in #142.

Fixes #142

### DIFF
--- a/agent/internal/executor/checkpoint.go
+++ b/agent/internal/executor/checkpoint.go
@@ -326,19 +326,32 @@ func captureCheckpoint(ctx context.Context, criuOpts *criurpc.CriuOpts, criuSett
 	}
 	timings.CRIUDumpDuration = criuDumpDuration
 
-	// Overlay rootfs diff capture is best-effort. Failures are logged but not
-	// propagated — a checkpoint without overlay diffs is still valid for restore
-	// (the base container image provides the filesystem).
-	if state.UpperDir != "" {
-		overlayCaptureStart := time.Now()
-		if _, err := snapshotruntime.CaptureRootfsDiff(state.UpperDir, checkpointDir, data.Overlay.Exclusions, data.Overlay.BindMountDests); err != nil {
-			log.Error(err, "Failed to capture rootfs diff")
-		}
-		if _, err := snapshotruntime.CaptureDeletedFiles(state.UpperDir, checkpointDir); err != nil {
-			log.Error(err, "Failed to capture deleted files")
-		}
-		timings.OverlayCaptureDuration = time.Since(overlayCaptureStart)
+	overlayCaptureDuration, err := captureOverlay(state.UpperDir, checkpointDir, data)
+	if err != nil {
+		return nil, err
 	}
+	timings.OverlayCaptureDuration = overlayCaptureDuration
 
 	return timings, nil
+}
+
+// captureOverlay writes the container's rootfs diff and whiteout list into the staged
+// checkpoint directory.
+//
+// Failures are returned, not logged and dropped: an artifact missing rootfs-diff.tar
+// restores with every container-created file silently absent, and the caller has no way
+// to tell that from a complete checkpoint. Failing here keeps the staged directory from
+// being promoted, so the work order lands CheckpointFailed instead of Captured.
+func captureOverlay(upperDir, checkpointDir string, data *types.CheckpointManifest) (time.Duration, error) {
+	if upperDir == "" {
+		return 0, nil
+	}
+	start := time.Now()
+	if _, err := snapshotruntime.CaptureRootfsDiff(upperDir, checkpointDir, data.Overlay.Exclusions, data.Overlay.BindMountDests); err != nil {
+		return 0, fmt.Errorf("failed to capture rootfs diff: %w", err)
+	}
+	if _, err := snapshotruntime.CaptureDeletedFiles(upperDir, checkpointDir); err != nil {
+		return 0, fmt.Errorf("failed to capture deleted files: %w", err)
+	}
+	return time.Since(start), nil
 }
--- a/agent/internal/executor/checkpoint_test.go
+++ b/agent/internal/executor/checkpoint_test.go
@@ -69,3 +69,18 @@ func TestCheckpointNeedsSourceKill(t *testing.T) {
 	assert.False(t, CheckpointNeedsSourceKill(errors.New("prepare failed")))
 	assert.False(t, CheckpointNeedsSourceKill(fmt.Errorf("commit PageBroker checkpoint: %w", errors.New("failed"))))
 }
+
+// A rootfs-diff capture failure must fail the checkpoint. Swallowing it publishes an
+// artifact whose restore silently drops every container-created file.
+func TestCaptureOverlayFailsCheckpointOnRootfsDiffError(t *testing.T) {
+	manifest := &types.CheckpointManifest{}
+
+	// Absent upperdir: nothing to capture, no failure.
+	duration, err := captureOverlay("", t.TempDir(), manifest)
+	require.NoError(t, err)
+	assert.Zero(t, duration)
+
+	// Unreadable upperdir: tar fails and the error must propagate.
+	_, err = captureOverlay(filepath.Join(t.TempDir(), "missing-upperdir"), t.TempDir(), manifest)
+	require.ErrorContains(t, err, "failed to capture rootfs diff")
+}


yuanjiewei:
Rebased onto main. The only conflict was in the test file — main added `TestCheckpointPageBrokerPrepareFailureDoesNotMutate` and `TestCheckpointNeedsSourceKill` right where this PR adds `TestCaptureOverlayFailsCheckpointOnRootfsDiffError`; all three are independent, so I kept them all. `captureOverlay` merged cleanly. Tests pass on linux (go 1.26): `ok github.com/ai-dynamo/snapshot/agent/internal/executor 0.132s`.

yuanjiewei:
Rebased and green. This one has been open since Aug 31 without a review — could someone take a look?

Short version: `CaptureRootfsDiff` failures were logged and dropped, so a checkpoint missing `rootfs-diff.tar` got published as Captured. It restores with every container-created file silently absent, and nothing distinguishes it from a complete artifact. This returns the error so the staged directory is never promoted.

oleg-kushniriov (commented):
/

agent/internal/executor/checkpoint.go
```diff
@@ -326,19 +326,32 @@ func captureCheckpoint(ctx context.Context, criuOpts *criurpc.CriuOpts, criuSett
...
-	if state.UpperDir != "" {
-		overlayCaptureStart := time.Now()
-		if _, err := snapshotruntime.CaptureRootfsDiff(state.UpperDir, checkpointDir, data.Overlay.Exclusions, data.Overlay.BindMountDests); err != nil {
-			log.Error(err, "Failed to capture rootfs diff")
-		}
-		if _, err := snapshotruntime.CaptureDeletedFiles(state.UpperDir, checkpointDir); err != nil {
-			log.Error(err, "Failed to capture deleted files")
-		}
-		timings.OverlayCaptureDuration = time.Since(overlayCaptureStart)
+	overlayCaptureDuration, err := captureOverlay(state.UpperDir, checkpointDir, data)
```
oleg-kushniriov, line 329: captureOverlay runs after criu.ExecuteDump with LeaveRunning=false, so an unreadable `state.UpperDir` now kills the workload instead of producing a degraded artifact. The most likely cause (unmounted runtime.storageDir) is detectable in inspectContainer before anything destructive happens.

agent/internal/executor/checkpoint.go
```diff
@@ -326,19 +326,32 @@ func captureCheckpoint(ctx context.Context, criuOpts *criurpc.CriuOpts, criuSett
...
+// being promoted, so the work order lands CheckpointFailed instead of Captured.
+func captureOverlay(upperDir, checkpointDir string, data *types.CheckpointManifest) (time.Duration, error) {
+	if upperDir == "" {
+		return 0, nil
+	}
+	start := time.Now()
+	if _, err := snapshotruntime.CaptureRootfsDiff(upperDir, checkpointDir, data.Overlay.Exclusions, data.Overlay.BindMountDests); err != nil {
+		return 0, fmt.Errorf("failed to capture rootfs diff: %w", err)
+	}
+	if _, err := snapshotruntime.CaptureDeletedFiles(upperDir, checkpointDir); err != nil {
```
oleg-kushniriov, line 353: With CaptureDeletedFiles errors now fatal, a single transient lstat ENOENT inside findWhiteoutFiles fails the checkpoint after the source process is already dead. Skipping os.IsNotExist entries in the walk keeps the step strict without making it flaky.

Suggestion: In findWhiteoutFiles, treat os.IsNotExist(err) in the walk callback as skip-and-continue (return nil for files, filepath.SkipDir for directories) and still propagate every other error.

agent/internal/executor/checkpoint.go
```diff
@@ -326,19 +326,32 @@ func captureCheckpoint(ctx context.Context, criuOpts *criurpc.CriuOpts, criuSett
...
-		if _, err := snapshotruntime.CaptureRootfsDiff(state.UpperDir, checkpointDir, data.Overlay.Exclusions, data.Overlay.BindMountDests); err != nil {
-			log.Error(err, "Failed to capture rootfs diff")
-		}
-		if _, err := snapshotruntime.CaptureDeletedFiles(state.UpperDir, checkpointDir); err != nil {
-			log.Error(err, "Failed to capture deleted files")
-		}
-		timings.OverlayCaptureDuration = time.Since(overlayCaptureStart)
+	overlayCaptureDuration, err := captureOverlay(state.UpperDir, checkpointDir, data)
+	if err != nil {
+		return nil, err
```
leeZardNav, line 331: similar to Oleg's comment, could it be that old code did produced a checkpoint that was able to be restored? do we have data to check it? if not can we collect or try some scenarios?
asking as if we'll take the strict approach we might (or not) lose on mostly passable checkpoint and without data we don't know if we correctly take strict approach or not

the comment in the code said it is still valid for restore. as the bug suggested it isn't but to what extent?

agent/internal/executor/checkpoint.go
leeZardNav, line 311: if we do want to take this path, should we also work on the restore part? i think there's also a part that ignore the errors in ApplyDeletedFiles on restore side

PR #149, opened 2026-08-31
docs: mntnsCompatMode is a dump-time setting, and the RKE2 mount-v2 issue

### LINKED ISSUE #144: CRIU restore fails on mount-v2 sharing copy for nvidia-runtime bind mounts; mntnsCompatMode workaround is dump-time only and undocumented as such
## Environment

- Kubernetes: RKE2 (containerd at `/run/k3s/containerd` / `/var/lib/rancher/rke2/agent/containerd`)
- `runtimeClassName: nvidia` on both source and restore pods, single GPU node

## Summary

Restore of a checkpoint captured with `runtimeClassName: nvidia` fails consistently (3/3 reproductions) in CRIU's mount-v2 engine while restoring the nvidia device bind mounts (`/dev/nvidia0` etc.), with `mnt-v2: Failed to copy sharing ... Invalid argument`. Setting `config.criu.mntnsCompatMode=true` avoids the error, but only for checkpoints captured **after** the flip — restoring an existing (pre-flip) checkpoint with the flag set still fails identically, because the restore path reads CRIU dump-time settings back out of the checkpoint's own `manifest.yaml` rather than honoring the current chart/agent config at restore time.

## Steps to reproduce

1. Capture a GPU pod (`runtimeClassName: nvidia`, `nvidia.com/gpu: 1`) via `PodSnapshot`, default chart values (`config.criu.mntnsCompatMode` unset/false).
2. Restore that checkpoint to a fresh pod (`runtimeClassName: nvidia`, same GPU limit, restore labels/annotations per the chart's protocol).
3. Restore fails; `restore.log` shows CRIU's mount-v2 engine failing to replicate the propagation/sharing group for the nvidia device mounts.

## Expected

Restore succeeds for a checkpoint captured with `runtimeClassName: nvidia` bind mounts on RKE2, matching the project's own e2e (`test_successful_restore_recovers_cpu_gpu_and_fs_from_snapshot`, which also runs `runtimeClassName: nvidia` on both source and restore pods).

## Actual

```
mnt-v2: Failed to copy sharing from -1:/dev/nvidia0 to <id>: Invalid argument
...
Restoring FAILED
```

(Reproduced 3/3, with `runtimeClassName: nvidia` on both source and vessel.)

## Analysis (source-verified, file:line against this repo)

- The external-mount handling is correct and not the bug: `BuildMountPolicy` (`agent/internal/runtime/mounts.go:80-102`) externalizes every non-OCI-managed, non-`/dev/shm` mount, which covers `/dev/nvidia0`/`nvidiactl`/`nvidia-uvm`/`nvidia-uvm-tools` automatically. `BuildDumpOptions` (`agent/internal/criu/dump.go:42,55`) turns that into `criuOpts.ExtMnt`, which is persisted into `manifest.yaml`'s `criuDump.extMnt` (`agent/internal/types/manifest.go:60-71`), and `buildRestoreExtMounts` (`agent/internal/criu/restore.go:171-184`) replays it verbatim at restore. So the nvidia mounts genuinely are declared external on both sides.
- `mnt-v2: Failed to copy sharing` is CRIU's own mount-namespace-restore engine failing to reconcile the mount's *propagation/sharing group* recorded at dump time against the restore mount namespace — a separate mechanism from whether the mount is in `ExtMnt`. The externalization predicate never inspects propagation mode (`agent/internal/types/inspect.go:11-19`'s `MountInfo` has no peer/master group field), so this is not something the current predicate can fix; it's upstream CRIU behavior interacting with the host's mount-propagation setup.
- `mntnsCompatMode` is the one restore-time knob aimed at this class of problem: chart value `config.criu.mntnsCompatMode` (`charts/snapshot/values.yaml:232`, default `false`) → `configmap.yaml:49` → `agent/internal/types/config.go:134` → applied restore-only in `BuildRestoreOpts` (`agent/internal/criu/restore.go:141`). It is **not** mirrored into `BuildDumpOptions` at all — it's a pure restore-side CRIU RPC field.
- The failure mode when restoring a pre-flip checkpoint with the flag now set: `BuildRestoreOpts` builds its CRIU options from the checkpoint's own persisted `manifest.yaml` (`CRIUDump.CRIU`, read at `agent/internal/criu/restore.go:141`), not from the agent's live config at restore time for this field — so flipping `mntnsCompatMode` in the chart after a checkpoint already exists does not change how that checkpoint restores; only checkpoints captured after the flip pick it up, if the option is (or should be) captured into the manifest at dump time. This makes the workaround easy to apply too late and get a false negative ("mntnsCompatMode didn't fix it") when the real issue is that the flag needs to be set before the *capture*, not just before the restore.

## Workaround (verified)

Set `config.criu.mntnsCompatMode=true` (`--set config.criu.mntnsCompatMode=true` on `helm install`/`upgrade`) **before** capturing the checkpoint you intend to restore. Restoring a checkpoint captured before the flag was set fails the same way even with the flag now on.

## Ask

1. Investigate the `mnt-v2: Failed to copy sharing` failure for nvidia-runtime bind mounts specifically on RKE2 — likely a CRIU-version/kernel/mount-propagation mismatch between the CI environment (vCluster-hosted e2e) and RKE2's mount setup, but worth confirming upstream against known CRIU mount-v2 sharing-group issues. (Note: the agent's `Dockerfile` pins `CRIU_REF=criu-dev`, a moving branch, so the exact CRIU build baked into the `v0.1.0-alpha.1` image tag isn't a fixed CRIU release.)
2. Document `mntnsCompatMode` as effective at **dump time**, not restore time, in `values.yaml`'s comment and the README — the current comment ("Mount namespace compatibility mode") doesn't convey that it must be set before capture to help a given checkpoint, or that toggling it after capture cannot retroactively fix an existing artifact.
3. Consider either a restore-time override for this option (independent of what's baked into the checkpoint's `manifest.yaml`), or explicit validation/warning when a restore is attempted with a `mntnsCompatMode` mismatch between the running agent config and the checkpoint's own dump-time record.

yuanjiewei:
`config.criu.mntnsCompatMode` maps to a CRIU **restore** RPC field, so it reads as a restore-time knob. It is not one.

`BuildRestoreOpts` builds its options from the checkpoint's persisted manifest (`settings := m.CRIUDump.CRIU`), not from the agent's live config. The value in force **at capture** is what a checkpoint restores with. Flipping the chart value does not change an existing checkpoint; only later captures pick it up.

That makes the workaround easy to apply too late and read as a false negative — a restore that still fails after enabling the flag looks like the flag did not help, when it simply was not set before the capture. Verified 3/3 in #144: a pre-flip checkpoint fails identically with the flag on; re-capturing with it on succeeds.

**Change:** docs only.

- `values.yaml`: the comment now says it is applied at dump time and a re-capture is required.
- `README.md`: a Known Issues section with that entry, plus the CRIU mount-v2 sharing-copy failure for nvidia bind mounts on RKE2 and its `mntnsCompatMode=true` + re-capture workaround.

The mount-v2 entry is explicit that external-mount handling is *not* the bug — the device mounts are correctly declared external on both sides. It is CRIU failing to reconcile the propagation group recorded at dump time.

Scope: ask (2) of #144. Ask (1), the upstream CRIU/kernel investigation, and ask (3), a restore-time override, are out of scope.

**Tests:** `helm lint` and `helm unittest charts/snapshot` pass; `helm template` output unchanged apart from the comment.

Fixes #144

## Summary by CodeRabbit

- **Documentation**
  - Added guidance explaining that `mntnsCompatMode` must be enabled before capturing a checkpoint and that its setting is retained in checkpoint metadata for restoration.
  - Clarified that changing the configuration does not affect existing checkpoints until they are captured again; the default remains disabled.
  - Documented an RKE2/NVIDIA runtime restore issue involving mount handling and provided a Helm-based recapture workaround.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,51 @@ Current limitations:
 
 Multi-GPU and Arm support are on the roadmap.
 
+## Known Issues
+
+### `mntnsCompatMode` is applied at dump time, not restore time
+
+`config.criu.mntnsCompatMode` reads as a restore-side option — it maps to a CRIU
+restore RPC field — but it takes effect **at capture**. The value in force when a
+checkpoint is dumped is persisted into that checkpoint's `manifest.yaml`, and
+`BuildRestoreOpts` reads the setting back out of the manifest
+(`m.CRIUDump.CRIU`) rather than from the agent's live configuration.
+
+Consequences:
+
+- Flipping `config.criu.mntnsCompatMode` does **not** change how an already
+  captured checkpoint restores. Only checkpoints captured after the change pick
+  it up.
+- A restore that still fails after enabling the flag is not evidence the flag
+  does not help — re-capture first, then restore.
+
+### CRIU mount-v2 sharing-copy failure with nvidia-runtime bind mounts on RKE2
+
+Restoring a checkpoint captured with `runtimeClassName: nvidia` can fail in
+CRIU's mount-v2 engine while restoring the NVIDIA device bind mounts:
+
+```text
+mnt-v2: Failed to copy sharing from -1:/dev/nvidia0 to <id>: Invalid argument
+...
+Restoring FAILED
+```
+
+The device mounts are correctly declared external on both sides; this is CRIU's
+mount-namespace-restore engine failing to reconcile the mount's
+propagation/sharing group recorded at dump time against the restore mount
+namespace — a separate mechanism from external-mount handling, and dependent on
+the host's mount-propagation setup.
+
+**Workaround:** set `config.criu.mntnsCompatMode=true` **before** capturing the
+checkpoint you intend to restore, then re-capture. Per the note above, enabling
+it after the capture cannot fix an existing artifact.
+
+```bash
+helm upgrade --install snapshot ./charts/snapshot \
+  --namespace "${NAMESPACE}" \
+  --set config.criu.mntnsCompatMode=true
+```
+
 ## Documentation
 
 **Get started**
--- a/charts/snapshot/values.yaml
+++ b/charts/snapshot/values.yaml
@@ -242,9 +242,17 @@ config:
     manageCgroupsMode: soft # CRIU cgroup management mode (ignore/soft/full/strict)
     imageIoMode: direct     # CRIU image I/O mode (writeback/direct); direct uses O_DIRECT to avoid page-cache double-buffering
 
-    # Restore-specific options (only apply during CRIU restore, not dump)
+    # Restore-side options. These shape the restore, not the dump; note that
+    # mntnsCompatMode below is nonetheless read at capture time and persisted.
     rstSibling: true         # Restore as sibling process (required for go-criu swrk mode)
-    mntnsCompatMode: false   # Mount namespace compatibility mode
+    # Mount namespace compatibility mode. This is a restore-side CRIU option,
+    # but it is APPLIED AT DUMP TIME: the value in effect when a checkpoint is
+    # captured is persisted into that checkpoint's manifest.yaml, and
+    # BuildRestoreOpts reads it back from there (m.CRIUDump.CRIU) rather than
+    # from the agent's live config. Flipping this value therefore does NOT
+    # change how an existing checkpoint restores — set it before the capture
+    # and re-capture to take effect. See Known Issues in the top-level README.
+    mntnsCompatMode: false
     evasiveDevices: true     # Use any device path if original is inaccessible
     forceIrmap: true         # Force resolving inotify/fsnotify watch names
 


charts/snapshot/values.yaml
```diff
@@ -205,7 +205,14 @@ config:
 
     # Restore-specific options (only apply during CRIU restore, not dump)
     rstSibling: true         # Restore as sibling process (required for go-criu swrk mode)
-    mntnsCompatMode: false   # Mount namespace compatibility mode
+    # Mount namespace compatibility mode. This is a restore-side CRIU option,
+    # but it is APPLIED AT DUMP TIME: the value in effect when a checkpoint is
```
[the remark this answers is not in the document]
yuanjiewei: Done in cff4816. Reworded the section heading rather than moving the option, since `mntnsCompatMode` genuinely is a restore-side CRIU option — it is only the *value* that is read at capture time and persisted. The heading now says so instead of claiming the whole section is restore-only.

yuanjiewei:
Rebased onto main — the conflict was `## Status` being replaced by `## Limitations` upstream. Kept your section and put Known Issues after it. Chart still renders clean.

yuanjiewei:
Rebased and green — ready for review. Docs only.

The finding behind it: `mntnsCompatMode` reads like a restore option but is persisted into the manifest at dump time and read back from there, so flipping it does not change how an existing checkpoint restores. We lost time to exactly that before tracing it to `BuildRestoreOpts`.

README.md
```diff
@@ -154,6 +154,51 @@ Current limitations:
...
+propagation/sharing group recorded at dump time against the restore mount
+namespace — a separate mechanism from external-mount handling, and dependent on
+the host's mount-propagation setup.
+
+**Workaround:** set `config.criu.mntnsCompatMode=true` **before** capturing the
+checkpoint you intend to restore, then re-capture. Per the note above, enabling
+it after the capture cannot fix an existing artifact.
+
+```bash
+helm upgrade --install snapshot ./charts/snapshot \
```
shayasoolin, line 197: The shown upgrade may reset existing custom values. Tell users to update their existing values file, or preserve existing Helm values.

README.md
```diff
@@ -154,6 +154,51 @@ Current limitations:
 
 Multi-GPU and Arm support are on the roadmap.
 
+## Known Issues
+
+### `mntnsCompatMode` is applied at dump time, not restore time
```
shayasoolin, line 159:
```suggestion
### `mntnsCompatMode` is persisted at capture time and applied at restore time
```

shayasoolin (commented):
Thanks for the PR, looks good, please see a couple of comments.
Can you please change the description to use "Addresses #144" rather than "Fixes #144"?

README.md
```diff
@@ -154,6 +154,51 @@ Current limitations:
 
 Multi-GPU and Arm support are on the roadmap.
 
+## Known Issues
```
shayasoolin: I wonder if we really want two separate known issues here, or just one.
The actual issue is: "CRIU mount-v2 restore can fail for NVIDIA bind mounts on RKE2"
`mntnsCompatMode` is the workaround. Its capture-time persistence is an important caveat of that workaround, but not a separate failure.
WDYT?

README.md
```diff
@@ -154,6 +154,51 @@ Current limitations:
...
+
+Consequences:
+
+- Flipping `config.criu.mntnsCompatMode` does **not** change how an already
+  captured checkpoint restores. Only checkpoints captured after the change pick
+  it up.
+- A restore that still fails after enabling the flag is not evidence the flag
+  does not help — re-capture first, then restore.
+
+### CRIU mount-v2 sharing-copy failure with nvidia-runtime bind mounts on RKE2
```
shayasoolin: I believe the issue is broader than just for RKE2. I experienced it in a different flavor. Shall we use a more generic title such as "CRIU mount-v2 sharing failure with NVIDIA bind mounts" and avoid mentioning RKE2 specifically?

PR #155, opened 2026-09-01
feat(agent): coordinate CUDA interposer lifecycle

galletas1712:
## Summary
- implement the native `cuinterposer-coordinator` protocol and topology validation
- detect live shim participants through Unix sockets and persisted restore state through `cuinterposer.state`
- invoke prepare before native CUDA checkpointing
- pre-open and invoke the coordinator across CRIU mount-namespace replacement during restore
- keep orchestration dormant when the forwarding-only shim from #78 has no control sockets or state file

This is layer 3 of [stack #156](https://github.com/ai-dynamo/snapshot/stack/156). It adds coordinator and agent orchestration, but not POSIX allocation tracking or checkpoint behavior.

## Validation
- all Go tests
- production CUDA-builder compile
- static coordinator has no dynamic dependencies
- intermediate branch builds independently

### DIFF
--- a/agent/cmd/cuinterpose/coordinator.c
+++ b/agent/cmd/cuinterpose/coordinator.c
@@ -4,8 +4,670 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-int
-main(void)
+#define _GNU_SOURCE
+
+#include <errno.h>
+#include <limits.h>
+#include <stdbool.h>
+#include <stdint.h>
+#include <stdio.h>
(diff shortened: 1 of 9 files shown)


galletas1712:
Superseded by #216 (coordinator); the agent orchestration moved to #214. The re-cut stack (#212 → #220) supersedes this PR. Its body is quoted verbatim in the replacement's Origin section, and each review thread here has a row in the replacement's "Review threads carried" table with what was done about it.

PR #119, opened 2026-08-26
refactor(cuda): decompose CustomStorage helper

hhzhang16:
## Summary

Follow-up to #109 and #96 that decomposes the qualified CUDA CustomStorage helper; does not change its protocol or behavior. It does not need to merge in the initial CustomStorage code-freeze window.

## Source ownership

```mermaid
flowchart LR
    CLI["main.cpp<br/>CLI dispatch"] --> Server["daemon_server.*<br/>socket, health, shutdown"]
    Server --> CUDA["cuda_operation.*<br/>driver operation + context lifetime"]
    CUDA --> Scheduler["transfer_scheduler.*<br/>workers, deadline, cancellation"]
    Scheduler --> Adapter["transfer_engine.*<br/>Snapshot-local NIXL POSIX adapter"]
```

## Main changes

- `main.cpp`: reduced to CLI parsing and command dispatch
- `daemon_server`: Unix-socket serving, health requests, shutdown, and RPC response handling
- `cuda_operation`: CUDA initialization, operation completion, and target-scoped primary context lifetime
- `transfer_scheduler`: per-extent worker creation, shared deadline, sibling cancellation, and
  result aggregation

- Preserves the existing cancellation test and adds scheduler empty-batch, invalid-result, and backend-failure coverage.
- standalone and production Docker targets to build and test
- No change to CUDA operation ordering, timeout behavior, unknown-outcome handling, target identity, or retained-context lifetime.
- Exact-image legacy and POSIX Qwen3-0.6B checkpoint/restore/inference smoke pass

PR #164, opened 2026-09-01
perf(cuda): restore CustomStorage ranks concurrently

hhzhang16:
## Summary

Restore independent CUDA CustomStorage process extents concurrently within one Snapshot target container. It also removes the TP1-only qualification guard by recording and validating the GPU subset used by each CUDA process.

## Review focus

The highest-value review is the two-phase batch lifecycle in `cuda_operation.cpp`: validate and prepare every CUDA restore handle, transfer every independent extent concurrently, complete all handles, then retain contexts by exact target identity.

## Main changes

- Add a bounded `restore-batch` helper request carrying validated per-process restore targets.
- Discover the GPU UUID subset reported for each CUDA process after workload quiescence and record that selection for checkpoint. VMM-only owners that are absent from NVML's compute-process table conservatively use the complete validated target allocation; the helper still rejects any driver-returned stream outside that set.
- Validate each restore process manifest against the destination GPU allocation.
- Prepare each process restore sequentially, move independent extents concurrently through the existing NIXL/POSIX transfer adapter, then complete CUDA operations in reverse captured process-tree order so descendants resume before parents.
- Keep single-process POSIX and all legacy operations on their existing paths.
- Preserve one failure domain: an extent failure cooperatively cancels sibling workers before completion and no process is unlocked. A later CUDA completion or context-adoption failure is not rollback-safe; it is fatal to the helper and the partially restored workload must be discarded rather than replayed.
- Cap a request at 64 target processes, 64 concurrent extent workers, and 2 GiB aggregate pinned transfer memory.
- Raise the helper's default memory request/limit to 1/4 GiB so the qualified TP4 configuration has room for its 1 GiB pinned-buffer set plus CUDA/NIXL overhead.
- Emit batch wall time plus context, CUDA API, transfer, completion, adoption, byte-count, and throughput telemetry.

## Results/Validation

The integrated TP4/EP4 diagnostic used DeepSeek-V4-Flash-NVFP4, 1M maximum context, 64 sequences, GMS, checkpoint communications, four physical B200s, and POSIX CustomStorage for both GMS and engine artifacts.

| Exact-final restore phase | Run 1 | Run 2 |
| --- | ---: | ---: |
| GMS helper operation | 78.824 s | 81.289 s |
| GMS CUDA restore | 78.841 s | 81.302 s |
| Engine helper operation | 119.796 s | 120.181 s |
| Engine CUDA restore | 119.849 s | 120.227 s |
| Sum of target external restores | 204.661 s | 206.124 s |
| GMS-start to engine-complete | 212.043 s | 218.079 s |
| Pod creation to Ready | 219 s | 226 s |

The GMS batch moved 195,723,001,856 bytes at 2.427 and 2.405 GiB/s. The engine batch moved 328,195,899,392 bytes at 2.688 and 2.659 GiB/s. Both runs restored under the proposed 4 GiB helper limit. All four inference outputs matched exactly before and after restore; all NCCL ranks resumed; GMS restored and validated all four owners while fenced.

For historical before/after context, #121's single-authoritative-read validation used the same requested model and TP4/EP4 + GMS profile, but it was not a counterbalanced same-artifact A/B against this final build:

_Note: This is not a direct #121-versus-candidate A/B because validation includes #119's ordered GMS integration, but some helpful data. These are observed integrated timings, not an isolated attribution to this PR._

| Restore phase | #121 single-reader result | Concurrent-rank exact-final results |
| --- | ---: | ---: |
| GMS target external restore | 147.233 s | 79.933 / 82.442 s |
| GMS CUDA restore | 145.136 s | 78.841 / 81.302 s |
| Engine target external restore | 252.308 s | 124.728 / 123.682 s |
| Engine CUDA restore | 247.008 s | 119.849 / 120.227 s |
| Pod creation to Ready | 422 s | 219 / 226 s |


hhzhang16:

PR #165, opened 2026-09-01
fix(agent): recreate large shared allocations on restore

galletas1712:
## Summary

- preserve large shared POSIX-exportable device allocations in anonymous pinned host memory owned by the workload process before CUDA checkpoint
- release the original exportable device backing before native CUDA checkpoint
- let CRIU capture the pinned host carrier as ordinary process memory; no carrier sidecar files are created
- after restore, create fresh exportable device backing, restore bytes with H2D, then unregister and unmap the host carrier
- transfer creator ranks in parallel and use bounded control timeouts
- force the path in `test_cucheckpoint.py` without changing the production threshold, and assert that no carrier files appear

This is a transparent r615 workaround for large shared allocations whose bytes restore but whose original post-restore POSIX export state is stale. It does not change the driver, disable FLA/FABRIC, require GMS, or retain a second device allocation. Small, non-shared, non-POSIX-exportable, and non-device allocations remain on the native CUDA checkpoint path.

## Validation

- exact 578 MiB two-GPU symmetric-memory A/B:
  - PR #133 without this workaround fails during coordinator restore when creator re-export returns `CUDA_ERROR_INVALID_VALUE`
  - this patch restores successfully and verifies CUDA-graph output and allocation bytes
- native CUDA checkpoint test builds the shim with `HOST_CARRIER_THRESHOLD=1`, exercises the carrier path, and rejects any `cuinterposer-host-*.bin` artifact
- GLM 5.2 / DeepSeek-V4-Flash-NVFP4 SGLang TP8/EP8 on eight B200 GPUs, context length 32768, with warm compile/JIT/FlashInfer caches:
  - source startup: 87.02 s
  - checkpoint: 290.891 s total; CRIU dump 262.580 s; CUDA checkpoint 27.632 s
  - active restore: 37.410 s total; CRIU 12.914 s; CUDA 24.291 s
  - standby restore: 40.165 s total; CRIU 12.921 s; CUDA 27.033 s
  - aggregate CRIU process artifact: about 100 GiB; no separate carrier artifact
  - coherent post-restore and post-promotion inference both returned `17 * 23 = 391`
- clean CUDA-image build with `-Wall -Wextra -Werror`
- `git diff --check`
- `python3 -m py_compile agent/cmd/cuinterpose/tests/test_cucheckpoint.py`

### DIFF
--- a/agent/cmd/cuinterpose/Makefile
+++ b/agent/cmd/cuinterpose/Makefile
@@ -8,6 +8,9 @@ COORDINATOR := $(BUILD_DIR)/cuinterposer-coordinator
 MULTICAST_FORWARD_TEST := $(BUILD_DIR)/multicast-forward-test
 MIN_GLIBC := GLIBC_2.34
 INTERPOSER_CPPFLAGS := -DCUINTERPOSER_DLSYM_VERSION=\"$(MIN_GLIBC)\"
+ifneq ($(HOST_CARRIER_THRESHOLD),)
+INTERPOSER_CPPFLAGS += -DCUINTERPOSER_HOST_CARRIER_THRESHOLD=$(HOST_CARRIER_THRESHOLD)
+endif
 COMMON_FLAGS := -std=gnu11 -O2 -Wall -Wextra -Werror
 INTERPOSER_SOURCES := interpose.c multicast.c posix.c symbols.c util.c
 INTERPOSER_HEADERS := multicast.h posix.h protocol.h symbols.h util.h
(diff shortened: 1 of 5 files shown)


galletas1712:
Superseded by #218 (lifecycle: host carrier for every tracked creator allocation, restored through one pinned arena per process). The re-cut stack (#212 → #220) supersedes this PR. Its body is quoted verbatim in the replacement's Origin section, and each review thread here has a row in the replacement's "Review threads carried" table with what was done about it.

PR #166, opened 2026-09-02
fix(agent): collapse duplicate CUDA handle references

galletas1712:
## Summary

- collapse redundant driver references acquired by `cuMemRetainAllocationHandle` when a tracked allocation already has a live backing handle
- apply the same one-backing-reference invariant to repeated tracked POSIX imports
- preserve distinct logical handles while releasing the redundant physical driver reference immediately
- add a fake-CUDA regression that verifies alias release keeps the backing alive and final release drops it exactly once

## Validation

- `make -C agent/cmd/cuinterpose clean test` in the CUDA test image
- `go test ./agent/internal/criu ./agent/internal/cuda -count=1`
- live 1 GiB A/B allocation test:
  - before: 1 GiB remained after releasing all three logical aliases
  - after: residual allocation was 0 bytes
- GLM-5.2-NVFP4 SGLang TP8/EP8, context length 32768, checkpoint captured successfully with all eight GMS devices saving 60,129,542,144 bytes
- two restored engines completed CUDA restore; active inference returned `42`
- after active-engine failure, the shadow acquired the lock in 135 ms, registered in 4.70 s, and post-failover inference returned `42`
- after active teardown, GPU memory returned exactly to the single-engine baseline

### DIFF
--- a/agent/cmd/cuinterpose/interpose.c
+++ b/agent/cmd/cuinterpose/interpose.c
@@ -1523,7 +1523,9 @@ CUresult CUDAAPI
 cuMemRetainAllocationHandle(CUmemGenericAllocationHandle* output, void* address)
 {
   retain_fn function = (retain_fn)cuinterposer_lookup_real_symbol("cuMemRetainAllocationHandle");
+  release_fn release = (release_fn)cuinterposer_lookup_real_symbol("cuMemRelease");
   struct mapping* mapping;
+  struct handle* existing;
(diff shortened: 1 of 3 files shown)


galletas1712:
Superseded by #217 (tracking: one driver handle per allocation, aliases collapsed through the tables). The re-cut stack (#212 → #220) supersedes this PR. Its body is quoted verbatim in the replacement's Origin section, and each review thread here has a row in the replacement's "Review threads carried" table with what was done about it.

PR #167, opened 2026-09-02
fix(agent): rewrite bound abstract UNIX sockets on clone restore

galletas1712:
## Summary

- broaden clone-restore socket rewriting from the CUDA UVM FD listener only to every retained bound abstract UNIX socket
- detect bound abstract sockets by the leading NUL namespace marker (`len(name) > 1 && name[0] == 0`); skip unbound sockets (empty name or bare `\0`)
- index UNIX sockets by inode before rewriting so connected pairs keep peer links and receive distinct clone-private `dynamo-` addresses derived from the restore ID and original name
- fail restore preparation when a bound abstract socket references a peer inode missing from the checkpoint image
- leave filesystem-path UNIX sockets and TCP port rewriting unchanged

The previous rewrite matched only `SOCK_SEQPACKET` CUDA UVM FD listeners with no peer. Retained NCCL datagram sockets (for example `\0tmp/nccl-socket-…`) and other bound abstract endpoints otherwise restore with their original addresses and collide with `EADDRINUSE` when a second clone lands on the same node.

## Validation

- `go test ./agent/internal/criu ./agent/internal/cuda -count=1`
- A/B restore on the same checkpoint and node:
  - without rewriting retained NCCL datagram addresses, the second clone failed with `Address already in use`
  - with the rewrite, both GLM-5.2-NVFP4 SGLang TP8/EP8 engines restored successfully
- active inference returned `42`; after killing the active restored engine, the shadow registered in 4.70 s and returned `42`

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved restoration of abstract UNIX sockets by correctly rewriting addresses for connected internal socket pairs.
  - Preserved connections to externally managed sockets.
  - Skipped unbound sockets and rejected incomplete socket metadata to prevent invalid restorations.
  - Improved handling of distinct datagram socket addresses during restoration.

- **Tests**
  - Added coverage for UNIX socket rewriting, external peers, missing metadata, unbound sockets, and long socket names.

### DIFF
--- a/agent/internal/criu/restore_images.go
+++ b/agent/internal/criu/restore_images.go
@@ -4,7 +4,6 @@
 package criu
 
 import (
-	"bytes"
 	"crypto/sha256"
 	"encoding/binary"
 	"encoding/hex"
@@ -24,8 +23,8 @@ import (
 const (
 	filesImageFilename            = "files.img"
 	placeholderMountNamespacePath = "/proc/self/ns/mnt"
-	cudaUVMFDSocketNamePrefix     = "\x00cuda-uvmfd-"
 	linuxUnixSocketStateListen    = 10
+	linuxUnixSocketStateClose     = 7
 	linuxTCPStateEstablished      = 1
 	linuxTCPStateClose            = 7
 	linuxTCPStateListen           = 10
@@ -137,15 +136,38 @@ func rewriteSocketMetadata(image *crit.CriuImage, restoreID uint64) ([]int, bool
 		forbiddenPorts[port] = struct{}{}
 	}
 
-	rewritten := false
+	unixSockets := make(map[uint32]*sk_unix.UnixSkEntry)
 	for i, entry := range image.Entries {
 		fileEntry, ok := entry.Message.(*fdinfo.FileEntry)
 		if !ok {
 			closeFDs(reservationFDs)
 			return nil, false, fmt.Errorf("unexpected %s entry %d type %T", filesImageFilename, i, entry.Message)
 		}
-		if fileEntry.GetType() == fdinfo.FdTypes_UNIXSK && fileEntry.Usk != nil &&
-			rewriteCloneConflictingUnixSocketAddress(fileEntry.Usk, restoreID) {
+		if fileEntry.GetType() == fdinfo.FdTypes_UNIXSK && fileEntry.Usk != nil {
+			unixSockets[fileEntry.Usk.GetIno()] = fileEntry.Usk
+		}
+	}
+
+	rewritten := false
+	for _, socket := range unixSockets {
+		if !isBoundAbstractUnixSocket(socket) {
+			continue
+		}
+		if socket.Ino == nil || socket.Peer == nil {
+			closeFDs(reservationFDs)
+			return nil, false, fmt.Errorf("bound abstract UNIX socket is missing inode or peer metadata")
+		}
+		if peer := socket.GetPeer(); peer != 0 {
+			if _, ok := unixSockets[peer]; !ok {
+				closeFDs(reservationFDs)
+				return nil, false, fmt.Errorf(
+					"bound abstract UNIX socket inode %d has unresolved peer inode %d",
+					socket.GetIno(),
+					peer,
+				)
+			}
+		}
+		if rewriteCloneConflictingUnixSocketAddress(socket, restoreID) {
 			rewritten = true
 		}
 	}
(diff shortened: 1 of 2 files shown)


galletas1712:
Superseded by #174, which includes the NCCL socket isolation fix and retargets the full stack on main.

agent/internal/criu/restore_images.go
```diff
@@ -137,15 +136,38 @@ func rewriteSocketMetadata(image *crit.CriuImage, restoreID uint64) ([]int, bool
...
 	for i, entry := range image.Entries {
 		fileEntry, ok := entry.Message.(*fdinfo.FileEntry)
 		if !ok {
 			closeFDs(reservationFDs)
 			return nil, false, fmt.Errorf("unexpected %s entry %d type %T", filesImageFilename, i, entry.Message)
 		}
-		if fileEntry.GetType() == fdinfo.FdTypes_UNIXSK && fileEntry.Usk != nil &&
-			rewriteCloneConflictingUnixSocketAddress(fileEntry.Usk, restoreID) {
+		if fileEntry.GetType() == fdinfo.FdTypes_UNIXSK && fileEntry.Usk != nil {
+			unixSockets[fileEntry.Usk.GetIno()] = fileEntry.Usk
```
dfeigin-nv: a socket with no inode becomes key 0 and overwrites any earlier one, I think

agent/internal/criu/restore_images.go
```diff
@@ -416,13 +437,7 @@ func rewriteCloneConflictingUnixSocketAddress(entry *sk_unix.UnixSkEntry, restor
...
-func isCUDAUVMFDListener(entry *sk_unix.UnixSkEntry) bool {
-	return entry != nil &&
-		entry.Type != nil &&
-		entry.State != nil &&
-		entry.Peer != nil &&
-		*entry.Type == uint32(unix.SOCK_SEQPACKET) &&
-		*entry.State == linuxUnixSocketStateListen &&
-		*entry.Peer == 0 &&
-		bytes.HasPrefix(entry.Name, []byte(cudaUVMFDSocketNamePrefix))
+func isBoundAbstractUnixSocket(entry *sk_unix.UnixSkEntry) bool {
```
dfeigin-nv, line 451: Do we mind that we may catch more than just CUDA sockets?

agent/internal/criu/restore_images.go
```diff
@@ -137,15 +136,38 @@ func rewriteSocketMetadata(image *crit.CriuImage, restoreID uint64) ([]int, bool
...
+		}
+		if socket.Ino == nil || socket.Peer == nil {
+			closeFDs(reservationFDs)
+			return nil, false, fmt.Errorf("bound abstract UNIX socket is missing inode or peer metadata")
+		}
+		if peer := socket.GetPeer(); peer != 0 {
+			if _, ok := unixSockets[peer]; !ok {
+				closeFDs(reservationFDs)
+				return nil, false, fmt.Errorf(
+					"bound abstract UNIX socket inode %d has unresolved peer inode %d",
```
dfeigin-nv, line 175: a socket connected to something outside the checkpoint now fails the restore

This isn't new right?

dfeigin-nv (approved):

agent/internal/criu/restore_images.go
galletas1712: The broader match is intentional: every restored, owned abstract binding needs a clone-private name when multiple containers restore into the same network namespace. f3eb92e narrows this at the important boundary: CRIU USK_EXTERN entries are synthetic connection targets, not restored bindings, so they are preserved. It also corrects bare-NUL abstract addresses as bound rather than unbound.

agent/internal/criu/restore_images.go
galletas1712: A valid CRIU ext-unix-sk image includes a synthetic USK_EXTERN entry for the outside peer, so a genuinely unresolved peer still indicates an incomplete image. The old test did not model that. f3eb92e replaces it with a realistic external-peer test: the restored client binding is rewritten, while the USK_EXTERN target address and peer link are preserved for CRIU reconnection.

agent/internal/criu/restore_images.go
galletas1712: Fixed in f3eb92e. UNIX socket inode metadata is now validated before map insertion, and duplicate inodes are rejected instead of silently overwriting an entry (including key 0). I also added a missing-inode regression test.

PR #171, opened 2026-09-02
docs(guides): align framework examples on small single-GPU sizing and pins

oleg-kushniriov:
> Stacked on #170 (`feat/snapshot-restore-standby-env`). Review only the last commit until #170 merges; the base will then be retargeted to `main`.

### Summary

Makes the vLLM, SGLang, and TensorRT-LLM guide programs suitable as e2e workloads while keeping them valid user examples. Groundwork for per-framework checkpoint/restore e2e tests.

### Changes

- **Engine sizing via env with small single-GPU defaults** (source manifests set the same values explicitly):
  - vLLM: `SNAPSHOT_MAX_MODEL_LEN=2048`, `SNAPSHOT_GPU_MEMORY_UTILIZATION=0.30`
  - SGLang: `SNAPSHOT_PAGE_SIZE=16`, plus `tp_size=1`, `trust_remote_code=True`
  - TensorRT-LLM: `SNAPSHOT_MAX_NUM_TOKENS=1024`, `SNAPSHOT_MAX_BATCH_SIZE=1`, `SNAPSHOT_FREE_GPU_MEMORY_FRACTION=0.10` (previously hardcoded to the same values)
- **`<framework>-precheck` sentinel**: each program writes its pre-checkpoint generation output to `$SNAPSHOT_CONTROL_DIR/{vllm,sglang,trtllm}-precheck`. The dump kills the source container, so this gives tests durable evidence that the engine served a generation before capture without relying on logs.
- **Runtime image pins**: SGLang `lmsysorg/sglang@sha256:9e148f…` → `lmsysorg/sglang:v0.5.17-cu130-runtime`; TensorRT-LLM `1.3.0rc23` → `1.3.0rc24`. Dockerfile `ARG` defaults and the guide `export` lines updated together.
- Guide text updated for the new variables and sentinel files.

### Not in this PR

- Model-cache init container / PVC for vLLM and TensorRT-LLM (SGLang already has one) — follow-up.
- The e2e tests themselves — follow-up on top of this branch.

### Verification

- `python3 -m py_compile` on the three `app.py` — ok
- All three `deployment.yaml` parse
- No stale pins left (`grep sha256:9e148f|1.3.0rc23 docs/` — empty)
- Image bumps are **not** yet validated on a GPU cluster; if either breaks checkpoint/restore in the first e2e run, that single pin will be reverted.

### Review follow-ups

[...]

PR #170, opened 2026-09-02
docs(guides): standardize on SNAPSHOT_RESTORE_STANDBY in guides, examples, and e2e

oleg-kushniriov:
### Summary

Introduces `SNAPSHOT_RESTORE_STANDBY` as the single restore standby environment variable and removes the deprecated `DYN_SNAPSHOT_RESTORE_STANDBY` name everywhere in the repo.

### Why

- `api/podcontract` already declared `SNAPSHOT_RESTORE_STANDBY` as canonical and the `DYN_` name as deprecated, but the guide programs and e2e still used only the deprecated one.
- Nothing outside this repo imports `podcontract.LegacyRestoreStandbyModeEnv` or the `podcontract` package (GitHub-wide code search). Snapshot itself never read or injected the variable — it only published the name — so removing it changes no controller/agent behavior.

### Changes

- `api/podcontract`: remove `LegacyRestoreStandbyModeEnv`; test asserts only the canonical name is not injected by the generic builder.
- `docs/reference/restore-pod-contract.md`: describe the single name and the producer/workload roles.
- `docs/guides/{vllm,sglang,tensorrt-llm}/app.py`: read `SNAPSHOT_RESTORE_STANDBY`.
- `docs/guides/*/restore-deployment.yaml`: set `SNAPSHOT_RESTORE_STANDBY=1` on `main`. **Bug fix:** these manifests set no standby variable at all, so as documented the placeholder would start loading a model into the GPU while the agent restored into the same container.
- `agent/Dockerfile`: comment updated to the neutral name.
- `e2e/snapshot_e2e/workloads.py`: restore Pods set `SNAPSHOT_RESTORE_STANDBY=1`.

### Breaking change

`podcontract.LegacyRestoreStandbyModeEnv` is removed and the guide `app.py` programs no longer honor `DYN_SNAPSHOT_RESTORE_STANDBY`. Producers of restore Pods for these workloads must set `SNAPSHOT_RESTORE_STANDBY=1`.

### Verification

- `go vet ./podcontract/ && go test ./podcontract/` — ok
- `uv run --project e2e pytest e2e/tests/test_workload_scripts.py` — 5 passed
- Repo-wide grep for `DYN_SNAPSHOT_RESTORE_STANDBY` / `LegacyRestoreStandbyModeEnv` — no matches
- All three edited `restore-deployment.yaml` parse

## Summary by CodeRabbit

* **Documentation**
[...]

PR #173, opened 2026-09-02
ci: publish content-addressed framework e2e images

oleg-kushniriov:
> Stacked on #171 → #170 (stack #172). Review only the last commit until the lower PRs merge.

### Summary

Publishes the framework guide images (`docs/guides/{vllm,sglang,tensorrt-llm}/`) that the upcoming framework e2e tests run as workloads. Images are content-addressed, so they are built once per distinct set of image inputs, not per commit.

### Changes

- `hack/framework-image-tag.py <framework> [--check] [--allow-missing] [--github-output]`
  - Resolves `ghcr.io/ai-dynamo/snapshot/e2e-<framework>:<sha256[:12]>` over the guide Dockerfile (which carries the runtime base pin) and the files it copies. Deployment manifests are excluded so Pod-level edits never force a rebuild.
  - `--check` queries GHCR (404 on a not-yet-created package = not published) and fails unless the tag exists; `--github-output` appends `image/tag/exists/dockerfile/context`.
- `.github/workflows/e2e-framework-images.yaml`
  - Matrix `vllm | sglang | tensorrt-llm`, `ubuntu-latest`, 90-min cap. Resolves the tag → skips when published → frees hosted-runner disk (runtime bases are 10–30 GB) → `docker buildx build --platform linux/amd64 --push` with `type=gha` cache.
  - Triggers: `push` to `main` and `pull_request` on `docs/guides/**`, the script, and the workflow; `workflow_dispatch` with `framework` filter and `force`. Same-repo PRs publish (the tag is derived from the PR's content); fork PRs are skipped since the default token cannot push.
- `e2e/README.md`: "Framework Images" section (scheme, resolve/check, local build).

### Verification

- Script: deterministic tags for all three frameworks; `--check` correctly reports unpublished (exit 1 strict, 0 with `--allow-missing`); `GITHUB_OUTPUT` lines verified.
- Workflow YAML parses; first real run is this PR's own `E2E Framework Images` check — it will also be the first validation of the SGLang `v0.5.17-cu130-runtime` and TensorRT-LLM `1.3.0rc24` bumps from #171 building at all.

### Risks

- Hosted-runner disk: TensorRT-LLM's base is the largest. If the build still runs out of space after the cleanup step, the fallback is a larger runner label for that matrix entry.
- `nvcr.io/nvidia/tensorrt-llm/release` is pulled anonymously; if NGC requires auth for this tag the job will fail at `FROM` and a pull secret must be added.

### Review follow-ups

- `package_has_tag` pages the GHCR listing to exhaustion (early exit on match) instead of a fixed 5 pages, so accumulated content-addressed tags cannot push a valid tag out of the window.
- The images workflow also builds on pushes to `pull-request/[0-9]+` mirror branches, so fork PRs that touch a guide get an image from the trusted mirror push; the fork-PR job `if:` comment now describes what it does (early exit for the untrusted event).
- Dropped the buildx `type=gha` cache (only the app layers are new; `mode=max` uploads competed with the Go caches).
- "Not published" message printed once.

## Summary by CodeRabbit

- **New Features**
  - Added automated publishing of framework guide images for vLLM, SGLang, and TensorRT-LLM.
  - Added options to build all frameworks or select an individual framework, with an option to force rebuilding.
  - Added content-based image tags to identify images built from specific guide content.
  - Added checks to avoid rebuilding images that are already published.

- **Documentation**
  - Added guidance for resolving framework images, understanding image tags, and building images locally.

### DIFF
--- a/hack/framework-image-tag.py
+++ b/hack/framework-image-tag.py
@@ -54,7 +54,13 @@ def image_inputs(framework: str) -> list[Path]:
 
 
 def dockerfile(framework: str) -> Path:
-    return image_inputs(framework)[0]
+    # Named, not positional: the hash-input tuple has no ordering contract,
+    # and building from the wrong file would still publish a validly tagged
+    # image.
+    path = GUIDES_DIR / framework / f"Dockerfile.{framework}"
+    if path.name not in FRAMEWORKS[framework]:
+        raise ValueError(f"{path.name} is not a hash input for {framework}")
+    return path
 
 
 def inputs_digest(framework: str) -> str:
--- a/hack/framework-image-tag.py
+++ b/hack/framework-image-tag.py
@@ -30,7 +30,7 @@ import urllib.request
 from pathlib import Path
 
 GITHUB_API_VERSION = "2022-11-28"
-MAX_PAGES = 5
+PAGE_SIZE = 100
 ORG = "ai-dynamo"
 REGISTRY = f"ghcr.io/{ORG}/snapshot"
 TAG_LENGTH = 12
@@ -81,10 +81,15 @@ def image_ref(framework: str) -> tuple[str, str]:
 def package_has_tag(package: str, tag: str, headers: dict[str, str]) -> bool:
     encoded = urllib.parse.quote(package, safe="")
 
-    for page in range(1, MAX_PAGES + 1):
+    # Page until the listing is exhausted: content-addressed tags accumulate
+    # (one per guide change, never pruned), so a fixed page budget would
+    # eventually report a still-valid tag as unpublished.
+    page = 0
+    while True:
+        page += 1
         url = (
             f"https://api.github.com/orgs/{ORG}/packages/container/"
-            f"{encoded}/versions?per_page=100&page={page}"
+            f"{encoded}/versions?per_page={PAGE_SIZE}&page={page}"
         )
         request = urllib.request.Request(url, headers=headers)
         try:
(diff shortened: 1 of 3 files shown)


hack/framework-image-tag.py
```diff
@@ -0,0 +1,195 @@
...
+TAG_LENGTH = 12
+
+REPO_ROOT = Path(__file__).resolve().parent.parent
+GUIDES_DIR = REPO_ROOT / "docs" / "guides"
+
+# Files that end up in the image, per framework guide directory. The Dockerfile
+# carries the runtime base image pin, so a base bump changes the tag too.
+# Deployment manifests are deliberately excluded: they configure the Pod, not
+# the image, and must not force a rebuild.
+FRAMEWORKS: dict[str, tuple[str, ...]] = {
```
shayasoolin, line 50: I see this point as a critical one. I'd emphasize the importance of this dict in the comment above.

My concern about FRAMEWORKS: if it is a whitelist (list of files to include in the tag calculation), it is easy to add a file to the Dockerfile and forget to add it here. Then we can reuse an incorrect image tag.

If we make it a blacklist instead (i.e. list the files to exclude rather than the ones to include for the tag calculation), we may rebuild too often when unrelated files are added.

Maybe keep an explicit yes/no classification for every file in each guide directory: image input or excluded. The script can scan the folder and fail if it finds an unclassified file. This makes missing new files visible immediately, without unnecessary rebuilds.

oleg-kushniriov:
Closing: the guides now run the stock framework images with app.py mounted from a ConfigMap (see #176), so there is no Snapshot-built framework image to publish. The tag resolver and the E2E Framework Images workflow were dropped from the stack.

PR #174, opened 2026-09-02
fix(agent): restore restarted destination containers

galletas1712:
## Summary

- persist the container ID restored for each destination in the agent-owned `nvidia.com/restored-container-ids` Pod annotation
- use the restore informer only to enqueue Pod keys, then read the Pod directly from the API server before terminal and CRIU decisions so delayed informer updates cannot replay a completed restore
- when kubelet replaces a container after a fully successful restore, restore the new container while leaving destinations whose container IDs have not changed untouched
- keep failed and partially successful restores terminal; higher-level retry policy continues to require a new restore Pod
- treat successful restores without the annotation as legacy state and never replay CRIU into them
- serialize updates from concurrent destination workers so the annotation retains every destination record

The node agent owns container-level restore safety and runtime identity. The operator and CRDs are unchanged.

Independent of #167.

## Validation

- `go test ./agent/... ./api/... -count=1`
- `go test -race ./agent/internal/controller/... -count=1`
- stale informer state with an up-to-date API-server annotation does not trigger another restore
- API-server read failures requeue without invoking CRIU
- source-identical controller/API code exercised by GLM 5.2 TP8/EP8 dual-clone restore and successful shadow promotion on r615
- promoted shadow returned a coherent post-failover inference result

## Summary by CodeRabbit

* **Bug Fixes**
  * Restore tracking now persists across controller restarts, preventing duplicate restores.
  * Restores are retried only for destination containers that have restarted.
  * Untracked containers remain untouched during replenishment.
  * Invalid restore records are rejected and logged safely.
  * Restore processing uses current Pod state and retries when API reads fail.
  * Interrupted restores no longer replay unsafe operations.
  * Compatibility checks remain enforced for replacement restores.
  * Terminal restore events are deduplicated and cleared when no longer relevant.
  * Replenishment outcomes are correctly reported as pending or failed.

* **Documentation**
  * Documented container-replacement restore behavior and requirements.

### DIFF
--- a/agent/internal/controller/controller.go
+++ b/agent/internal/controller/controller.go
@@ -36,7 +36,6 @@ import (
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
-	corev1listers "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
@@ -71,12 +70,11 @@ type NodeController struct {
 	controlSentinelExistsFn func(int, string) (bool, error)
 	sendSignalFn            func(logr.Logger, int, syscall.Signal, string) error
 	restoreQueue            workqueue.TypedDelayingInterface[client.ObjectKey]
-	restorePodLister        corev1listers.PodLister
 
 	inFlight   map[string]struct{}
 	inFlightMu sync.Mutex
 
-	handledRestores sync.Map
+	restoredContainerIDsMu sync.Mutex
 
 	// contentIndexer is the PodSnapshotContent informer's indexer, indexed by source pod
 	// (podRefIndex). The source-pod informer uses it to map a pod event back to its work order.
@@ -249,13 +247,11 @@ func (w *NodeController) Run(ctx context.Context) error {
 
 	restorePods := restoreFactory.Core().V1().Pods()
 	restoreInformer := restorePods.Informer()
-	w.restorePodLister = restorePods.Lister()
 	if _, err := restoreInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
 		AddFunc: w.enqueueRestorePod,
 		UpdateFunc: func(_, newObj interface{}) {
 			w.enqueueRestorePod(newObj)
 		},
-		DeleteFunc: w.forgetRestorePod,
 	}); err != nil {
 		return fmt.Errorf("failed to add restore informer handler: %w", err)
 	}
@@ -373,14 +369,6 @@ func (w *NodeController) enqueueRestorePod(obj interface{}) {
 	w.restoreQueue.Add(client.ObjectKeyFromObject(pod))
 }
 
-func (w *NodeController) forgetRestorePod(obj interface{}) {
-	pod, ok := podFromInformerObj(obj)
-	if !ok {
-		return
-	}
-	w.handledRestores.Delete(string(pod.UID))
-}
-
 func (w *NodeController) restorePodRequested(pod *corev1.Pod) bool {
 	if pod.Spec.NodeName != w.config.NodeName {
 		return false
@@ -413,21 +401,18 @@ func (w *NodeController) processRestoreQueueItem(ctx context.Context, key client
 		}
 	}()
 
-	pod, err := w.restorePodLister.Pods(key.Namespace).Get(key.Name)
+	// The informer only supplies work keys. Read directly from the API server
+	// before CRIU decisions so a delayed annotation event cannot replay restore.
+	pod, err := w.clientset.CoreV1().Pods(key.Namespace).Get(ctx, key.Name, metav1.GetOptions{})
 	if apierrors.IsNotFound(err) {
 		return
 	}
 	if err != nil {
-		w.log.Error(err, "Failed to read restore pod from informer cache", "pod", key.String())
+		w.log.Error(err, "Failed to read restore pod from API server", "pod", key.String())
 		requeue = true
 		return
 	}
-	pod = pod.DeepCopy()
-	if w.restoreHandled(pod) {
-		requeue = w.removeRestoreFinalizerWithEvent(ctx, pod)
-		return
-	}
-	if isRestoreTerminal(pod) {
+	if isRestoreTerminal(pod) && !w.hasRestartedRestoreDestination(pod) {
 		requeue = w.handleTerminalRestorePod(ctx, pod)
 		return
 	}
@@ -648,6 +633,11 @@ func (w *NodeController) restorePodContainers(ctx context.Context, pod *corev1.P
 	// RestoreInProgress makes each worker check its completion sentinel before
 	// considering a CRIU replay.
 	recovering := restoreInProgress(pod)
+	replenishing := isRestoreSucceeded(pod)
+	restoredIDs, err := restoredContainerIDs(pod)
+	if err != nil {
+		return w.failRestorePod(ctx, pod, err)
+	}
 	message := fmt.Sprintf("Restoring %d destination container(s) from PodSnapshot %s", len(plan.mappings), plan.artifact.SnapshotName)
 	if err := w.applyRestoredCondition(ctx, pod, corev1.ConditionFalse, podcontract.RestoreReasonInProgress, message); err != nil {
 		emitPodEvent(ctx, w.clientset, w.log, pod, snapshotEventComponent, corev1.EventTypeWarning, restoreStatusUpdateFailedReason, err.Error())
@@ -659,7 +649,7 @@ func (w *NodeController) restorePodContainers(ctx context.Context, pod *corev1.P
 	for i, mapping := range plan.mappings {
 		i, destination := i, mapping.Destination
 		workers.Go(func() {
-			results[i] = w.restoreDestination(ctx, pod, plan.artifact, destination, podKey, recovering)
+			results[i] = w.restoreDestination(ctx, pod, plan.artifact, destination, podKey, recovering, replenishing, restoredIDs)
 		})
 	}
 	workers.Wait()
@@ -708,11 +698,19 @@ func (w *NodeController) restoreDestination(
 	pod *corev1.Pod,
 	artifact *restoreArtifact,
 	destination, podKey string,
-	recovering bool,
+	recovering, replenishing bool,
+	restoredIDs map[string]string,
 ) restoreResult {
 	result := restoreResult{destination: destination, state: restoreResultPending}
 	containerID, _ := w.resolveRestoreContainerID(ctx, pod, destination, podKey)
 	if containerID == "" {
+		return result
+	}
+	recordedID, tracked := restoredIDs[destination]
+	// An unchanged ID already holds the restored process. An untracked
+	// destination on a replenishment pass predates tracking and is left alone.
+	if (tracked && recordedID == containerID) || (!tracked && replenishing) {
+		result.state = restoreResultSucceeded
 		return result
 	}
 
@@ -729,6 +727,14 @@ func (w *NodeController) restoreDestination(
 		result.state = restoreResultFailed
 		log.Error(err, "Restore controller worker failed")
 		emitPodEvent(ctx, w.clientset, log, pod, snapshotEventComponent, corev1.EventTypeWarning, "RestoreWorkerFailed", err.Error())
+		return result
+	}
+	// Without a persisted container ID the next pass could replay CRIU
+	// into this live restored engine, so a failed record fails the destination.
+	if err := w.recordRestoredContainerID(ctx, pod, destination, containerID); err != nil {
+		result.state = restoreResultFailed
+		log.Error(err, "Failed to record restored container ID")
+		emitPodEvent(ctx, w.clientset, log, pod, snapshotEventComponent, corev1.EventTypeWarning, "RestoreRecordFailed", err.Error())
 		return result
 	}
 	result.state = restoreResultSucceeded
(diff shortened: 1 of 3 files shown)


galletas1712:
Resetting GitHub mergeability after retargeting onto main with only the incarnation-tracking commits.

Ronkahn21 (commented):
I think we could simplify this and allow replay to succseded/parialSecceded  without getting into details of which container was successful,

the idea is we just dont get into the loop of failed restore all over

galletas1712:
I considered gating only on the aggregate outcome, but I think we still need the recorded container IDs. A successful Pod can be reconciled on ordinary updates, so allowing every Succeeded outcome to replay would create a restore loop unless we can distinguish a restarted container from one already restored.

PartiallySucceeded is also unsafe to replay without per-destination state since we would not know which destinations failed, and could replay CRIU into a successful live destination.

That is why I keep Failed and PartiallySucceeded terminal and only restore a destination after Succeeded when its current container ID differs from the recorded ID.

agent/internal/controller/controller.go
oleg-kushniriov, line 880: recordRestoredContainerID writes to the shared `pod.Annotations` map under a mutex, but `executeRestore` reads from that same map with no lock at all - with 2+ destinations restoring at once, that's a data race that can crash the agent. (`concurrent map read and map write`)

agent/internal/controller/controller.go
```diff
@@ -973,13 +979,83 @@ func (w *NodeController) patchRestoreFinalizers(ctx context.Context, pod *corev1
...
+
+	w.restoredContainerIDsMu.Lock()
+	defer w.restoredContainerIDsMu.Unlock()
+
+	ids, err := restoredContainerIDs(pod)
+	if err != nil {
+		// Invalid records cannot prove that a running container is safe to replace.
+		return false
+	}
+	for _, status := range pod.Status.ContainerStatuses {
```
oleg-kushniriov, line 1013: For a genuinely fast crash loop, the Running window can be far shorter than the Waiting{CrashLoopBackOff} window (which itself grows toward minutes), so a Get landing during Running gets less likely over time, not more. Every miss makes hasRestartedRestoreDestination return false, so processRestoreQueueItem routes into handleTerminalRestorePod, which emits RestoreAlreadyCompleted and strips the finalizer - a false "done" signal repeated on every failed detection attempt, not a quiet wait.

Suggestion: Drop the State.Running gate; compare status.ContainerID whenever non-empty, or track RestartCount per destination.

agent/internal/controller/controller.go
```diff
@@ -973,13 +979,83 @@ func (w *NodeController) patchRestoreFinalizers(ctx context.Context, pod *corev1
...
+		return err
+	}
+	if _, err := w.clientset.CoreV1().Pods(pod.Namespace).Patch(ctx, pod.Name, ktypes.MergePatchType, patch, metav1.PatchOptions{}); err != nil {
+		return fmt.Errorf("record restored container ID: %w", err)
+	}
+	if pod.Annotations == nil {
+		pod.Annotations = map[string]string{}
+	}
+	pod.Annotations[podcontract.RestoredContainerIDsAnnotation] = string(value)
+	return nil
```
Ronkahn21: this is local duplicate to line 1041-1053
lets first copy the pod than change and then patch

agent/internal/controller/controller.go
```diff
@@ -973,13 +979,83 @@ func (w *NodeController) patchRestoreFinalizers(ctx context.Context, pod *corev1
 	return err
 }
 
-func (w *NodeController) restoreHandled(pod *corev1.Pod) bool {
-	_, handled := w.handledRestores.Load(string(pod.UID))
-	return handled
+func restoredContainerIDs(pod *corev1.Pod) (map[string]string, error) {
+	ids := make(map[string]string)
+	raw := pod.Annotations[podcontract.RestoredContainerIDsAnnotation]
```
Ronkahn21, line 1160: in case the stale cache you might rerun restore even if alraedy restored the pod

agent/internal/controller/controller.go
galletas1712: Fixed in 1846173. executeRestore now takes restoredContainerIDsMu for the Page Broker annotation read and releases it before invoking restoreFn. The annotation writer uses the same lock. Added TestPageBrokerReadConcurrentWithContainerIDRecords to exercise concurrent reads and writes; the controller suite passed go test -race ./agent/internal/controller/... -count=5.

agent/internal/controller/controller.go
galletas1712: Agreed; fixed in 1846173. Restart detection now compares every nonempty container ID without requiring State.Running. The worker still separately resolves a running container before execution, so a detected replacement in backoff remains pending instead of taking the terminal-completion path. Added Running/Waiting/Terminated and empty-ID cases, plus a waiting-to-running regression checking no RestoreAlreadyCompleted event, retained finalizer, unchanged record while pending, and exactly one restore when the replacement becomes available. Controller race tests passed five consecutive runs. GPU validation remains unverified because the vCluster control plane was already crash-looping; no cluster resources were changed.

agent/internal/controller/controller.go
```diff
@@ -973,13 +979,83 @@ func (w *NodeController) patchRestoreFinalizers(ctx context.Context, pod *corev1
...
+	if err != nil {
+		return err
+	}
+	if _, err := w.clientset.CoreV1().Pods(pod.Namespace).Patch(ctx, pod.Name, ktypes.MergePatchType, patch, metav1.PatchOptions{}); err != nil {
+		return fmt.Errorf("record restored container ID: %w", err)
+	}
+	if pod.Annotations == nil {
+		pod.Annotations = map[string]string{}
+	}
+	pod.Annotations[podcontract.RestoredContainerIDsAnnotation] = string(value)
```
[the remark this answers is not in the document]
galletas1712: Fixed in 1846173 by synchronizing the Page Broker annotation read with restoredContainerIDsMu, the same mutex used by recordRestoredContainerID. The lock is released before restore execution. This keeps the change local without adding worker parameters. Added concurrent read/write coverage; go test -race ./agent/internal/controller/... -count=5 passed.

agent/internal/controller/controller.go
```diff
@@ -970,13 +978,83 @@ func (w *NodeController) patchRestoreFinalizers(ctx context.Context, pod *corev1
...
+	}
+
+	w.restoredContainerIDsMu.Lock()
+	defer w.restoredContainerIDsMu.Unlock()
+
+	ids, err := restoredContainerIDs(pod)
+	if err != nil {
+		// Invalid records cannot prove that a running container is safe to replace.
+		return false
+	}
```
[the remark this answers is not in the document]
galletas1712: Addressed in 1846173. Added the error log with Pod identity and TestInvalidRestoredIDsAreLogged, retaining the conservative no-replay result on corrupt records. For the key-space concern, mapping destinations are literal container names, so they already match ContainerStatus.Name; I documented that invariant rather than adding a second mapping resolution path. The mapped-destination tests cover source main with destination engine-1.

agent/internal/controller/controller_test.go
```diff
@@ -1021,6 +1085,86 @@ func TestProcessRestoreQueueItemEmitsEventWhenRestoreAlreadyCompleted(t *testing
...
+	}
+
+	processQueuedRestorePod(t, w, pod)
+
+	require.Len(t, restored, 1, "a live restored sibling must not be replayed")
+	assert.Equal(t, "engine-1", restored[0].DestinationContainerName)
+	assert.Equal(t, "engine-1-restarted-id", restored[0].ContainerID)
+	assert.Equal(t, "engine-0-id", liveRestoredContainerID(t, w, pod, "engine-0"))
+	assert.Equal(t, "engine-1-restarted-id", liveRestoredContainerID(t, w, pod, "engine-1"))
+}
```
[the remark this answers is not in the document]
galletas1712: Added TestReplenishmentSkipsUntrackedDestination in 1846173. It calls restorePodContainers directly with a succeeded Pod, unchanged tracked engine-0, and an untracked replacement engine-1. It verifies no restore invocation, the unchanged engine-0 record, and no record created for engine-1. Calling the coordinator directly ensures this exercises the worker guard rather than stopping at the queue-level restart gate. Unit and race tests pass.

agent/internal/controller/controller.go
```diff
@@ -648,6 +635,8 @@ func (w *NodeController) restorePodContainers(ctx context.Context, pod *corev1.P
 	// RestoreInProgress makes each worker check its completion sentinel before
 	// considering a CRIU replay.
 	recovering := restoreInProgress(pod)
+	replenishing := isRestoreSucceeded(pod)
+	restoredIDs, _ := restoredContainerIDs(pod)
```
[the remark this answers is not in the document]
galletas1712: The malformed-JSON and JSON-null rejection and pre-worker failure coverage are preserved. In 1846173 I also added the missing error log and logging regression for corrupt terminal records. To clarify the chosen behavior: terminal restart detection still returns false rather than propagating the error to requeue, consistent with the later review suggestion. Corrupt state cannot authorize CRIU replay and is now operator-visible.

agent/internal/controller/controller.go
```diff
@@ -973,13 +979,83 @@ func (w *NodeController) patchRestoreFinalizers(ctx context.Context, pod *corev1
...
+	ids := make(map[string]string)
+	raw := pod.Annotations[podcontract.RestoredContainerIDsAnnotation]
+	if raw == "" {
+		return ids, nil
+	}
+	if err := json.Unmarshal([]byte(raw), &ids); err != nil {
+		return nil, fmt.Errorf("decode %s annotation: %w", podcontract.RestoredContainerIDsAnnotation, err)
+	}
+	if ids == nil {
+		return nil, fmt.Errorf("decode %s annotation: expected JSON object", podcontract.RestoredContainerIDsAnnotation)
```
Ronkahn21, line 1170: this in backward compatility will always fail

agent/internal/controller/controller.go
galletas1712: Live validation update for 1846173: after repairing the test vCluster, a disposable CPU checkpoint restored into engine-0 and engine-1. I stopped only engine-1 through containerd; kubelet created a replacement, the agent restored it and updated its recorded ID, while engine-0 retained its original ID and zero restarts. Source CPU/filesystem state was verified in both containers after replenishment, and test resources were cleaned up. The environment pytest also passed with the tested operator/agents. The first fixture attempt exposed a stale completion-sentinel issue in the test entrypoint (it exited immediately on restart); the passing run used inert long-lived placeholders. GPU capture/restore was attempted but blocked by another tenant’s DRA reservation, so GPU validation is not claimed.

agent/internal/controller/controller.go
```diff
@@ -681,7 +676,7 @@ func (w *NodeController) restorePodContainers(ctx context.Context, pod *corev1.P
 	for i, mapping := range plan.mappings {
 		i, destination := i, mapping.Destination
 		workers.Go(func() {
-			results[i] = w.restoreDestination(ctx, pod, plan, destination, podKey, recovering)
+			results[i] = w.restoreDestination(ctx, pod, plan, destination, podKey, recovering, replenishing, restoredIDs)
```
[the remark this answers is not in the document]
galletas1712: Addressed in 5306e1c. Replenishment policy is now persisted in the agent-owned `nvidia.com/restore-replenishing: "true"` annotation before writing `RestoreInProgress` or starting workers. It survives pending passes and agent restarts; untracked destinations are skipped before container resolution. Marker-write failure requeues without changing the restore condition, and failed/partially successful restores remain terminal.

Added a regression that reproduced the unsafe sibling replay before the fix and now passes: a tracked replacement stays pending (including an empty container ID), a fresh controller resumes from persisted state, and an untracked sibling without a completion sentinel is never restored. Once the replacement runs, only the tracked destination restores. Additional tests cover marker/status write failures, initial restore behavior, and terminal failures.

Validation: `go test -race ./agent/internal/controller/... -count=5`, `make test`, `make build`, and the full `make check` gate passed. `git merge-tree --write-tree HEAD origin/main` also reports no conflicts against current main. This incremental fix was tested locally; I have not rerun the vcluster tests for this commit.

agent/internal/controller/controller.go
```diff
@@ -1097,13 +1122,128 @@ func (w *NodeController) patchRestoreFinalizers(ctx context.Context, pod *corev1
...
+// persistRestoreReplenishment records the policy before the transient condition
+// changes to InProgress. A later pass or agent restart must not mistake an
+// untracked legacy sibling for a destination awaiting its initial restore.
+// The marker stays set; terminal failure handling still decides retry eligibility.
+func (w *NodeController) persistRestoreReplenishment(ctx context.Context, pod *corev1.Pod) (bool, error) {
+	if value := pod.Annotations[podcontract.RestoreReplenishingAnnotation]; value != "" {
+		if value != "true" {
+			return false, fmt.Errorf("invalid %s annotation: expected true", podcontract.RestoreReplenishingAnnotation)
+		}
+		return true, nil
```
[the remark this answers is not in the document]
galletas1712: Addressed in 71cabab by removing the replenishment annotation rather than adding ownership heuristics. The agent now persists `Restored=False/RestoreReplenishing` before workers start. Pending passes, agent restarts, and preflight waits preserve this phase. A phase-specific compatibility refusal preserves replacement-only policy through the explicit override. Producer metadata cannot select replenishment; a regression verifies that copying the former marker onto a new Pod does not skip initial execution.

The same audit tightened recovery: intent and completion evidence are bound to Pod UID, content UID, destination, and runtime container ID. The shared workload-facing `restore-complete` file alone no longer authorizes recovery. A started operation without completion evidence fails closed instead of replaying CRIU. This intentionally favors safety over automatic recovery in that uncertain crash window; active restores must be drained before upgrading from agents without the new evidence.

Tests cover phase/destination decisions, missing container status, status-write failure, pending passes across fresh controllers, compatibility override, stale prior-incarnation completion, interrupted intent/completion/release writes, and identity binding. `make test`, `make build`, `make check`, and five race-enabled runs of controller/runtime/podcontract tests pass. Merge-tree against current main is clean.

The vcluster read-only environment test also passes. I have not deployed or live-tested this commit: its agents were independently updated to a newer visible-devices image since our prior run, so I left those agents/workloads untouched pending confirmation rather than overwrite concurrent testing.

agent/internal/controller/controller.go
galletas1712: Addressed in d8a578a. `recordRestoredContainerID` now deep-copies the Pod, edits the copy, and derives the narrow annotation patch with `client.MergeFrom(pod).Data(next)`. Only after the API write succeeds do we publish the copied annotations to the worker-shared Pod, under the existing mutex. This removes the duplicated hand-built metadata representation and keeps a failed write from advancing local state.

Added coverage checking that local state is unchanged while the patch is in flight and after failure, existing sibling records are preserved, and an unrelated annotation added independently to the live Pod survives. Five race-enabled controller test runs, `make test`, `make build`, and `make check` pass.

agent/internal/controller/controller.go
galletas1712: Agreed that a cached read here would be unsafe. The current queue path uses the informer only for work keys and performs a direct clientset Pod GET from the API server before terminal/replenishment decisions; it does not use the controller-runtime cached client for those decisions.

In d8a578a I strengthened `TestProcessRestoreQueueItemUsesLivePodState` so the controller-runtime client really contains stale ID records while the direct API client contains the already-restored incarnation. No restore runs. Added a second regression proving an authoritative GET failure does not fall back to cached state or write restore status. Race tests and the full local gate pass. This does not claim protection against two independently active agent processes; the existing per-Pod queue serialization remains the execution model.

agent/internal/controller/controller.go
galletas1712: The absent-annotation legacy case returns an initialized empty map before JSON decoding, so it does not reach this error. This branch rejects an explicit JSON `null`, which is corrupt persisted state rather than an old Pod without tracking metadata.

Added explicit compatibility tests and documentation in d8a578a: nil/missing/empty annotations and `{}` decode successfully; an initial restore without records executes and populates them; an already-successful legacy Pod without records stays successful and is not replayed, including after a container restart. Automatic replenishment is intentionally unavailable for those legacy Pods because we cannot prove which incarnation was restored. Malformed JSON and explicit `null` retain the conservative no-replay behavior.

All local tests, five race-enabled controller runs, build, and `make check` pass. This clarification is separate from the documented requirement to drain in-flight restores before upgrading to incarnation-scoped recovery evidence.

PR #175, opened 2026-09-02
test(e2e): checkpoint/restore tests for the framework guide workloads

### LINKED ISSUE #263: Use stock upstream framework images in guides (deliver app.py via ConfigMap)
## Summary

Switch the vLLM, SGLang, and TensorRT-LLM guides from custom-built images to
the stock upstream framework images. Instead of baking `app.py` into a
`Dockerfile.<framework>`, the guides now deliver `app.py` through a Kubernetes
`ConfigMap` that is mounted into the pod at runtime.

The e2e test infrastructure (added on `main`) drives the guides directly: it
applies the `ConfigMap`, creates source and restore pods using the guide
manifests, and uses the upstream image as-is. A single image tag therefore
covers both the documented user path and the regression test path.

## What changed

- `deployment.yaml` / `restore-deployment.yaml` — replaced the custom image
  reference with the pinned upstream framework image; added a `ConfigMap`
  volume mount for `app.py`
- `app-configmap.yaml` (new) — holds the `app.py` content as a `ConfigMap`
  data key; applied by the e2e test and documented in the guide
- `Dockerfile.<framework>` — removed where its only function was copying
  `app.py`; the upstream image is used directly

## Test plan

- [ ] `e2e-static` CI job passes (framework manifest validation in
  `test_framework_manifests.py`)
- [ ] Manual: `kubectl apply -f docs/guides/vllm/` on a cluster with an
  available GPU produces a running pod with the `app.py` ConfigMap mounted
  correctly
- [ ] The existing GPU e2e suite (`test_frameworks.py`) runs against the guide
  manifests unchanged

### Why is this needed?

Previously, every change to `app.py` or a framework version required building
and pushing a custom image. That image lived outside the repo, so guides could
drift from the image the e2e tests exercised. Moving `app.py` into a
`ConfigMap` means:

- The guide is self-contained: `kubectl apply -f .` covers the full deployment
- The e2e tests exercise the exact same manifests and `app.py` as the guide
- Framework version upgrades are a one-line change to the image tag in the
  manifest, not a Dockerfile rebuild

### Alternatives you considered

_No response_

documentation

### Before submitting

- [x] I have searched the existing issues and found no duplicate of this request

oleg-kushniriov:
Closes https://github.com/ai-dynamo/snapshot/issues/263

## Summary

Switch the vLLM, SGLang, and TensorRT-LLM guides from custom-built images to
the stock upstream framework images. Instead of baking `app.py` into a
`Dockerfile.<framework>`, the guides now deliver `app.py` through a Kubernetes
`ConfigMap` that is mounted into the pod at runtime.

The e2e test infrastructure (already on `main`) drives the guides directly: it
applies the `ConfigMap`, creates source and restore pods using the guide
manifests, and uses the upstream image as-is. A single image tag therefore
covers both the documented user path and the regression test path.

## What changed

- `deployment.yaml` / `restore-deployment.yaml` — replaced the custom image
  reference with the pinned upstream framework image; added a `ConfigMap`
  volume mount for `app.py`
- `app-configmap.yaml` (new) — holds the `app.py` content as a `ConfigMap`
  data key; applied by the e2e test and documented in the guide
- `Dockerfile.<framework>` — removed where its only function was copying
  `app.py`; the upstream image is used directly

## Why

Previously, every change to `app.py` or a framework version required building
and pushing a custom image. That image lived outside the repo, so guides could
drift from the image the e2e tests exercised. Moving `app.py` into a
`ConfigMap` means:

- The guide is self-contained: `kubectl apply -f .` covers the full deployment
- The e2e tests exercise the exact same manifests and `app.py` as the guide
- Framework version upgrades are a one-line change to the image tag in the
  manifest, not a Dockerfile rebuild

## Test plan

- [ ] `e2e-static` CI job passes (framework manifest validation in
  `test_framework_manifests.py`)
- [ ] Manual: `kubectl apply -f docs/guides/vllm/` on a cluster with an
  available GPU produces a running pod with the `app.py` ConfigMap mounted
  correctly
- [ ] The existing GPU e2e suite (`test_frameworks.py`) runs against the guide
  manifests unchanged

## Summary by CodeRabbit

* **Documentation**
  * Updated vLLM, SGLang, and TensorRT-LLM guides to use pinned upstream runtime images with mounted application configuration.
  * Added restore deployment guidance, ConfigMap setup, model selection, and namespace configuration.
  * Simplified deployment and checkpoint instructions with clearer Pod requirements and troubleshooting guidance.
  * Removed legacy custom-image build instructions and outdated framework-specific end-to-end examples.
  * Improved restore diagnostics with progress tracking and clearer failure reporting.
  * Updated links and examples to reflect the current framework guide locations.


docs/guides/tensorrt-llm/app.py
```diff
@@ -5,9 +5,18 @@
...
 from pathlib import Path
 
+# A restore placeholder must stay a minimal, inert process: Snapshot restores
+# the checkpointed tree alongside it, and importing TensorRT-LLM here would map
+# the framework and CUDA libraries into the placeholder for nothing. Decide
+# before importing the framework.
+if os.environ.get("SNAPSHOT_RESTORE_STANDBY") == "1":
+    while True:
+        time.sleep(3600)
+
```
Ronkahn21: when restore we change the commad to sleep infinity

docs/guides/tensorrt-llm/app.py
```diff
@@ -113,9 +118,21 @@ def main() -> None:
...
+                text = generate_text(llm, ["Reply with one word: restored"])[0]
+                progress.write_text("generated\n", encoding="utf-8")
+                print(f"TensorRT-LLM restored output={text!r}", flush=True)
+                serve_api(llm, text)
+            except Exception:
+                CONTROL_DIR.joinpath("trtllm-restore-error").write_text(
+                    traceback.format_exc(),
+                    encoding="utf-8",
+                )
+                raise
```
Ronkahn21, line 127: why? this is not needed, the file approch is not something we want to encourage

docs/guides/vllm/app.py
```diff
@@ -114,16 +122,30 @@ async def main() -> None:
...
+                    "Reply with one word: restored",
+                    "snapshot-restore-check",
+                )
+                progress.write_text("generated\n", encoding="utf-8")
+                print(f"vLLM restored output={text!r}", flush=True)
+                await serve_api(engine, text)
+            except Exception:
+                CONTROL_DIR.joinpath("vllm-restore-error").write_text(
+                    traceback.format_exc(),
+                    encoding="utf-8",
```
Ronkahn21, line 142: same for all

docs/guides/vllm/deployment.yaml
```diff
@@ -28,15 +28,27 @@ spec:
...
               value: Qwen/Qwen3-0.6B
             - name: VLLM_MAX_MODEL_LEN
               value: "2048"
             - name: VLLM_GPU_MEMORY_UTILIZATION
               value: "0.30"
+            # app.py calls AsyncLLM directly rather than vLLM's CLI wrapper,
+            # which sets this automatically when unset. Without it, worker
+            # startup defaults to fork (or switches to spawn only if vLLM
+            # detects CUDA already initialized), which is unreliable across
+            # checkpoint/restore.
```
Ronkahn21: this should be part of doc not here

docs/guides/sglang.md
```diff
@@ -8,35 +8,39 @@ tooling at runtime.
 ## Build
 
 Start with an SGLang image that includes SGLang, CUDA, and
-`torch_memory_saver`. Add one program that prepares SGLang for checkpoint and
-resumes it after restore. Select the model when deploying the source pod.
+`torch_memory_saver` -- unmodified, with one program mounted into it that
+prepares SGLang for checkpoint and resumes it after restore. There is no
+Snapshot-specific image to build or push: `deployment.yaml` pins the exact
+upstream image, and `app.py` is mounted from a ConfigMap. Select the model
+when deploying the source pod.
```
Ronkahn21: this entire guide need redo as it not build anymroe 
lets not update iit like this
cc @ozbarshalom for viz

ozbarshalom (approved):

e2e/snapshot_e2e/infra/setup.py
```diff
@@ -522,6 +522,12 @@ def create_vcluster(namespace: str, name: str, k8s_version: str) -> None:
             },
         },
     }
+    if os.environ.get("SNAPSHOT_E2E_MODEL_CACHE_SERVER"):
```
Ronkahn21: could just pass it as argement instead of using env var

e2e/snapshot_e2e/framework_workloads.py
```diff
@@ -0,0 +1,276 @@
...
+        "apiVersion": "v1",
+        "kind": "PersistentVolumeClaim",
+        "metadata": {"name": cache.pvc_name, "namespace": config.namespace},
+        "spec": {
+            "accessModes": ["ReadWriteMany"],
+            "storageClassName": "",
+            "volumeName": cache.pvc_name,
+            "resources": {"requests": {"storage": SHARED_CACHE_CAPACITY}},
+        },
+    }
```
Ronkahn21: can we have just yaml file tampalte we load instad of using inline, it allow easier change without changeing the script

docs/guides/sglang/Dockerfile.sglang
```diff
@@ -1,23 +0,0 @@
-# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
-# SPDX-License-Identifier: Apache-2.0
```
ozbarshalom, line 2: We would not like to change the current usage guide based on the E2E changes.
Using configmap instead of a custom image is another approach, which we could consider to document in addition to the existing one. please open a dedicated PR for that.

ozbarshalom (changes requested):

danbar2 (dismissed):
LGTM

docs/guides/sglang/Dockerfile.sglang
oleg-kushniriov: Did not we decided to do this change and align with E2E tests?

shayasoolin: @ozbarshalom why not accept this change? We see it as a simplification of the current guide, that requires building an image out of the globally available vllm one (for example). From our experience, it's a heavy and very slow process, plus you then have to maintain those images, update/patch them as new versions are out etc.
I strongly suggest to take this path of adding `app.py` as a side-script mounted into the existing container.

It's not a matter of doing what the e2e does, it's much more user-facing than that. When trying to deploy per the guides we experienced the cons of following the Dockerfile approach.

docs/guides/sglang/app.py
```diff
@@ -61,10 +62,16 @@ def generate_text(engine: Any, prompt: str) -> str:
...
+    # returned). release_memory_occupation() asserts is_fully_idle(), which
+    # also requires last_batch to be empty. The default "in_place" mode freezes
+    # scheduler state untouched, so pausing right after a generation leaves a
+    # stale last_batch and the release fails with "should be called only when
+    # server is idle" (SGLang 0.5.17).
     engine.loop.run_until_complete(
-        engine.tokenizer_manager.pause_generation(
-            PauseGenerationReqInput(mode="abort")
-        )
+        engine.tokenizer_manager.pause_generation(PauseGenerationReqInput(mode="retract"))
```
shayasoolin, line 74: This change, as well as the one around `VLLM_WORKER_MULTIPROC_METHOD` and perhaps other ones - come in this PR together with the `ConfigMap` change. Not sure where it's going but consider splitting so you can push those important changes first.

docs/guides/tensorrt-llm/app.py
```diff
@@ -113,9 +110,21 @@ def main() -> None:
...
         if CONTROL_DIR.joinpath("restore-complete").exists():
-            text = generate_text(llm, ["Reply with one word: restored"])[0]
-            print(f"TensorRT-LLM restored output={text!r}", flush=True)
-            serve_api(llm, text)
+            # The restored process keeps the source container's stdout, which
+            # is gone; a failure here would otherwise be invisible. Record it
+            # in the control directory next to the success sentinel.
+            try:
+                progress = CONTROL_DIR.joinpath("trtllm-restore-progress")
+                text = generate_text(llm, ["Reply with one word: restored"])[0]
```
shayasoolin: Nit - I'd use the same prompt for all engines (e.g. `The capital city of Germany is`). Using the 'restored' term here might be misleading as if it's related to the process or the expected output.

docs/guides/vllm/app.py
```diff
@@ -114,16 +118,30 @@ async def main() -> None:
...
+            try:
+                progress = CONTROL_DIR.joinpath("vllm-restore-progress")
+                await engine.wake_up()
+                progress.write_text("woken\n", encoding="utf-8")
+                await engine.resume_generation()
+                await engine.check_health()
+                progress.write_text("generation-resumed\n", encoding="utf-8")
+                text = await generate_text(
+                    engine,
+                    "Reply with one word: restored",
```
shayasoolin: Same here

docs/guides/sglang.md
```diff
@@ -57,69 +55,44 @@ generation succeeds and the API is listening. To validate the restored replica,
...
+
+kubectl create configmap sglang-app \
+  --namespace "$SNAPSHOT_NAMESPACE" \
+  --from-file=app.py
 ```
 
-In [`deployment.yaml`](sglang/deployment.yaml), replace the example `image` with
-the one pushed in step 2 and select the model through `SNAPSHOT_MODEL`. Both the
-init container and the main container carry each value:
+Re-run this command after editing `app.py` -- `kubectl create configmap` fails
```
shayasoolin: I'd drop this comment, we don't expect the user to modify `app.py` and this feels out of scope for the basic walkthrough.
But if you do keep it, I'd put the full command. The text is currently "re-run this" and then "but it's going to fail so run something different"...

shayasoolin: Also, stick to the exact same text in all 3 flavors.

docs/guides/README.md
```diff
@@ -2,8 +2,9 @@
 
 Using Snapshot is a three-stage flow:
 
-1. **Make the workload snapshot-ready** — the per-framework guides below build
-   one from the framework's runtime image. The requirement they satisfy is the
+1. **Make the workload snapshot-ready** — the per-framework guides below run
+   the framework's stock runtime image unmodified, with a small program
+   mounted into it from a ConfigMap. The requirement they satisfy is the
```
ozbarshalom:
```suggestion
1. **Make the workload snapshot-ready** — the per-framework guides below run the framework's runtime image unmodified and mount an entrypoint into it from a ConfigMap, implementing the
```

docs/guides/README.md
```diff
@@ -19,8 +20,8 @@ deployment in stage 1 differ.
 
 ## 1. Make the workload snapshot-ready
 
-The reference method — build a snapshot-ready image, then deploy it — per
-inference framework:
+Per inference framework, mount a small entrypoint into the framework's stock
+runtime image and deploy it:
```
ozbarshalom:
```suggestion
Per inference framework, mount an entrypoint into the framework's runtime image and deploy it:
```

docs/guides/sglang.md
```diff
@@ -1,44 +1,41 @@
-# Build and deploy an SGLang replica
+# Deploy an SGLang replica
 
-This guide makes an SGLang workload snapshot-ready using the **custom-image
-method**: start from an SGLang runtime image and add a small entrypoint that
-implements Snapshot's [workload contract](../reference/workload-contract.md).
-Building an image is the reference way to package a compliant workload, not a
-requirement of Snapshot — any container whose entrypoint satisfies the contract
-works. The Snapshot agent injects the restore tooling at runtime.
+This guide makes an SGLang workload snapshot-ready by mounting a small
```
ozbarshalom: avoid using the term "small entrypoint" small is subjective

docs/guides/sglang.md
```diff
@@ -1,44 +1,41 @@
...
+entrypoint into an SGLang runtime image, implementing Snapshot's [workload
+contract](../reference/workload-contract.md). The example runs an SGLang
+image that includes SGLang, CUDA, and `torch_memory_saver`, unmodified --
+there is no Snapshot-specific image to build or push. `deployment.yaml` pins
```
ozbarshalom: No need to mention that there is no Snapshot-specific image

docs/guides/tensorrt-llm.md
```diff
@@ -1,42 +1,40 @@
...
-
-This guide makes a TensorRT-LLM workload snapshot-ready using the **custom-image
-method**: start from the TensorRT-LLM runtime image and add a small entrypoint
-that implements Snapshot's [workload contract](../reference/workload-contract.md).
-Building an image is the reference way to package a compliant workload, not a
-requirement of Snapshot — any container whose entrypoint satisfies the contract
-works. The Snapshot agent injects the restore tooling at runtime.
+# Deploy a TensorRT-LLM replica
+
+This guide makes a TensorRT-LLM workload snapshot-ready by mounting a small
```
ozbarshalom: same comments as in SGLang

docs/guides/vllm.md
```diff
@@ -1,44 +1,39 @@
-# Build and deploy a vLLM replica
+# Deploy a vLLM replica
 
-This guide makes a vLLM workload snapshot-ready using the **custom-image
-method**: start from vLLM's runtime image and add a small entrypoint that
-implements Snapshot's [workload contract](../reference/workload-contract.md).
-Building an image is the reference way to package a compliant workload, not a
-requirement of Snapshot — any container whose entrypoint satisfies the contract
-works. The Snapshot agent injects the restore tooling at runtime.
+This guide makes a vLLM workload snapshot-ready by mounting a small entrypoint
```
ozbarshalom: same comments as in SGLang

ozbarshalom (changes requested):

ozbarshalom (approved):

danbar2 (approved):

PR #176, opened 2026-09-02
ci(e2e): run framework checkpoint/restore tests on pull requests and nightly

### LINKED ISSUE #240: Compatibility matrix for vLLM, TRT-LLM and SGLang on nightlies
## Problem

The framework guides under `docs/guides/` (vLLM, SGLang, TensorRT-LLM) are the primary way users adopt Snapshot, but nothing in CI exercises them. Whether a given Snapshot build still checkpoints and restores each framework is only discovered by hand, usually after a framework image bump or an agent change breaks one of them.

## Goal

A nightly compatibility matrix that checkpoints and restores each framework guide workload on a GPU runner and reports a per-framework pass/fail, so regressions in either Snapshot or the pinned framework images surface within a day.

## Scope

- Run the checkpoint/restore e2e for `vllm`, `sglang`, and `tensorrt-llm` as a matrix, one vCluster per framework.
- Trigger nightly, on `workflow_dispatch` with a framework selector, and on `copy-pr-bot` mirror branches for PRs that touch `e2e/**`, `docs/guides/**`, or `hack/**`.
- On every PR, run the cluster-free manifest and workload checks so the guide manifests cannot drift silently.
- On failure, upload diagnostics (cluster YAML, describes, events, agent/operator/workload logs, agent `dmesg`).
- Fall back to `main`'s Snapshot images on branches that have no image of their own, with a visible warning.

## Non-goals

- Building operator/agent images for mirror branches. Until that exists, agent and operator changes are covered nightly only.
- Fixing framework-specific restore defects found by the matrix. Those are tracked separately; the gate stays red when they occur.

## Implementation

- #175 adds the framework checkpoint/restore tests and the guide manifest fixes they needed.
- #176 wires them into CI: `e2e-frameworks.yaml` (GPU matrix), the `e2e-static` job in `ci.yml`, `copy-pr-bot.yaml`, and the `hack/` helpers for image resolution and Datadog GPU monitoring.

## Done when

- The nightly run publishes a green/red result per framework.
- A PR touching the guides or e2e harness gets a matrix run on its mirror branch.
- The cluster-free checks run on every PR.

oleg-kushniriov:
### Summary

Fixes #240

Wires the framework checkpoint/restore tests from #175 into CI: a per-framework GPU matrix on copy-pr-bot mirror branches, nightly, and dispatch; a cluster-free job on every PR; and the helper scripts the GPU job needs. This PR holds only `.github/` and `hack/`.

### Changes

- **`.github/workflows/e2e-frameworks.yaml`**: matrix `vllm | sglang | tensorrt-llm`, `runs-on: prod-deploy-tester-v1`, one vCluster per framework, `SNAPSHOT_E2E_PVC_SIZE=64Gi`. Triggers: `push` to `pull-request/[0-9]+` filtered to `e2e/**`, `docs/guides/**`, `hack/**`, and the workflow itself (the GPU runners reject `pull_request` events); nightly; `workflow_dispatch` with `framework`, `snapshot_tag`, and `datadog_gpu_monitoring`. Steps mirror `e2e.yaml` plus the shared model cache variables, the test, and on failure a diagnostics artifact (cluster YAML, describes, events, agent/operator/workload logs, agent `dmesg`). Job timeout is derived from the test budgets documented next to `SOURCE_READY_TIMEOUT_SECONDS`; concurrency cancels superseded runs per ref.
- **`.github/copy-pr-bot.yaml`** (`enabled: true`). Cannot be verified from a branch since the bot reads its config from the default branch; the first PR after merge must produce a `pull-request/<n>` branch and a run.
- **`.github/workflows/ci.yml`**: `e2e-static` job running the cluster-free framework manifest and workload script checks on every PR.
- **`hack/resolve-snapshot-image-tag.py --fallback-main`**: on any ref other than `main`/`release/*`, resolve head tag → branch-slugged tag → newest plain `v0.0.0-g<sha8>` common to operator and agent, with a `::warning::` and a step-summary note. Listings are paged to exhaustion so accumulated tags cannot push a valid one out of the window. Strict default unchanged.
- **`hack/datadog-gpu-monitoring.py`**: turns the host cluster's Datadog GPU monitoring off for the run and back on afterwards if it was on. Dispatch input `datadog_gpu_monitoring=leave` skips it.
- **Shared model cache**: workflow points at the CI cluster's NFS export via `SNAPSHOT_E2E_MODEL_CACHE_SERVER/PATH`; repository variables take precedence over the defaults, and a step reports which is in use. Defaults stay as literals until repository variables can be created (`admin: false`).

### GPU results

Latest, on the tree at the top of this stack: [34030830624](https://github.com/ai-dynamo/snapshot/actions/runs/34030830624), all three frameworks green (vLLM 7 min, SGLang 7 min, TensorRT-LLM 9 min).

History while the guide fixes in #175 were landing, all `workflow_dispatch` with fallback to `main`'s Snapshot images:

| Run | vLLM | SGLang | TensorRT-LLM |
|---|---|---|---|
| [33619417391](https://github.com/ai-dynamo/snapshot/actions/runs/33619417391) (pre-cache) | ✗ DNS/timeout | ✗ DNS | ✗ DNS |
| [33621269608](https://github.com/ai-dynamo/snapshot/actions/runs/33621269608) (cache) | ✗ CRIU segfault on restore | ✗ pause not idle | ✅ |
| [33622441930](https://github.com/ai-dynamo/snapshot/actions/runs/33622441930) (vLLM only) | ✅ | – | – |
| [33623640483](https://github.com/ai-dynamo/snapshot/actions/runs/33623640483) (`retract`) | ✗ CRIU segfault | ✗ hang after restore | ✅ |
| [33626331376](https://github.com/ai-dynamo/snapshot/actions/runs/33626331376) (error sentinel) | ✅ | ✗ hang after restore | ✅ |
| [33629204404](https://github.com/ai-dynamo/snapshot/actions/runs/33629204404) (progress markers) | ✗ CRIU segfault | ✅ | ✅ |
| [34030830624](https://github.com/ai-dynamo/snapshot/actions/runs/34030830624) (current tree) | ✅ | ✅ | ✅ |

The intermittent vLLM CRIU segfault is an agent-side defect and is documented in #175's known issues; the gate is intentionally kept red when it occurs.

### Verification

- Workflow YAML parses; resolver paths exercised against a fake GHCR listing.
- Run 34030830624 above, dispatched on this branch.

### Review follow-ups

- **PR path filter**: `agent/**`, `operator/**`, `api/**`, `charts/**` removed. A mirror-branch run tests `main`'s Snapshot images and cannot validate those changes; they are covered nightly. Follow-up, not here: build operator/agent images for mirror branches, then drop the fallback.
- **Timeouts**: job timeout recomputed after the image-resolution wait was removed; the relationship to the test budgets is documented in `frameworks.py`.
- **Cache defaults**: kept as a fallback because neither author nor reviewer can create repository variables; remove the literals once the variables exist.

## Summary by CodeRabbit

- **New Features**
  - Added end-to-end checkpoint and restore coverage for vLLM, SGLang, and TensorRT-LLM workloads.
  - Added framework workload configurations, model caching support, restore validation, and inference checks.
  - Added automated diagnostics collection and GPU monitoring management for test runs.
  - Added flexible snapshot image resolution with branch and fallback tag support.

- **Documentation**
  - Documented framework test selection, image configuration, model caches, and image resolution behavior.

- **Tests**
  - Added cluster-free manifest validation and expanded continuous integration coverage.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,3 +59,23 @@ jobs:
             **/go.sum
             go.work.sum
       - run: make check
+
+  # Cluster-free e2e checks: the workload scripts and the framework guide
+  # manifests against the restore-pod contract. Seconds to run, so they gate
+  # every pull request rather than only the GPU-gated e2e workflows.
+  e2e-static:
+    runs-on: ubuntu-latest
+    env:
+      UV_LINK_MODE: copy
+    steps:
+      - uses: actions/checkout@v7
+      - uses: astral-sh/setup-uv@c771a70e6277c0a99b617c7a806ffedaca235ff9 # v9.0.0
+        with:
+          version: "0.10.11"
+          checksum: "5a360b0de092ddf4131f5313d0411b48c4e95e8107e40c3f8f2e9fcb636b3583"
+          enable-cache: "false"
+          download-from-astral-mirror: "false"
+      - run: |
+          uv run --locked --project e2e pytest \
+            e2e/tests/test_framework_manifests.py \
+            e2e/tests/test_workload_scripts.py -vv
--- a/e2e/snapshot_e2e/inference.py
+++ b/e2e/snapshot_e2e/inference.py
@@ -1,0 +1,101 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Inference requests against a framework guide program, from inside its pod.
+
+The guide programs expose `POST /generate` with `{"prompt": ...}` on port 8000
+and answer `{"text": ...}`. There is no Service in front of them, so the
+request is issued from inside the pod through exec. Python's standard library
+is used rather than curl because every framework runtime image ships Python
+and none is guaranteed to ship curl.
+"""
+
+from __future__ import annotations
+
+import json
+import shlex
+
+from snapshot_e2e import k8s
+from snapshot_e2e.frameworks import API_PORT
+from snapshot_e2e.frameworks import REQUEST_TIMEOUT_SECONDS
+
+_CLIENT = """
+import json, sys, urllib.request
+payload = json.dumps({"prompt": sys.argv[1]}).encode()
+request = urllib.request.Request(
+    sys.argv[2],
+    data=payload,
+    headers={"Content-Type": "application/json"},
+    method="POST",
+)
+with urllib.request.urlopen(request, timeout=float(sys.argv[3])) as response:
+    body = response.read().decode()
+print("__snapshot_e2e_generate__" + body)
+"""
+
+_MARKER = "__snapshot_e2e_generate__"
+
+
+def request_generate(
+    namespace: str,
+    pod: str,
+    prompt: str,
+    *,
+    timeout: int = REQUEST_TIMEOUT_SECONDS,
+    port: int = API_PORT,
+    container: str | None = None,
+) -> str:
+    """POSTs the prompt to the program's /generate and returns the generated text.
+
+    Raises AssertionError if the endpoint does not answer with non-empty text;
+    the raw exec output is included so a framework error is visible verbatim.
+    """
+    url = f"http://127.0.0.1:{port}/generate"
+    command = (
+        f"python3 -c {shlex.quote(_CLIENT)} {shlex.quote(prompt)} "
+        f"{shlex.quote(url)} {timeout}"
+    )
+    output = k8s.exec_command(namespace, pod, command, container=container)
+    marker_at = output.rfind(_MARKER)
+    if marker_at < 0:
+        raise AssertionError(
+            f"/generate on {namespace}/{pod} produced no response; exec output:\n{output}"
+        )
+    body = output[marker_at + len(_MARKER):].strip()
+    try:
+        text = json.loads(body)["text"]
+    except (ValueError, KeyError, TypeError) as exc:
+        raise AssertionError(
+            f"/generate on {namespace}/{pod} returned malformed body {body!r}: {exc}"
+        ) from exc
+    if not isinstance(text, str) or not text.strip():
+        raise AssertionError(f"/generate on {namespace}/{pod} returned empty text: {body!r}")
+    return text
+
+
+_FILE_MARKER = "__snapshot_e2e_file__"
+
+
+def read_control_file(
+    namespace: str, pod: str, path: str, *, container: str | None = None
+) -> str | None:
+    """Content of a control-directory file, or None when it does not exist.
+
+    exec merges stdout and stderr and hides the remote exit status, so a bare
+    `cat` of a missing file would come back as its error text and pass a
+    "non-empty" assertion. The marker is printed only when the file exists.
+    """
+    quoted = shlex.quote(path)
+    output = k8s.exec_command(
+        namespace,
+        pod,
+        f"if [[ -f {quoted} ]]; then printf '%s' {shlex.quote(_FILE_MARKER)}; cat {quoted}; fi",
+        container=container,
+    )
+    return parse_control_file(output)
+
+
+def parse_control_file(output: str) -> str | None:
+    if not output.startswith(_FILE_MARKER):
+        return None
+    return output[len(_FILE_MARKER):]
--- a/e2e/snapshot_e2e/inference.py
+++ b/e2e/snapshot_e2e/inference.py
@@ -71,31 +71,3 @@ def request_generate(
     if not isinstance(text, str) or not text.strip():
         raise AssertionError(f"/generate on {namespace}/{pod} returned empty text: {body!r}")
     return text
-
-
-_FILE_MARKER = "__snapshot_e2e_file__"
-
-
-def read_control_file(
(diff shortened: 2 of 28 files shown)


.github/workflows/ci.yml
```diff
@@ -59,3 +59,23 @@ jobs:
             **/go.sum
             go.work.sum
       - run: make check
+
+  # Cluster-free e2e checks: the workload scripts and the framework guide
+  # manifests against the restore-pod contract. Seconds to run, so they gate
+  # every pull request rather than only the GPU-gated e2e workflows.
```
Ronkahn21: cluster free and e2e does not go with one another 
does it just some integration

e2e/snapshot_e2e/lifecycle.py
```diff
@@ -216,6 +217,60 @@ def detail() -> str:
...
+            last_error = f"{type(exc).__name__}: {exc}"
+            return None
+        if output.startswith(f"{marker}:error"):
+            body = output.split("\n", 1)[1] if "\n" in output else ""
+            raise AssertionError(
+                f"restored program in {namespace}/{pod} failed after restore "
+                f"({error_file}):\n{body}"
+            )
+        if output.startswith(f"{marker}:ready"):
+            return output.split("\n", 1)[1] if "\n" in output else ""
```
danbar2: `exec_command` runs `/bin/bash -lc` with stderr merged into stdout, so anything a profile script or a driver warning prints lands in front of the marker and both of these `startswith` checks miss it. `outcome()` then returns `None` forever and you get a timeout that says "neither sentinel observed yet" — even in the case where the error file *was* read and the traceback was sitting in `output` the whole time. Which is the one failure mode this function exists to prevent.

`request_generate` over in `inference.py` already uses `rfind` for exactly this reason, so it's mostly a consistency thing:

```suggestion
        marker_at = output.rfind(marker)
        if marker_at < 0:
            return None
        tail = output[marker_at:]
        body = tail.split("\n", 1)[1] if "\n" in tail else ""
        if tail.startswith(f"{marker}:error"):
            raise AssertionError(
                f"restored program in {namespace}/{pod} failed after restore "
                f"({error_file}):\n{body}"
            )
        if tail.startswith(f"{marker}:ready"):
            return body
```

`wait_for_file` above has the same issue and then some — line 209 is `response == marker`, so even a trailing newline breaks it. That one predates this PR though, so I'll leave it to you whether it belongs here.

Not urgent, all three frameworks are green today so clearly these images' login shells are quiet. It's the image bump six months from now that worries me.

e2e/snapshot_e2e/manifests/frameworks/sglang/restore-deployment.yaml
```diff
@@ -0,0 +1,134 @@
...
+          imagePullPolicy: IfNotPresent
+          # Inert placeholder: the agent restores the checkpointed process into
+          # this container, so app.py must not run here.
+          command:
+            - /bin/sh
+            - -c
+            - exec sleep infinity
+          args:
+            - --mode
+            - snapshot
```
danbar2: Pretty sure these three rode along from `deployment.yaml` — that's where `--mode snapshot` belongs, on the source pod that actually runs `app.py`. vLLM and TensorRT-LLM's restore manifests don't have them, and neither does `docs/guides/sglang/restore-deployment.yaml`, so it's new here rather than inherited.

Inert as written — k8s just appends args to command, so you get `/bin/sh -c 'exec sleep infinity' --mode snapshot` and `--mode` ends up as `$0` for a script that never looks at it. But it reads like the placeholder is meant to run in snapshot mode, which is the exact opposite of what the comment two lines up says, and it becomes live the moment someone puts `app.py` back in `command`.

```suggestion
```

Somewhat related: `test_restore_pod_is_standby_placeholder_for_this_run` only asserts on `main["command"]`. An `assert "args" not in main` alongside it would've caught this and would stop it coming back.

e2e/snapshot_e2e/inference.py
```diff
@@ -0,0 +1,101 @@
...
+        f"if [[ -f {quoted} ]]; then printf '%s' {shlex.quote(_FILE_MARKER)}; cat {quoted}; fi",
+        container=container,
+    )
+    return parse_control_file(output)
+
+
+def parse_control_file(output: str) -> str | None:
+    if not output.startswith(_FILE_MARKER):
+        return None
+    return output[len(_FILE_MARKER):]
```
danbar2: These two don't have any callers left. `test_frameworks.py` was calling `read_control_file` in two places as of 052b159; moving to `wait_for_restore_outcome` orphaned them both.

I wouldn't normally flag 25 dead lines, but this is the version with the problems the newer function fixed — `parse_control_file` does `startswith(_FILE_MARKER)`, there's no error-file handling, and no polling. Good docstring, obvious name, sitting in a module called `inference`. The next person who needs to read a control file is going to land on this well before they find `wait_for_restore_outcome` buried in `lifecycle.py`.

```suggestion
```

If you'd rather keep a shared parser that's fine too — fix `parse_control_file` to use `rfind` and have `wait_for_restore_outcome` call it, so there's one parser to keep correct instead of three.

danbar2 (commented):
Three notes from a closer read of the e2e harness. None of them are blockers for the GPU results — suggestions attached where the fix is mechanical.

danbar2 (approved):

PR #177, opened 2026-09-02
feat/framework e2e model cache

oleg-kushniriov:
> Stacked on #176 → #175 → #173 → #171 → #170 (stack #172). Review only the last two commits until the lower PRs merge.

### Summary

Fixes from the first GPU run of the framework tests (`workflow_dispatch` run [33619417391](https://github.com/ai-dynamo/snapshot/actions/runs/33619417391)), which got all three frameworks through vCluster setup, image resolution, environment validation, and into the test, then failed in three different ways.

### What the run showed

| Framework | Observation | Cause |
|---|---|---|
| TensorRT-LLM | source pod exited 1: `Temporary failure in name resolution` for `huggingface.co` ~25s after start | DNS in a freshly created vCluster is not immediately usable; the other two pods resolved HF a minute later |
| vLLM | source pod still initializing when the 300s budget expired | ~4 min image pull (10–20 GB) + model download + engine init + `torch.compile`/CUDA-graph capture |
| SGLang | `AssertionError: release_memory_occupation should be called only when server is idle` | the guide pauses with `mode="abort"`; SGLang 0.5.17 does not leave the scheduler idle in that mode |

### Changes

- **Shared model cache** (`frameworks.SharedModelCache`, `framework_workloads.use_shared_model_cache`, `shared_model_cache_volume`, `lifecycle.ensure_pv`): with `SNAPSHOT_E2E_MODEL_CACHE_SERVER/PATH` set, the test creates a static NFS PV/PVC (Retain, `storageClassName: ""`, NFS 4.1 mount options), mounts it at `/models` on every framework container with `HF_HOME=/models` and `HF_HUB_OFFLINE=1`, and drops the guide's download init container / cache volume. The vCluster setup enables `sync.toHost.persistentVolumes` so the mount options reach the node. Without the variables the guides' download path is unchanged.
  - The workflow points at the CI cluster's shared export; repository variables `SNAPSHOT_E2E_MODEL_CACHE_SERVER/PATH` take precedence over the defaults (an admin can set them and drop the literals).
- **Source readiness budget** 300s → 900s (`frameworks.SOURCE_READY_TIMEOUT_SECONDS`), with the observed numbers in the comment.
- **SGLang guide**: `PauseGenerationReqInput()` (default mode) instead of `mode="abort"` before `release_memory_occupation()`. The preceding generation has completed, so there is nothing to abort. This changes the SGLang image digest, so `E2E Framework Images` rebuilds it on this PR.
- Cluster-free tests for the cache rewrite (init container dropped, single `model-cache` volume, mount + env, contract pieces untouched), the PV/PVC shape, and env parsing. README documents both weight sources.

### Verification

- `pytest e2e/tests/test_framework_manifests.py e2e/tests/test_workload_scripts.py` — 30 passed.
- Workflow YAML parses; `setup.py` compiles.
- GPU run pending: needs the rebuilt SGLang image from this PR's `E2E Framework Images` run, then a `workflow_dispatch` on this branch.

### DIFF
--- a/.github/workflows/e2e-frameworks.yaml
+++ b/.github/workflows/e2e-frameworks.yaml
@@ -81,6 +81,12 @@ jobs:
       SNAPSHOT_E2E_READY_TIMEOUT_SECONDS: "480"
       SNAPSHOT_E2E_RESULT_FILE: ${{ github.workspace }}/.snapshot-e2e-setup-result.json
       SNAPSHOT_E2E_FRAMEWORK: ${{ matrix.framework }}
+      # Shared CI model cache (Hugging Face cache layout on an NFS export):
+      # pods in this cluster cannot resolve huggingface.co, and downloading
+      # weights per run would cost minutes anyway. Repository variables take
+      # precedence; the defaults are the CI cluster's export.
+      SNAPSHOT_E2E_MODEL_CACHE_SERVER: ${{ vars.SNAPSHOT_E2E_MODEL_CACHE_SERVER || 'dynamoinfracimodelcache.file.core.windows.net' }}
+      SNAPSHOT_E2E_MODEL_CACHE_PATH: ${{ vars.SNAPSHOT_E2E_MODEL_CACHE_PATH || '/dynamoinfracimodelcache/models' }}
       PYTHONUNBUFFERED: "1"
       UV_LINK_MODE: copy
     steps:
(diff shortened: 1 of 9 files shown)


oleg-kushniriov:
Folded into #176 so the CI PR itself passes e2e; no separate layer needed.

PR #179, opened 2026-09-02
fix(agent): build criu and restore tar against an older glibc floor

oleg-kushniriov:
### Problem

Both `criu` and the restore-time `tar` are exec'd inside the destination placeholder's mount namespace, so their dynamic-linker requirements are bounded by that container's glibc, not the agent image's. Building either against the agent's own `ubuntu:24.04`-based stage (`GLIBC_2.38`) makes restore fail on any destination image with an older glibc, with an exec-time `GLIBC_X.XX not found` error.

### Approaches tried

**Static linking (abandoned)** — the commit history on this branch shows a first attempt at statically linking `criu` against musl. That solved the GLIBC exec crash, but a fully static binary cannot `dlopen()` at all — CRIU treats a configured plugin as required, not optional, so restore aborted trying to load `snapshot_inet_remap.so` (needed for restoring into a pod with a different IP than the source). Confirmed against the real cluster: `restore.log` ends at `Unable to load .../snapshot_inet_remap.so: Dynamic loading not supported` with no further progress. Static linking and criu's plugin architecture are fundamentally incompatible, and static linking also raised open licensing questions for bundling a statically-linked GNU tar. Both are set aside in favor of the fix below.

### Fix

Build `criu` on `ubuntu:22.04` instead of the agent's own `ubuntu:24.04`-based image, keeping normal dynamic linking. This narrows the failure window (destination needs glibc ≥ 2.35 instead of ≥ 2.38) rather than eliminating it, but keeps `dlopen`-based plugin loading intact.

The restore-time `tar` (`runtime.ApplyRootfsDiff`) has the identical glibc-floor problem, since it's exec'd inside the same destination namespace. It's now bundled from the same `criu-builder` stage (dynamically linked, no static/musl build), so both binaries share one glibc floor and one shared-library closure.

Two build issues surfaced switching base images:
- `make install-lib` installs Python bindings via `pip`; added `python3-pip` since 22.04's minimal image doesn't ship it.
- Ubuntu 22.04's git (GnuTLS-backed libcurl) mishandles GitHub's HTTP/2 smart-http response framing, failing with a misleading `could not read Username` credential prompt. Forced `http.version=HTTP/1.1` on the criu clone to route around it.

### Verification

Verified locally (native arch, `criu-builder` target):
- Stage builds cleanly, including the CUDA and inet-remap plugins.
- `criu` is dynamically linked with a `GLIBC_2.35` floor (down from `2.38`); bundled `tar` matches.
- `criu --version` runs successfully; inet-remap plugin `.so` is present and normally glibc-linked.
- Bundled `tar`'s dependency closure resolves correctly against `ubuntu:22.04`; functional smoke test (extract with `--skip-old-files`, `--xattrs`) passes.

End-to-end restore against the real cluster in progress.

Signed-off-by: Oleg Kushniriov <okushniriov@nvidia.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved checkpoint restoration compatibility across supported environments.
  * Restore operations now use the bundled archive utility for more reliable extraction.
  * Preserved supported file metadata while excluding incompatible extended attributes.
  * Added validation to reject invalid archive utility paths and provide clearer failure details.

* **Tests**
  * Added coverage for archive utility selection, path validation, metadata filtering, and archive extraction compatibility.


Ronkahn21:
Could we do the same to solve the tar issue as well ?

oleg-kushniriov:
not sure

agent/Dockerfile
```diff
@@ -166,28 +178,47 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
     && make DESTDIR=/criu-install install-criu install-lib install-cuda_plugin
 
+# criu's own shared-library dependencies (libbsd, libgnutls, libnftables,
+# libnl-3, libnet, libuuid, and their transitive deps) must be resolved here,
+# against this stage's older glibc -- not later in the agent stage, whose ldd
+# would resolve the same library names against its own (newer) filesystem and
+# silently reintroduce the GLIBC_2.38 floor this stage exists to avoid. The
+# agent stage's closure step is no-clobber, so it fills in gaps (tar, ip,
+# cuda-checkpoint) without overwriting what's collected here.
+RUN set -eu; \
```
Ronkahn21, line 185: could we just set the ubuntu version to ubuntu 22.04

oleg-kushniriov: The agent's own base image is NVIDIA's CUDA 13.0 devel image, only published for Ubuntu 24.04 - retargeting the whole thing to 22.04 would mean an older/different CUDA toolkit, risking breakage for cuda-helper-builder and CUDA compatibility (including on newer GPUs like B200).

It also wouldn't help: the GLIBC floor is set by whatever runs inside the destination container at restore time, not by the agent's own image. So we scope the 22.04 downgrade to just the criu-builder stage - only criu/tar get the lower floor, everything else stays on the current CUDA base.

Ronkahn21 (changes requested):
dont merge need legal stuff

agent/Dockerfile
```diff
@@ -243,24 +285,35 @@ COPY --from=builder /snapshot-agent /usr/local/bin/snapshot-agent
...
+# PATH), so it must be the criu-builder-sourced copy above, matching criu's
+# own glibc floor -- not this (newer) stage's own tar.
+COPY --from=criu-builder /criu-install/usr/local/bin/tar /snapshot-binaries/tar
+
+# ip is used for address/route restore during restore (criu/net.c) and
+# resolves through PATH inside the namespace. Network lock is set to SKIP
+# (--network-lock skip) in BuildRestoreOpts, so criu never forks
+# iptables-restore/ip6tables-restore on this path and neither binary needs to
+# be in the bundle. See BuildRestoreOpts for why SKIP is safe for restore
+# specifically.
```
dfeigin-nv, line 306: nit: the "network lock is SKIP so ip isn't needed" reasoning doesn't hold up — network-lock just gates TCP packet-drop locking during dump/restore, it's unrelated to whether criu restores links/addrs/routes. What actually skips that is the netns being external (net.c gates restore_ifaddr/restore_route/etc on `!ns->ext_key`). Conclusion's still right, just wrong cause — might be worth fixing before someone copies this reasoning elsewhere.

dfeigin-nv:
side note, unrelated to this diff specifically: the existing post-condition check further down still ldd's /snapshot-binaries/ip, but per the comment a few lines above (and the placeholder-stage one), ip is never exec'd on this restore path since netns is always external. not from this PR so not blocking, just flagging since I was in the area.

PR #212, opened 2026-09-04
feat(agent): package the cuinterpose shim and coordinator

galletas1712:
## Summary

This is the first layer of the eleven-PR [cuinterpose stack #293](https://github.com/ai-dynamo/snapshot/stack/293): #212 → #213 → #222 → #223 → #215 → #216 → #217 → #292 → #218 → #219 → #220. Closed PR #214 is intentionally not part of the active stack.

Cuinterpose lets Snapshot checkpoint and restore same-node CUDA memory shared between processes, including CUDA multicast objects. This PR establishes only the build, packaging, and shared wire contract. The shim and coordinator are placeholders at this layer.

`agent/cmd/cuinterpose/Makefile` builds `libcuinterpose.so` without linking CUDA, builds a static `cuinterpose-coordinator`, enforces hidden visibility and the glibc 2.34 ceiling, and discovers the fake-driver test suites added by later layers. The CUDA helper image installs CUDA 13.1 headers and packages both binaries under `/snapshot-binaries/snapshot-cuda` for restore.

`protocol.h` defines the fixed header and state-record layouts, participant and allocation identity, diagnostics, and the final operation vocabulary:

```text
HANDSHAKE → INSPECT
capture: PREPARE_MULTICAST → SAVE_ALLOCATIONS → PREPARE_UNICAST
restore: LOAD_ALLOCATIONS → RESTORE_UNICAST → RESTORE_MULTICAST_*
```

Allocation content is represented generically by `ALLOCATION_CONTENT` and `CONTENT_ONLY`; host carriers are an implementation behind the save/load operations in #292 and #218. The protocol also reports live raw imports and unsupported exportable CUDA resource creations so #216 can refuse checkpoint before destructive work.

The bounded Unix-socket helpers support fixed messages, optional records, and one `SCM_RIGHTS` descriptor. They suppress `SIGPIPE`, reject malformed ancillary data, and close received descriptors on every error path. Generic arrays, cleanup, identity, I/O, parsing/randomness, tables/ranges, and timing are centralized under `agent/cmd/cuinterpose/util/`; feature modules keep domain-specific logic.

## Stack boundary

Based on `main`. The next PR, #213, delivers `cuda-checkpoint` and the shim into workload containers. This layer does not inject the shim, discover participants, intercept CUDA, or execute lifecycle operations.

## Validation

[...]

PR #213, opened 2026-09-04
feat: deliver CUDA tools and launch multi-GPU sources

galletas1712:
## Summary

Layer 2 of the eleven-PR [cuinterpose stack #293](https://github.com/ai-dynamo/snapshot/stack/293). Closed PR #214 is intentionally not in the active stack.

This PR applies the CUDA launch-job requirement where Snapshot shapes source Pods and makes the required binaries available at stable paths. It does not opt a workload into cuinterpose.

`ShapeCUDATools` creates one `snapshot-cuda` `emptyDir`, runs an init container from the configured Snapshot agent image, copies `cuda-checkpoint` and `libcuinterpose.so`, and mounts both tools read-only at `/tmp/snapshot-cuda` in every target container. Tool delivery is unconditional for Snapshot targets and is independent of cuinterpose opt-in.

Only a target that may use more than one GPU has its command rewritten under:

```text
/tmp/snapshot-cuda/cuda-checkpoint --launch-job ...
```

The CUDA job file is persisted in Snapshot's control volume. The operator counts `nvidia.com/gpu` resources and referenced DRA claims; indeterminate claim sizes are conservatively treated as multi-GPU because unnecessary wrapping is harmless while missing it causes a late capture failure. Existing source Jobs are verified for universal tool delivery and, where required, the launch wrapper before adoption.

Delivery and interposition remain separate. The mounted shim is inert until #222 puts it in `LD_PRELOAD`; opting into cuinterpose does not change the GPU-count-based launch rule.

The `snapshot-control` volume, `/snapshot-control` mount, and canonical and legacy control-directory variables are reserved. If a workload predefines them, source shaping accepts only one `emptyDir`, one exact writable `subPath=<container-name>` mount, and literal `/snapshot-control` variable values. PVC, `hostPath`, alternate-volume, broad-mount, read-only, duplicate, and `valueFrom` collisions are rejected atomically. This matches restore-side fail-closed validation and prevents the coordinator socket directory from becoming cross-Pod storage.

At capture, the manifest records `cudaTools.delivered`. Restore recreates `/tmp/snapshot-cuda` before CRIU by using the fixed `ns-bind-mount` role, allowing CRIU to reopen the launch wrapper and shim at their original file-backed paths.

## Stack boundary

Based on #212. #222 adds explicit Pod opt-in and preloads the shim. This PR does not create control sockets, run the coordinator, or track CUDA state.

## Validation

On the final stack, `make test` passes in `api`, `agent`, and `operator`. The source-shaping tests cover the exact reserved-volume contract and atomic rejection of unsafe collisions. The pinned CUDA 13.1 builder compiles the production binaries and passes all cuinterpose fake-driver suites. The final published stack head `0eae9c4f9b66093332f00a332afaf56048e1e64a` passed the full `make check` gate, the CUDA 13.1 production build, and all 73 sanitizer-backed native cuinterpose tests. Its physical-GPU suite passed all 3 tests with no skips on two DRA-assigned NVIDIA B200 GPUs. The unicast test kept an explicit allocation-ID ticket-backed peer mapping per worker live across capture and restore; the multicast test required a nonzero multicast VA and shim-logical handle, exercised `BindAddr`, and passed its collective and captured-graph replay. Detailed hardware evidence and measurements are in #220.

### DIFF
--- a/agent/cmd/ns-bind-mount/main.c
+++ b/agent/cmd/ns-bind-mount/main.c
@@ -2,15 +2,17 @@
  * SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  *
- * ns-bind-mount installs and removes the two mounts used by restore:
+ * ns-bind-mount installs and removes the fixed mounts used by restore:
  *
  *   mount-bundle-fd <namespace-fd>
+ *   mount-snapshot-cuda-fd <namespace-fd>
  *   mount-checkpoint-fd <namespace-fd> <checkpoint-path>
  *   unmount-bundle-fd <namespace-fd> [created]
+ *   unmount-snapshot-cuda-fd <namespace-fd> [created]
  *   unmount-checkpoint-fd <namespace-fd> [created]
  *
  * The caller pins the target mount namespace and passes its descriptor through
- * ExtraFiles. Bundle and checkpoint policy is deliberately fixed here: callers
+ * ExtraFiles. Bundle, snapshot-cuda, and checkpoint policy is deliberately fixed here: callers
  * cannot select arbitrary host sources, container destinations, or attributes.
  */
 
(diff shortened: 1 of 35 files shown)


galletas1712:

PR #214, opened 2026-09-04
feat(agent): orchestrate the cuinterpose coordinator around checkpoint and restore

galletas1712:
## Summary

Part of the **cuinterpose** stack: the CUDA interposer that lets Snapshot checkpoint and restore CUDA memory shared between processes (tensor-parallel workers, NCCL, FlashInfer, PyTorch symmetric memory) and CUDA multicast objects. Nine PRs, each one component: packaging, delivery, agent orchestration, then the shim and coordinator layer by layer, then the GPU tests. The design document is held back until the code has been reviewed.

This PR is the agent's side of the contract. What to review: `agent/internal/cuda/cuinterpose.go` (detection and coordinator invocation) and the manifest fields.

- Manifest: `cuinterpose.requested` (the opt-in, read from the source Pod; drives the restore-time shim mount through `ns-bind-mount mount-cuinterpose-fd`, because CRIU re-opens file-backed mappings by path) and `cuinterpose.prepared` (the coordinator's prepare completed and wrote `cuinterpose.state`; drives the coordinator on restore). A prepared checkpoint without its state file is refused instead of silently restoring natively.
- Detection is fail-closed (`CheckCuinterposeEnablement`): a Pod that requested the shim must show a control socket for every CUDA process, a Pod that did not must show none; anything else fails the checkpoint with the processes named. Only sockets count; procfs environ is not evidence.
- Around the coordinator: stale `cuinterpose-*.sock` entries are removed before CRIU recreates the processes (a stale path makes CRIU's `bind()` fail); the coordinator binary is opened before the mount namespace changes; its `phase=… status=ok` lines are parsed and logged as structured fields; `cuinterpose_prepare` and `cuinterpose_restore` are timing phases; the restore mount is ordered before the PageBroker staging mount (which pops the last mount).
- Tests: detection truth table and socket cases; argv contract against a fake coordinator binary for prepare and restore, including failure reporting; Go constants pinned to `protocol.h` and the coordinator's flag names; restore mount ordering with a fake mounter; manifest round trip; `ns-bind-mount` destination pinned to the podcontract constant.
- Until the shim opens control sockets (lifecycle PR), an annotated Pod with CUDA processes is refused at checkpoint by the fail-closed rule.

## Origin

The Go orchestration of #155 and #110, and the detection rule of #152/#165.

<details><summary>#155 feat(agent): coordinate CUDA interposer lifecycle (Go part; the C part is quoted in the coordinator PR)</summary>

[...]

PR #215, opened 2026-09-04
feat(agent): cuinterpose shim symbol resolution and forwarding

galletas1712:
## Summary

Layer 5 of the eleven-PR [cuinterpose stack #293](https://github.com/ai-dynamo/snapshot/stack/293). Closed PR #214 is intentionally not in the active stack.

This PR replaces the inert shim with transparent CUDA forwarding and defines the POSIX ticket primitives later tracking layers use. No allocation or multicast state is tracked at this layer.

Cuinterpose intercepts all four ways an application can reach a driver entry point: direct ELF binding, `dlsym`, the `cuGetProcAddress*` family, and the CUDA runtime's `cudaGetDriverEntryPoint*` family. Resolver wrappers call the real resolver first and substitute only successful CUDA results, so the shim never invents APIs unsupported by the installed driver. CUDA 13.1 requests select the device-explicit multicast bind ABI.

The intercepted surface covers CUDA VMM create/release/retain/map/unmap/access/export/import/property APIs and the corresponding multicast operations. Every wrapper checks symbol availability and forwards arguments/results unchanged in this PR.

`posix.c` defines the sealed 256-byte memfd ticket used in later layers. A ticket names the original creator participant and endpoint, resource kind, and random 128-bit allocation ID. The allocation ID is the opaque resource identity and export-cache key. An importer sends an `EXPORT` request carrying the participant, resource kind, and allocation identity needed for lookup. The creator rejects unknown IDs and returns a fresh real CUDA descriptor over `SCM_RIGHTS` for a known cached resource. Both peer traffic and coordinator traffic terminate on the same owner-only per-process Unix socket once #217 starts it. The peer request path uses scoped descriptor ownership and direct error returns; it has no `goto` cleanup path.

Tests cover direct and resolver forwarding, ABI selection, ticket validation, descriptor passing, malformed ancillary data, unavailable symbols, and the rule that resolver failure is not substituted.

## Stack boundary

Based on #223. #216 implements coordinator orchestration and global validation. #217 turns the forwarding wrappers into active POSIX-FD tracking and starts the per-process listener.

## Validation

On the final stack, `make test` passes in all Go modules. The pinned CUDA 13.1 builder compiles the production shim and coordinator with `-Werror`; the forwarding suite passes 12 tests, and the protocol suite passes 9 tests, with ASan/UBSan enabled. The final published stack head `0eae9c4f9b66093332f00a332afaf56048e1e64a` passed the full `make check` gate, the CUDA 13.1 production build, and all 73 sanitizer-backed native cuinterpose tests. Its physical-GPU suite passed all 3 tests with no skips on two DRA-assigned NVIDIA B200 GPUs. The unicast test kept an explicit allocation-ID ticket-backed peer mapping per worker live across capture and restore; the multicast test required a nonzero multicast VA and shim-logical handle, exercised `BindAddr`, and passed its collective and captured-graph replay. Detailed hardware evidence and measurements are in #220.

### DIFF
--- a/agent/cmd/cuinterpose/interpose.c
+++ b/agent/cmd/cuinterpose/interpose.c
@@ -5,23 +5,109 @@
  */
 
 /*
- * Placeholder shim. It builds and is packaged as libcuinterpose.so so that the
- * delivery path (init container, LD_PRELOAD, restore mount) is final before
- * the interposer exists, and it intercepts nothing. The following changes
- * replace it: symbol resolution and forwarding, allocation tracking, the
- * checkpoint lifecycle, multicast objects.
+ * CUDA virtual memory management (VMM) entry points the shim replaces.
+ *
+ * This layer forwards every call to the real driver unchanged. It exists so
+ * that symbol resolution (symbols.c) can be built and tested on its own; the
+ * bookkeeping that tracks shared allocations is added on top of these
+ * wrappers in a later change.
  */
 
+#define _GNU_SOURCE
+
 #include <cuda.h>
-#include <string.h>
 
 #include "export.h"
 #include "protocol.h"
+#include "symbols.h"
 
-CUINTERPOSE_API const struct cuinterpose_build_info cuinterpose_build_info = {CUDA_VERSION, CUINTERPOSE_VERSION};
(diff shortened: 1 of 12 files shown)


galletas1712:

PR #217, opened 2026-09-04
feat(agent): track POSIX CUDA VMM state

galletas1712:
## Summary

Layer 7 of the eleven-PR [cuinterpose stack #293](https://github.com/ai-dynamo/snapshot/stack/293). Closed PR #214 is intentionally not in the active stack.

This PR implements running-state POSIX CUDA VMM tracking without checkpoint/restore lifecycle mechanics.

`cuMemCreate` has an explicit three-way gate:

| Requested handle type | Runtime behavior | Checkpoint behavior |
| --- | --- | --- |
| `0` | Private allocation passes through natively | Allowed |
| Exactly `CU_MEM_HANDLE_TYPE_POSIX_FILE_DESCRIPTOR` | Tracked and virtualized | Supported |
| Any other nonzero type, including FABRIC or combinations | Passes through to the driver | Refused by #216 preflight |

Only successful unsupported exportable creations increment the process-wide diagnostic. It is intentionally sticky for that process generation because cuinterpose did not track the resource and cannot prove that releasing one handle eliminated every derived sharing dependency. FABRIC logs once that checkpoint will be refused. A fork child discards inherited CUDA state and starts with a fresh count.

Tracked allocations receive stable random allocation IDs and tagged logical handles. Repeated imports and retained handles collapse onto one real driver handle per allocation per process. Mappings are indexed by address range; access grants merge by location; partial/ambiguous operations are refused; a possibly partial driver access failure marks state unknown so inspection fails closed.

On first POSIX export, the creator performs one real CUDA export, keeps that descriptor in the export cache under the allocation ID, and returns a sealed ticket FD to the application. An importing shim reads the ticket, connects to the original creator endpoint, requests the ticket's participant/resource/allocation identity, receives a fresh descriptor over `SCM_RIGHTS`, imports it, and returns another logical handle. The random allocation ID is the opaque resource identity and export-cache key. Re-export by an importer still names the original creator.

A non-ticket descriptor import passes through and is counted as live raw sharing until its driver handle is released. #216 refuses checkpoint while that count is nonzero. The export cache has independent locking, in-flight pinning, drain-on-drop, and lifecycle quiesce/resume so peer requests remain serviceable without the main state lock.

The library constructor establishes the participant ID and `/snapshot-control/cuinterpose-<namespace-pid>.sock`. `CUINTERPOSE_PARTICIPANT_ID` is the explicit identity override used by tests and controlled launchers. The socket is mode `0600`; the Pod-local control `emptyDir` is mounted through a container-name `subPath`, so ordinary other Pods have no filesystem path to it and Snapshot-managed target containers receive isolated views. The workload Pod remains one trust domain if its author deliberately gives a sidecar the reserved volume. This boundary does not claim protection from node root or an equivalently privileged workload. Fork children discard inherited CUDA bookkeeping and lazily create a new identity/socket on first CUDA activity. Control-endpoint ownership uses scoped cleanup and direct returns rather than `goto`.

## Stack boundary

Based on #216. #292 adds the isolated current allocation-content storage module; #218 invokes it from `SAVE_ALLOCATIONS`/`LOAD_ALLOCATIONS` and implements unicast teardown/rebuild.

## Validation

The pinned CUDA 13.1 builder passes 13 tracking tests, 9 table/cache tests, and 9 protocol tests with ASan/UBSan where configured. Coverage includes private/POSIX/FABRIC gate behavior, handle alias collapse, range mapping/access semantics, ticket exchange through a forked child, raw import accounting, listener concurrency, descriptor exhaustion, and table churn. `make test` passes in all Go modules. The final published stack head `0eae9c4f9b66093332f00a332afaf56048e1e64a` passed the full `make check` gate, the CUDA 13.1 production build, and all 73 sanitizer-backed native cuinterpose tests. Its physical-GPU suite passed all 3 tests with no skips on two DRA-assigned NVIDIA B200 GPUs. The unicast test kept an explicit allocation-ID ticket-backed peer mapping per worker live across capture and restore; the multicast test required a nonzero multicast VA and shim-logical handle, exercised `BindAddr`, and passed its collective and captured-graph replay. Detailed hardware evidence and measurements are in #220.

### DIFF
--- a/agent/cmd/cuinterpose/export_cache.c
+++ b/agent/cmd/cuinterpose/export_cache.c
@@ -1,0 +1,219 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
+ * All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#define _GNU_SOURCE
+
+#include "export_cache.h"
+
+#include <fcntl.h>
+#include <pthread.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+
+#include "util/table.h"
+
+struct entry {
+  int fd;
+  unsigned in_flight; /* requests that duplicated fd and have not finished */
+  bool draining; /* no new requests; waiting for in_flight to reach zero */
+};
+
+static pthread_mutex_t lock = PTHREAD_MUTEX_INITIALIZER;
+static pthread_cond_t idle = PTHREAD_COND_INITIALIZER;
+static struct table entries;
+static bool accepting = true;
+
+/* Caller holds lock. Waits until nobody is using the entry, then closes it. */
+static void
+retire_locked(struct key key, struct entry* entry)
+{
+  entry->draining = true;
+  while (entry->in_flight != 0)
+    pthread_cond_wait(&idle, &lock);
+  table_remove(&entries, key);
+  close(entry->fd);
+  free(entry);
+}
+
+int
+cuinterpose_export_cache_put(const uint8_t id[CUINTERPOSE_ALLOCATION_ID_SIZE], int fd)
+{
+  struct key key = key_bytes(id);
+  struct entry* entry = calloc(1, sizeof(*entry));
+  struct entry* existing;
+
+  if (entry == NULL)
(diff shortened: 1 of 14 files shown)


galletas1712:

PR #220, opened 2026-09-04
test(agent): exercise cuinterpose checkpoint and restore on GPUs

galletas1712:
## Summary

Layer 11 of the eleven-PR [cuinterpose stack #293](https://github.com/ai-dynamo/snapshot/stack/293), based on #219. Closed PR #214 is not in the active stack.

This PR adds real-driver two-GPU checkpoint/restore tests under `agent/cmd/cuinterpose/tests/gpu`. The uninterposed controller launches an interposed parent and two worker ranks through `cuda-checkpoint --launch-job`. It drives interposer preparation, native CUDA lock/checkpoint/restore/unlock, interposer reconstruction, and workload assertions. It does not deploy the operator or namespace-entry wrapper.

### Shared and private VMM regression

The tests distinguish actual sharing from export capability:

- A ticket-backed peer allocation remains live across checkpoint and restore.
- A large seeded allocation is explicitly exported, so carrier throughput continues to measure the interposer-owned path.
- An additional POSIX-capable VMM allocation is never exported or bound into multicast. It remains native-owned through the updated #218/#219 lifecycle.
- After restore, the private allocation's existing VA must contain its original seeded bytes; its original handle must support `cuMemGetAllocationPropertiesFromHandle`; retain-by-address and normal cleanup must succeed.
- Carrier byte/count assertions include only actually shared creators rather than every tracked creator.

The suite also retains raw-import preflight refusal, creator/peer content checks, PyTorch symmetric-memory sharing, captured all-reduce replay, and post-restore raw import coverage. The multicast test requires real multicast selection, exercises bindings including `BindAddr`, and validates its collective and captured graph after reconstruction.

The suite skips absent prerequisites: Python CUDA/PyTorch bindings, build tools or prebuilt binaries, launch-job support, two distinct GPUs, and NVLink/multicast support where required. `CUINTERPOSE_BUILD_DIR` selects node-built binaries; otherwise `CUDA_HOME` supplies build headers.

## Stack transport

The ownership change is introduced in #218, merged through #219, then merged here. The GPU regression is a separate commit in this PR. Existing branch history is preserved; no force-push is needed. No merged-test-branch NIXL, CustomStorage, driver patch, or phase-batching code is included.

## Validation

Preparation checks passed:

```text
git diff --check
cc -fsyntax-only -Wall -Wextra -Werror (interposer, CUDA 13.1 headers)
.venv/bin/python -m py_compile tests/gpu/worker.py tests/gpu/test_posix.py
```

[...]

PR #222, opened 2026-09-04
feat: enable cuinterpose for annotated Pods

galletas1712:
## Summary

Layer 3 of the eleven-PR [cuinterpose stack #293](https://github.com/ai-dynamo/snapshot/stack/293). Closed PR #214 is intentionally not in the active stack.

This PR introduces the explicit workload opt-in and makes the shim part of each target process from startup:

```yaml
metadata:
  annotations:
    nvidia.com/cuinterpose: enabled
```

An absent annotation disables interposition; any other present value is rejected. CUDA tool delivery and multi-GPU launch wrapping are independent contracts owned by #213. For an enabled template, `ShapeCuinterposeCapture` only places `/tmp/snapshot-cuda/libcuinterpose.so` first in each target's `LD_PRELOAD`. Existing preload entries are retained, duplicates are removed, and a `valueFrom`-backed `LD_PRELOAD` is rejected because it cannot be rewritten safely. The annotation does not cause a one-GPU target to use `cuda-checkpoint --launch-job`.

The shape is idempotent and mutates only a deep copy until every check succeeds. The operator verifies the immutable contract before adopting an existing source Job and records the opt-in annotation on the generated `PodSnapshot`.

At this layer the packaged shim is still a placeholder. The durable contribution is the process-start contract: later layers can intercept each target process and CUDA-using fork child from its first CUDA operation.

## Stack boundary

Based on #213. #223 adds agent-side detection and coordinator invocation. This PR does not start shim sockets, assign participant identities, or alter CUDA behavior.

## Validation

On the final stack, `make test` passes in `api`, `agent`, and `operator`. The pinned CUDA 13.1 builder compiles the production binaries and passes all cuinterpose fake-driver suites. The final published stack head `0eae9c4f9b66093332f00a332afaf56048e1e64a` passed the full `make check` gate, the CUDA 13.1 production build, and all 73 sanitizer-backed native cuinterpose tests. Its physical-GPU suite passed all 3 tests with no skips on two DRA-assigned NVIDIA B200 GPUs. The unicast test kept an explicit allocation-ID ticket-backed peer mapping per worker live across capture and restore; the multicast test required a nonzero multicast VA and shim-logical handle, exercised `BindAddr`, and passed its collective and captured-graph replay. Detailed hardware evidence and measurements are in #220.

### DIFF
--- a/api/podcontract/cuinterpose.go
+++ b/api/podcontract/cuinterpose.go
@@ -1,0 +1,140 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package podcontract
+
+import (
+	"fmt"
+	"slices"
+	"strings"
+	"unicode"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+// cuinterpose is the CUDA interposer: an LD_PRELOAD library that lets Snapshot
+// checkpoint and restore CUDA memory shared between processes on one node. A
+// Pod opts in with CuinterposeAnnotation. The CUDA tools contract
+// (cudatools.go) puts the shim in every checkpoint target independently of
+// that annotation; shaping only puts the already-mounted shim first in
+// LD_PRELOAD.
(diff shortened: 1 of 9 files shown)


galletas1712:

PR #225, opened 2026-09-06
📝 docs(guides): verify TensorRT-LLM without a GPU driver (backport #224)

leeZardNav:
Backport of #224 to `release/0.1`. Clean cherry-pick of 4009e4a — the branch carried the identical pre-fix text, so no adaptation.

`import tensorrt_llm` needs `libcuda.so.1`, which the NVIDIA container runtime injects only when the container is given GPU access, so the verify step fails on any build host without a GPU. `importlib.util.find_spec` resolves the package without importing it.

See #224 for details.

### DIFF
--- a/docs/guides/tensorrt-llm.md
+++ b/docs/guides/tensorrt-llm.md
@@ -86,11 +86,15 @@ docker run --rm \
   --platform linux/amd64 \
   --entrypoint python3 \
   "$TENSORRT_LLM_SNAPSHOT_IMAGE" \
-  -c 'import pathlib; import tensorrt_llm; assert pathlib.Path("/app/app.py").is_file()'
+  -c 'import importlib.util, pathlib; assert importlib.util.find_spec("tensorrt_llm"); assert pathlib.Path("/app/app.py").is_file()'
 ```
 
 The command produces no output when both TensorRT-LLM and `/app/app.py` are
-present. Any failure prints an error and returns a non-zero exit status.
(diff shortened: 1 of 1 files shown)


docs/guides/tensorrt-llm.md
```diff
@@ -86,11 +86,15 @@ docker run --rm \
   --platform linux/amd64 \
   --entrypoint python3 \
   "$TENSORRT_LLM_SNAPSHOT_IMAGE" \
-  -c 'import pathlib; import tensorrt_llm; assert pathlib.Path("/app/app.py").is_file()'
+  -c 'import importlib.util, pathlib; assert importlib.util.find_spec("tensorrt_llm"); assert pathlib.Path("/app/app.py").is_file()'
```
oleg-kushniriov, line 89: Heads up: there's a PR in progress that removes this whole "build a custom TensorRT-LLM image, verify it with docker run" step entirely, since the guide moves to running the stock upstream image directly with app.py mounted from a ConfigMap instead of a Snapshot-built image. Doesn't block this backport - just flagging for whoever later backports that change to release/0.1, since it'll supersede this fix

Ronkahn21 (approved):

docs/guides/tensorrt-llm.md
leeZardNav: ok, thanks for the heads up! i'll merge it and whenever the fix comes in it'll override it, right?

PR #250, opened 2026-09-08
build: publish PageBroker as its own image

### LINKED ISSUE #238: PVC storage capture and restore flows should work with pagebroker

oleg-kushniriov:
Fixes #238

## Summary

Build and publish the PageBroker daemon as its own image, `ghcr.io/ai-dynamo/snapshot/pagebroker`, instead of bundling it into the agent image. No behaviour change: `pageBroker.enabled` still defaults to `false`.

### Image
- `agent/pagebroker/Dockerfile` is now the real build. The build stage runs the C++ test suite, compiles the daemon with protobuf linked statically, and fetches the protobuf corresponding source pinned to the exact `libprotobuf-dev` source version it linked against. The runtime stage is `nvcr.io/nvidia/distroless/cc` and carries only the daemon plus `/legal/`: the daemon source, the pinned protobuf source and its `VERSION`, a `source/README.txt` recording the base image's source location, and a generated `THIRD-PARTY.txt`.
- The agent Dockerfile drops the `pagebroker-builder` stage, the binary copy, the usage smoke test, and the C++ toolchain from its tester stage.

### Build and publish
- New `make docker-build-pagebroker`.
- `push-artifacts.yaml` gets a `build-pagebroker` job with its own cache scope. The Helm chart job depends on all three images.
- `hack/resolve-snapshot-image-tag.py` requires the `snapshot/pagebroker` package to be published for the commit under test.

### Chart
- New `image.pageBroker.repository` and `image.pageBroker.pullPolicy`. The sidecar renders at `image.agent.tag`, so the agent and daemon cannot skew. The old `pageBroker.image.*` block is removed.

### Docs
- Build-from-source guide, chart README values and retrieved-images tables, API reference.

## Verification

- `helm unittest` 10 passed, `helm lint` clean, `helm template --set pageBroker.enabled=true` renders `ghcr.io/ai-dynamo/snapshot/pagebroker:<agent tag>`.
- Not built locally: Docker Desktop on the authoring machine is failing with I/O errors on its image store. The first `push-artifacts` run on this branch builds the image and runs the C++ tests, which previously ran nowhere in CI.

## Notes for review

- The C++ test suite now gates the PageBroker image build. A failing daemon test blocks publishing.

## Summary by CodeRabbit

* **New Features**
  * PageBroker is now built, published, and deployed as a dedicated container image.
  * Helm charts support configuring the PageBroker image repository and pull policy while sharing the agent image tag.
  * PageBroker images include required licensing and source materials.

* **Documentation**
  * Updated deployment, API, chart, and source-build documentation for PageBroker image configuration and compatibility requirements.

* **Tests**
  * Added Helm validation and automated checks for PageBroker image builds, tags, and publication.


Ronkahn21 (commented):
...

Ronkahn21 (changes requested):
Required OSRB
CC @danbar2

hhzhang16 (dismissed):

danbar2:
## Base image for the PageBroker image needs to come from an NVIDIA-managed registry

Flagging this before the image is published, since it is cheap to change now and expensive to change after the first tag ships.

`agent/pagebroker/Dockerfile` builds on:

```dockerfile
ARG PAGEBROKER_BASE_IMAGE=ubuntu:24.04@sha256:33ceb71981b602c1a7443a53469e4dba065f7503eab3078a2d7a57a2ab987517
FROM ${DOCKER_PROXY}library/${PAGEBROKER_BASE_IMAGE}
```

With `DOCKER_PROXY` empty — which is how CI builds it — that resolves to `docker.io/library/ubuntu:24.04`.

Ubuntu 24.04 **is** an approved base container, so the image itself is fine. The issue is where it is pulled from. The OSS container release requirements say approved base containers must be pulled from an NVIDIA-managed repository, and the single exception is the public NVIDIA CUDA images on Docker Hub. A `library/ubuntu` pull from Docker Hub is not covered by that exception.

For reference, the approved base container families are:

| Base container | Notes |
|---|---|
| NVIDIA CUDA | the one family that may be pulled from Docker Hub (`nvidia/cuda`) |
| PyTorch | |
| TensorFlow | |
| Ubuntu 20.04 / 22.04 / 24.04 | approved, but must come from an NVIDIA-managed registry |
| Docker Distroless container images | |
| DLFW base containers | what the agent image uses |
| BioNeMo FW | |
| OV Kit Kernel | |
| DeepStream x86 Triton (DS 8.0) | |
| DGX `cuda:base` / `cuda:runtime` / `cuda:devel` | |

Everything else in this repo already satisfies this, which is why the PageBroker image stands out:

| Image | Base | Registry |
|---|---|---|
| agent | `nvcr.io/nvidia/cuda-dl-base:25.11-cuda13.0-devel-ubuntu24.04` | nvcr.io |
| operator | `nvcr.io/nvidia/distroless/go:v4.0.8@sha256:938bd28c…` | nvcr.io |
| **pagebroker** | **`ubuntu:24.04@sha256:33ceb719…`** | **Docker Hub** |

### Suggested fix

**`nvcr.io/nvidia/distroless/cc`**, pinned by digest. It is the C/C++ sibling of the distroless base the operator already ships on, so it is a base this repo has established, and it is the right shape for a single C++ daemon. It also has a second benefit worth taking: on a distroless base you would link protobuf statically, which removes `libprotobuf32t64` — the image's only third-party runtime package — and leaves the image with no third-party runtime components at all. That shrinks the source and attribution work described in the companion comment to almost nothing.

If distroless turns out to be impractical for the daemon, the fallback is an nvcr.io-hosted Ubuntu 24.04 rather than `library/ubuntu`.

Either way, keep the digest pin. It is what ties the recorded base source location to the base actually built on, which is the same commitment the agent and operator images already make.

danbar2:
## PageBroker image is missing corresponding source and a consolidated attribution file

Companion to the base-image comment. The agent and operator images both discharge two obligations that the PageBroker image does not yet:

1. **corresponding source** for every open-source binary added on top of the base — permissive as well as copyleft — shipped in-image under `/legal/source/`
2. a **consolidated `/legal/THIRD-PARTY.txt`**, generated at build time so it cannot drift from what is actually installed

Today `agent/pagebroker/Dockerfile` ships the NVIDIA source and one upstream notice:

```dockerfile
COPY --from=build /src /legal/source/pagebroker
RUN mkdir -p /legal/libprotobuf \
    && cp /usr/share/doc/libprotobuf32t64/copyright /legal/libprotobuf/copyright
```

That covers our own code and a single notice file. It does not ship the Debian source for `libprotobuf32t64`, and there is no `THIRD-PARTY.txt`. A per-component `copyright` file discharges attribution for that one component; it does not discharge source, and it is not the consolidated form.

### The good news: the scope here is tiny

I diffed the built image's package set against the base by pulling the base's `var/lib/dpkg/status` from the registry, so this is the authoritative delta rather than the apt closure:

| Package | Version | Source package | License |
|---|---|---|---|
| `libprotobuf32t64` | 3.21.12-8.2ubuntu0.3 | `protobuf` | BSD-3-Clause |

That is the entire delta. `libprotobuf32t64` declares `libc6 (>= 2.38)`, `libgcc-s1 (>= 3.3.1)`, `libstdc++6 (>= 13.1)` and `zlib1g (>= 1:1.1.4)`, and all four are already in the digest-pinned base at satisfying versions (Ubuntu 24.04.4 LTS, 92 packages) — so apt installs exactly one package and upgrades none. One `.dsc` plus tarball, one copyright stanza.

The build-stage toolchain (`build-essential`, `libgtest-dev`, `libprotobuf-dev`, `pkg-config`, `protobuf-compiler`) correctly does not reach the runtime stage, so GoogleTest and the compiler carry no obligation here. Good call splitting the stages that way.

### Suggested fix: reuse what the agent image already has

`agent/compliance/collect-sources.sh` and `agent/compliance/gen-attribution.sh` do exactly this job and are parameterised, not agent-specific — `gen-attribution.sh` takes the vendor directory as an argument and tolerates it being absent, which is the case here since this image has no Go code. Roughly:

```dockerfile
FROM ${PAGEBROKER_BASE_IMAGE} AS sources
COPY compliance/base-packages-pagebroker.tsv /tmp/base-packages.tsv
COPY compliance/collect-sources.sh /tmp/collect-sources.sh
RUN sh /tmp/collect-sources.sh /tmp/base-packages.tsv /sources/dpkg
```

then `COPY --from=sources /sources /legal/source` and a `gen-attribution.sh` invocation in the runtime stage, matching the agent.

Two mechanical notes:

- **Build context.** `make docker-build-pagebroker` currently passes `agent/pagebroker/` as the context, so `agent/compliance/` is not reachable. Either widen the context to `agent/` (and narrow the `COPY . .` in the build stage accordingly, so `/legal/source/pagebroker` does not swallow the whole agent tree), or move the shared compliance scripts somewhere both Dockerfiles can see.
- **Baseline manifest.** The delta is computed against a committed baseline, and `agent/compliance/base-packages.tsv` is the CUDA-DL base — this image needs its own capture, plus the equivalent `verify-base-packages` CI check so a base bump cannot silently skew the delta.

### Worth sequencing after the base-image decision

If the base moves to `nvcr.io/nvidia/distroless/cc` as suggested in the other comment, protobuf gets linked statically and `libprotobuf32t64` disappears from the image. The delta becomes empty, and all of the above reduces to shipping the protobuf source alongside the NVIDIA source under `/legal/source/` plus a short `THIRD-PARTY.txt` — no baseline manifest, no `collect-sources.sh` wiring, no CI check. Static linking does not remove the source obligation for protobuf; it removes the Debian machinery around it.

So: settle the base first, then wire whichever of the two shapes falls out.

danbar2:
## Two smaller gaps in the `/legal` tree

The distroless rebase and the static protobuf link in 0ab184b address both earlier comments, and the runtime image now adds zero packages on top of its base — that is a better outcome than either comment asked for. Two loose ends remain in the legal materials, both small.

### 1. `/legal/source/` has no README recording the base image's source location

The agent and operator images ship `/legal/source/README.txt` (from `agent/compliance/README.source.txt`), which says what the tree contains and — the part that matters here — where the **base image's** source is published. The requirement is to document every location where required source is made available, including the base container's, and the base's own source is what covers the libraries this daemon links dynamically (`libstdc++6`, `libgcc_s`, `libc6`).

The PageBroker image has no such README, so that location is recorded nowhere in the image.

For `nvcr.io/nvidia/distroless/cc:v4.0.8` it is:

```
https://developer.download.nvidia.com/distroless-oss/cc/v4.0.8/
```

I checked that it resolves and holds per-package `.dsc` files and source tarballs for that exact version, including `glibc_2.41-12+deb13u3` and `gcc-14_14.2.0-19` — the sources for the libraries the binary links. Worth adding a short `README.txt` to `/legal/source/` in the same shape as the agent's, listing the two entries already in the tree plus that URL.

### 2. The protobuf source is not pinned to the version actually linked

This one is a correctness issue rather than a missing file.

`libprotobuf.a` is linked in the **`build`** stage, which installs its own `libprotobuf-dev`. The corresponding source is fetched in the **`sources`** stage, which installs a *second, independent* `libprotobuf-dev` and then runs:

```dockerfile
apt-get source --only-source --download-only protobuf
```

with no version. That resolves to whatever the archive's candidate source version is at the time that stage builds.

Those are two separate apt transactions in two separate stages with independent layer caches. If the archive publishes a protobuf point release between the two layers being built — or if `build` is served from cache while `sources` rebuilds — the image ships source for a different version than the code it actually linked. That is exactly the drift the source obligation exists to prevent, and it fails silently: nothing in the build errors out, and the shipped `.dsc` looks perfectly legitimate.

`agent/compliance/collect-sources.sh` already avoids this by pinning to `${source:Version}` for the installed binary, which is the pattern to match.

Cleanest fix is to fetch the source in the `build` stage, where the `.a` being linked actually comes from, and pin it:

```dockerfile
RUN mkdir -p /legal/source/protobuf \
    && (cd /legal/source/protobuf \
        && apt-get source --only-source --download-only \
             "protobuf=$(dpkg-query -W -f='${source:Version}' libprotobuf-dev)") \
    && cp /usr/share/doc/libprotobuf-dev/copyright /legal/source/protobuf/copyright
```

then `COPY --from=build /legal/source/protobuf ...` into the sources stage. That removes the second `libprotobuf-dev` install entirely and makes the shipped source provably the source of the linked object, rather than probably.

If you would rather keep the fetch in the `sources` stage, the equivalent is to write the version out in `build` and read it back:

```dockerfile
# build stage
RUN dpkg-query -W -f='${source:Version}' libprotobuf-dev > /src/.protobuf-src-version
```

Pinning within the `sources` stage to *its own* installed version is not sufficient — it makes that stage self-consistent while still allowing it to diverge from `build`.

agent/pagebroker/Dockerfile
```diff
@@ -18,19 +19,49 @@ WORKDIR /src
...
+RUN apt-get update \
+    && for f in /etc/apt/sources.list.d/*.sources; do \
+         [ -f "$f" ] && sed -i 's/^Types: deb$/Types: deb deb-src/' "$f" || true; \
+       done \
+    && [ -f /etc/apt/sources.list ] \
+       && sed -i 's/^deb \(.*\)$/deb \1\ndeb-src \1/' /etc/apt/sources.list || true \
+    && apt-get install -y --no-install-recommends libprotobuf-dev \
+    && apt-get update \
+    && mkdir -p /legal/source/protobuf \
+    && (cd /legal/source/protobuf && apt-get source --only-source --download-only protobuf) \
```
danbar2: need to pin protobuf to specific version

danbar2:
## Concrete patches for the two `/legal` gaps

Restating my previous comment as drop-in changes. Both are in `agent/pagebroker/`; the runtime stage does not change.

---

## Fix 1 — add `/legal/source/README.txt`

The operator image already ships one and is the right template here: same situation, distroless base with no system packages, one class of third-party content plus a pointer to the base's own source. See [`operator/compliance/README.source.txt`](https://github.com/ai-dynamo/snapshot/blob/main/operator/compliance/README.source.txt) and how it is wired at `operator/Dockerfile:22`.

Without it, the base image's source location — which is what covers the `glibc`, `libstdc++` and `libgcc` this binary links dynamically — is recorded nowhere in the image.

**New file: `agent/pagebroker/compliance/README.source.txt`**

```text
Corresponding source
================================================================================

Upstream source for the third-party components redistributed in this image.

  protobuf/    Debian source package (.dsc, upstream tarball, Debian diff) for
               protobuf, statically linked into /usr/local/bin/pagebroker, at
               the source version the linked libprotobuf.a was built from.

  pagebroker/  NVIDIA-authored source for the daemon itself, as built. Build
               artifacts and protoc-generated files are excluded; `make
               generate` reproduces the latter from the shipped .proto.

This image is distroless and installs no system packages, so statically linked
protobuf is the only third-party content this image adds. The base image
contains third-party components of its own -- glibc, libstdc++ and libgcc,
which this binary links dynamically. Source for the base image's own contents
is published by NVIDIA and is not duplicated here:

  https://developer.download.nvidia.com/distroless-oss/cc/v4.0.8/

NVIDIA-authored code in this image is Apache-2.0 and published at
https://github.com/ai-dynamo/snapshot.

Per-component license texts are in /legal/THIRD-PARTY.txt.
```

I verified that URL resolves and holds per-package `.dsc` files and source tarballs for exactly `v4.0.8`, including `glibc_2.41-12+deb13u3` and `gcc-14_14.2.0-19`.

Note the `COPY` path in Fix 2 is `compliance/README.source.txt`, not `agent/pagebroker/compliance/...` as the operator's line reads — `make docker-build-pagebroker` passes `agent/pagebroker/` as the build context, so paths are relative to that.

---

## Fix 2 — pin the protobuf source to the version actually linked

Today `libprotobuf.a` is linked in the `build` stage, and the source is fetched in the `sources` stage from a **second, independent** `libprotobuf-dev` install, unpinned. Two stages, two apt transactions, two layer caches: if the archive publishes a point release between them, or one is cache-hit while the other rebuilds, the image ships source for a different version than the code it linked — silently, with a `.dsc` that looks perfectly legitimate.

The fix moves the fetch into `build`, where the `.a` comes from, and pins it. That also deletes the duplicate `libprotobuf-dev` install entirely.

**`build` stage — append after the smoke test:**

```dockerfile
# Corresponding source for the statically linked protobuf. Fetched in this
# stage and pinned to the source version of the libprotobuf-dev that provided
# the libprotobuf.a linked above, so the shipped source cannot be a different
# version than the linked object.
RUN for f in /etc/apt/sources.list.d/*.sources; do \
      [ -f "$f" ] && sed -i 's/^Types: deb$/Types: deb deb-src/' "$f" || true; \
    done \
    && { [ ! -f /etc/apt/sources.list ] \
         || sed -i 's/^deb \(.*\)$/deb \1\ndeb-src \1/' /etc/apt/sources.list; } \
    && apt-get update \
    && mkdir -p /legal/source/protobuf \
    && (cd /legal/source/protobuf \
        && apt-get source --only-source --download-only \
             "protobuf=$(dpkg-query -W -f='${source:Version}' libprotobuf-dev)") \
    && cp /usr/share/doc/libprotobuf-dev/copyright /legal/source/protobuf/copyright \
    && rm -rf /var/lib/apt/lists/*
```

> Keep the single quotes around `'${source:Version}'`. It is a `dpkg-query` format string, not a shell variable — unquoted, `sh` tries to expand it and fails with a bad-substitution error.

**`sources` stage — replace the whole stage:**

```dockerfile
# Assemble the legal materials so the runtime stage (distroless, no shell) can
# receive them in a single COPY.
FROM ${DOCKER_PROXY}library/${PAGEBROKER_BUILD_IMAGE} AS sources

COPY --from=build /legal/source/protobuf /legal/source/protobuf
COPY --from=build /src /legal/source/pagebroker
COPY compliance/README.source.txt /legal/source/README.txt

RUN rm -rf /legal/source/pagebroker/build \
           /legal/source/pagebroker/pagebroker \
           /legal/source/pagebroker/v1/pagebroker.pb.*

RUN { \
      printf '================================================================================\n'; \
      printf 'THIRD-PARTY SOFTWARE NOTICES AND ATTRIBUTION\n'; \
      printf 'NVIDIA Dynamo Snapshot (PageBroker image)\n'; \
      printf '================================================================================\n\n'; \
      printf 'This image statically links protobuf. Corresponding source ships under\n'; \
      printf '/legal/source/protobuf/.\n\n'; \
      printf '================================================================================\n'; \
      printf 'COMPONENT: protobuf\n'; \
      printf '================================================================================\n'; \
      cat /legal/source/protobuf/copyright; \
    } > /legal/THIRD-PARTY.txt
```

Two things changed beyond the `COPY`s: the stage no longer installs `libprotobuf-dev` or touches apt at all, and `THIRD-PARTY.txt` now reads the copyright from `/legal/source/protobuf/copyright` rather than `/usr/share/doc/libprotobuf-dev/copyright`, which is no longer present in this stage.

The runtime stage is unchanged — its existing `COPY --from=sources /legal /legal` carries the README through.

---

### Why this one matters beyond tidiness

Shipping source that does not correspond to the shipped binary is the specific failure the source obligation exists to prevent, and it is the kind of thing that is very hard to argue after the fact. Pinning makes the correspondence a property of the build rather than a timing coincidence.

PR #258, opened 2026-09-08
feat(agent): add Model Streamer transfer engine

tzachi-runai:
## Summary

Adds Model Streamer as a PageBroker transfer engine for filesystem-backed checkpoint restores. It builds a validated restore plan, maps destination files, submits byte-budgeted reads through the shared event loop, and routes responses to the owning restore. Submissions target 10 GB, with larger files submitted individually, and have a two-hour completion deadline. Terminally failed sessions are replaced for subsequent restores. Checkpoint publication uses filesystem copy.

Both transfer engines use shared filesystem helpers for path validation, restore sizing, destination conflict checks, and checkpoint publishing. The Model Streamer engine has no dependency on `PosixCopyEngine`. Filesystem validation and checkpoint publishing tests run against both engines, and the Model Streamer recovery test links without the POSIX engine.

The stack is based on the [dedicated PageBroker image](https://github.com/ai-dynamo/snapshot/pull/250). Model Streamer's native library, license, and metadata are packaged only in that image. `agent/Dockerfile` has no changes relative to #250, and `docker-build-agent` does not require a Streamer wheel. The PageBroker build preserves the distroless runtime, static protobuf linkage, C++ test gate, and collected source materials from #250.

The compatible wheel remains pinned by SHA-256, with the upstream ABI commit documented. `docker-build-pagebroker` accepts the wheel through `MODEL_STREAMER_WHEEL_DIR`. A matching artifact must be provisioned for hosted image builds while a compatible Streamer release is pending. Image approval remains a separate prerequisite for release.

This layer depends on the [PageBroker transfer foundation](https://github.com/ai-dynamo/snapshot/pull/264). Runtime configuration is added by the [top stack layer](https://github.com/ai-dynamo/snapshot/pull/265).

## Verification

Validated the rebased stack:

- `make check` and `make test`
- dedicated PageBroker image built locally with the pinned wheel
- 35 C++ tests and the Model Streamer recovery test passed during the image build
- final distroless image startup smoke test passed
- `agent/Dockerfile` matches #250 exactly

## API impact

Adds an internal PageBroker protocol engine value. This layer alone does not change the default engine or expose deployment configuration.

## Tracking issue


tzachi-runai:
need to fix the discription

agent/pagebroker/Dockerfile
```diff
@@ -1,8 +1,32 @@
 # SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+FROM python:3.12-slim AS model-streamer-install
+
+COPY model-streamer-wheel.sha256 /model-streamer-wheel.sha256
```
dfeigin-nv, line 12: Why not as version?

agent/Dockerfile
```diff
@@ -33,6 +35,31 @@ ARG AGENT_BASE_IMAGE=nvcr.io/nvidia/cuda-dl-base:25.11-cuda13.0-devel-ubuntu24.0
 # but placeholder builds MUST override it with --build-arg BASE_IMAGE=<image>
 ARG BASE_IMAGE=placeholder-requires-base-image-arg
 
+# =============================================================================
+# Stage: Model Streamer native SDK extracted from its Python wheel
+# =============================================================================
+FROM python:3.12-slim AS model-streamer-install
+
+COPY pagebroker/model-streamer-wheel.sha256 /model-streamer-wheel.sha256
+COPY --from=model-streamer-wheel / /wheels/
```
dfeigin-nv: Why does the agent need model streamer?

agent/pagebroker/transfer/model_streamer_api.hpp
```diff
@@ -0,0 +1,44 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include <cstddef>
+#include <cstdint>
+
+// Snapshot-owned declarations for the libstreamer.so C ABI. These signatures
+// match run-ai/runai-model-streamer commit bc21fd4182cc06ce9475452d16697d50ce3588c4
```
dfeigin-nv, line 10: There is no streamer release compatible with this ABI?

agent/pagebroker/transfer/model_streamer_api.hpp
```diff
@@ -0,0 +1,44 @@
...
+// (multi-submission, per-range destinations). The public 0.16.1 ABI is older
+// and is not compatible with these declarations.
+namespace snapshot::pagebroker::model_streamer_api {
+using SubmissionId = std::uint64_t;
+
+// ResponseCode::TimedOut in the pinned Model Streamer ABI. A finite response
+// timeout lets the single API thread notice and submit newly queued work.
+inline constexpr int kTimedOut = 16;
```
dfeigin-nv: maybe rename to kTimedOutStatusCode or something? 16 is unclear in itself

agent/pagebroker/transfer/model_streamer_api.hpp
```diff
@@ -0,0 +1,44 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include <cstddef>
+#include <cstdint>
+
```
dfeigin-nv, line 8: Was this file copied form runai-model-streamer? If so we should clearly state what path was it taken from

agent/pagebroker/transfer/restore_plan.hpp
```diff
@@ -0,0 +1,37 @@
...
+#include <cstdint>
+#include <filesystem>
+#include <string>
+#include <vector>
+
+namespace snapshot::pagebroker {
+using Path = std::filesystem::path;
+
+struct RestoreDirectory {
+  Path relative;
```
dfeigin-nv: relative to what?

agent/pagebroker/transfer/restore_plan.hpp
```diff
@@ -0,0 +1,37 @@
...
+  std::filesystem::perms permissions = std::filesystem::perms::unknown;
+};
+
+struct RestoreFile {
+  std::string source_locator;
```
dfeigin-nv, line 22: source_locator?

agent/pagebroker/transfer/restore_plan.hpp
```diff
@@ -0,0 +1,37 @@
...
+  Path relative;
+  uintmax_t bytes = 0;
```
dfeigin-nv: bytes means size?

agent/pagebroker/transfer/restore_plan.hpp
```diff
@@ -0,0 +1,37 @@
...
+  std::filesystem::perms permissions = std::filesystem::perms::unknown;
+};
+
+// A transfer engine resolves its storage source into this neutral plan before
+// submitting reads to Model Streamer. Destination paths must be unique,
+// normalized relative paths without dot components, and directories must be
+// ordered parent-first. Source locators are backend-specific and are never
+// joined to the destination; future backends may therefore use object URIs.
+struct RestorePlan {
+  std::filesystem::perms root_permissions = std::filesystem::perms::unknown;
```
dfeigin-nv, line 35: Is this whole file just for ModelStreamer?

agent/pagebroker/transfer/model_streamer_transfer_engine.hpp
```diff
@@ -0,0 +1,33 @@
...
+  void ValidateCheckpointDestination(const StorageBackend& destination) const override;
+  bool CheckpointDestinationConflicts(const StorageBackend& destination) const override;
+  void PublishCheckpoint(const Path& source, const StorageBackend& destination) const override;
+
+ private:
+  // Returns the current restore session, replacing a terminally failed one.
+  std::shared_ptr<ModelStreamerRestore> AcquireRestore() const;
+
+  Path storage_root_;
+  PosixCopyEngine posix_;
```
dfeigin-nv, line 29: Will posix_ be needed in S3?

agent/pagebroker/transfer/model_streamer_transfer_engine.cpp
```diff
@@ -0,0 +1,132 @@
...
+{
+}
+
+std::shared_ptr<ModelStreamerRestore>
+ModelStreamerTransferEngine::AcquireRestore() const
+{
+  std::shared_ptr<ModelStreamerRestore> previous;
+  std::shared_ptr<ModelStreamerRestore> current;
+  {
+    std::lock_guard lock(restore_mutex_);
```
dfeigin-nv, line 82: I didn't manage to understand this function

agent/pagebroker/transfer/model_streamer_restore.cpp
```diff
@@ -0,0 +1,456 @@
...
+#include "model_streamer_api.hpp"
+#include "utils/event_loop.hpp"
+
+namespace snapshot::pagebroker {
+namespace fs = std::filesystem;
+namespace streamer = model_streamer_api;
+namespace {
+// Use the Python streamer's default per-submission target: one quarter of its
+// 40 GB read-ahead budget. A single larger file is submitted on its own.
+constexpr uintmax_t kSubmissionByteBudget = 10'000'000'000;
```
dfeigin-nv, line 34: does this mean we have no more than 10GiB of memory requested at a time?

dfeigin-nv (commented):
Do we have benchmarks? I know without the criu-memory-providers the numbers won't be optimal, but I want to make sure the Streamer is utilized properly.

agent/pagebroker/Dockerfile
tzachi-runai: This is an unreleased build whose wheel version is 0.0.0, so the version doesn’t identify it reliably. The SHA pins the exact compatible artifact.

agent/Dockerfile
tzachi-runai: The agent Dockerfile builds and packages the PageBroker binary directly. That binary links against libstreamer.so, so the image needs the library. The Go agent itself doesn’t use Model Streamer.

agent/pagebroker/transfer/model_streamer_api.hpp
tzachi-runai: Not yet. The latest release, 0.16.1, uses the older API. This integration needs the newer API at bc21fd4, which is why it uses a pinned development wheel.

agent/pagebroker/transfer/model_streamer_api.hpp
tzachi-runai: These declarations match Model Streamer’s native API. I added the exact upstream paths alongside the pinned commit so their source is clear.

agent/pagebroker/transfer/model_streamer_api.hpp
tzachi-runai: implemented

agent/pagebroker/transfer/restore_plan.hpp
tzachi-runai: added a comment on this and changed the name to relative path

agent/pagebroker/transfer/restore_plan.hpp
tzachi-runai: It identifies the source file: currently a full filesystem path, and potentially an object URI for future backends. added comment clearifying

agent/pagebroker/transfer/restore_plan.hpp
tzachi-runai: Yes: it is the entire file size in bytes. Rename it to size_bytes to make that immediately clear.

agent/pagebroker/transfer/restore_plan.hpp
tzachi-runai: Currently only Model Streamer consumes it. The plan separates discovering source files from restoring them, so filesystem and future object-storage sources can use the same restore logic.

agent/pagebroker/transfer/model_streamer_transfer_engine.hpp
tzachi-runai: No.  It handles the filesystem operations in this PR, including checkpoint publishing. Direct S3 support would need S3-specific implementations; posix_ would remain relevant to filesystem-backed checkpoints.

agent/pagebroker/transfer/model_streamer_transfer_engine.cpp
tzachi-runai: In the current code, the transfer engine is created once and stays alive. It is not recreated after a request fails. The broker catches the error and aborts that transaction. Engine creation.

  AcquireRestore() is what performs the recovery inside that engine:

  - A healthy ModelStreamerRestore session is reused.
  - A failed session is replaced with a fresh one for subsequent restores.
  - Existing callers retain only the old wrapper while receiving errors and cleaning up.

agent/pagebroker/transfer/model_streamer_restore.cpp
tzachi-runai: It’s a 10 GB batch target per restore, not a total memory limit. A larger file is submitted on its own, and concurrent restores can exceed that amount collectively.

tzachi-runai:
@dfeigin-nv, regarding your [benchmark question](https://github.com/ai-dynamo/snapshot/pull/258#pullrequestreview-5195029734):

This PR establishes the integration for the upcoming S3 support, so I’d like to keep performance tuning in follow-up changes. I agree a basic measurement is useful to check that we’re using the streamer correctly. Would staging throughput and CPU/memory observations be sufficient here, with broader benchmarks following the S3 and memory-provider integration?

dfeigin-nv:
> @dfeigin-nv, regarding your https://github.com/ai-dynamo/snapshot/pull/258#pullrequestreview-5195029734:
> 
> This PR establishes the integration for the upcoming S3 support, so I’d like to keep performance tuning in follow-up changes. I agree a basic measurement is useful to check that we’re using the streamer correctly. Would staging throughput and CPU/memory observations be sufficient here, with broader benchmarks following the S3 and memory-provider integration?

agent/pagebroker/transfer/model_streamer_transfer_engine.cpp
dfeigin-nv: then maybe a comment?

agent/pagebroker/transfer/model_streamer_transfer_engine.hpp
dfeigin-nv: Is the ModelStreamer engine supposed to include another engine as a member?
There won't be an S3 engine. maybe you just need the filesystem functionalities?

agent/pagebroker/transfer/restore_plan.hpp
dfeigin-nv: but is it specific for filesystem?

agent/pagebroker/transfer/model_streamer_api.hpp
dfeigin-nv: We should ask Noa to create a release

agent/pagebroker/Dockerfile
dfeigin-nv: Let's ask Noa create a release. It shouldn't be an issue afaik

tzachi-runai:
- Setup: NVIDIA L40S, driver 595.91.07, existing integration image `pagebroker-transfer-20260908151556`.
- Workload: 8 GiB of initialized GPU data, producing an 8.449 GiB checkpoint across 36 files.
- Method: One checkpoint, two restores per engine, ordered POSIX → Model Streamer → Model Streamer → POSIX. Checkpoint files were fully read and hashed before each restore, outside timing.
- Same resource limits: PageBroker 32 CPUs / 72 GiB; agent 4 CPUs / 32 GiB.

“Staging” measures preparing and loading checkpoint files into PageBroker’s RAM-backed staging directory. “Agent restore” includes staging, CRIU restore, and CUDA restore; it excludes Pod startup and subsequent data verification.

| Run | Engine | Staging | Effective throughput | Agent restore |
|---|---|---:|---:|---:|
| 1 | POSIX | 3.861 s | 2.188 GiB/s | 11.546 s |
| 2 | Model Streamer | 1.814 s | 4.658 GiB/s | 9.439 s |
| 3 | Model Streamer | 1.763 s | 4.793 GiB/s | 9.488 s |
| 4 | POSIX | 3.438 s | 2.458 GiB/s | 11.070 s |

agent/Dockerfile
tzachi-runai: fixed as u requested where the stack is applied to Dan Bar's branch

agent/pagebroker/transfer/restore_plan.hpp
tzachi-runai: Yes, the current implementation is filesystem-based. When adding S3, I’ll adapt or extend the plan as needed. The intent is to support different sources while still restoring into filesystem staging

S3 → PageBroker staging files in RAM → CRIU restore → CUDA restore

 That means the filesystem destination fields in RestorePlan remain relevant when S3 becomes the source.

agent/pagebroker/transfer/model_streamer_transfer_engine.hpp
tzachi-runai: Agreed—posix_ is there to reuse filesystem operations. I’ll extract those into shared filesystem helpers so the engines remain independent. S3 will be another storage backend supported by the Model Streamer path.

agent/pagebroker/transfer/model_streamer_transfer_engine.cpp
tzachi-runai: done!

dfeigin-nv (approved):

PR #264, opened 2026-09-09
refactor(agent): prepare PageBroker transfer engines

tzachi-runai:
## Summary

Prepares PageBroker for multiple transfer engines without changing the selected runtime behavior:

- moves `TransferEngine` and `PosixCopyEngine` under `agent/pagebroker/transfer/`
- keeps the filesystem copy implementation owned by the POSIX engine
- adds a reusable single-threaded `EventLoop` with ordering, concurrent-producer, cancellation, and failure tests
- updates PageBroker build files and generated-file ignores for the new layout

This is the bottom layer of stack #304, based on the [dedicated PageBroker image](https://github.com/ai-dynamo/snapshot/pull/250). The next layers are the [Model Streamer transfer engine](https://github.com/ai-dynamo/snapshot/pull/258) and [runtime configuration](https://github.com/ai-dynamo/snapshot/pull/265).

## Verification

Validated the rebased stack:

- `make check` and `make test`
- dedicated PageBroker image build, including 28 C++ tests and the recovery test
- final distroless image startup smoke test

## API impact

No public API or runtime configuration change.

## Tracking issue


agent/pagebroker/utils/event_loop.cpp
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  event->Cancel(std::move(error));
+  return false;
+}
+
+void
+EventLoop::Stop() noexcept
+{
+  std::lock_guard stop_lock(stop_mutex_);
```
dfeigin-nv: Is Stop supposed to be called outside the dtor of EventLoop?  Like with external cancellation? If so I don't know what the mutex we take there - is there a chance for a deadlock?

dfeigin-nv (dismissed):

agent/pagebroker/utils/event_loop.cpp
tzachi-runai: I made sure that stop commands cannot come from within the worker thread, also added a comment on this. Changed the eventloop to use only 1 mutex as this is sufficient and removes the chance of deadlock,

dfeigin-nv (dismissed):

dfeigin-nv (approved):

dfeigin-nv (approved):

PR #265, opened 2026-09-09
feat(agent): configure PageBroker transfer engines

tzachi-runai:
## Summary

Exposes PageBroker transfer-engine selection end to end while preserving POSIX as the default:

- adds and validates `pageBroker.transferEngine` in agent configuration
- propagates the selected engine through checkpoint and restore requests
- supports `posix-copy` and `model-streamer`, normalizing empty values to `posix-copy`
- renders and validates the value in Helm
- includes configuration changes in the DaemonSet rollout checksum
- documents building separate agent and PageBroker images with a shared tag and using `image.pageBroker.repository`
- adds Go and chart tests
- updates DaemonSet test suites to load the ConfigMap used by the rollout checksum

This is the top layer of stack #304. It depends on the [Model Streamer transfer engine](https://github.com/ai-dynamo/snapshot/pull/258), with the stack based on the [dedicated PageBroker image](https://github.com/ai-dynamo/snapshot/pull/250).

## Verification

- `make check` and `make test` on the rebased stack
- all 19 Helm unit tests passed
- dedicated PageBroker image build, C++ tests, and runtime startup smoke test

## API impact

No Snapshot CRD API change. The Helm and agent configuration gain the optional `pageBroker.transferEngine` field; its default is `posix-copy`. The chart uses the separate PageBroker image introduced by #250.

## Tracking issue


charts/snapshot/templates/daemonset.yaml
```diff
@@ -28,8 +28,8 @@ spec:
         {{- with .Values.daemonset.podLabels }}
         {{- toYaml . | nindent 8 }}
         {{- end }}
-      {{- if or .Values.openshift.enabled .Values.daemonset.podAnnotations }}
```
dfeigin-nv, line 31: Why did we remove this if?

charts/snapshot/templates/daemonset.yaml
```diff
@@ -28,8 +28,8 @@ spec:
...
       annotations:
+        checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
```
dfeigin-nv, line 32: why do we need this anotation?

charts/snapshot/templates/daemonset.yaml
tzachi-runai: We now always add the config checksum annotation, so the annotations section is needed even without OpenShift or custom pod annotations. The OpenShift-specific condition is still there.

charts/snapshot/templates/daemonset.yaml
tzachi-runai: The agent reads its config only at startup. This checksum makes a Helm config change trigger a DaemonSet rollout, so changing the transfer engine takes effect without a manual restart.

dfeigin-nv (approved):

Issue #242, opened 2026-09-08
Github page for E2E status


danbar2:
## Summary

  Extend the framework checkpoint/restore E2E tests with performance measurements,
  historical comparisons, and a GitHub Pages dashboard.

  The nightly workflow should measure vLLM, SGLang, and TensorRT-LLM, retain results from previous runs, and
  make regressions or improvements visible over time.

  ## Required measurements

  At the end of each E2E test, print and store:

  - GPU model, UUID, driver version, and Kubernetes node.
  - Checkpoint duration: from creating the PodSnapshot, after the source pod is ready, until both PodSnapshot
    and PodSnapshotContent report Ready.

  - Restore-to-traffic duration: from the agent’s RestoreRequested event until the restored framework is ready
    to accept traffic.

  - Total test duration: from the beginning of test setup through successful restored inference validation.
  - Test outcome and result-file path.

  Also record the restore-pod-creation-to-traffic duration as a secondary metric to help distinguish
  scheduling delays from restore execution.

  Durations should use a monotonic clock. Results should retain timestamped lifecycle events so additional
  measurements can be derived later.

  ## Result storage and comparison

  Each matrix job should upload a versioned JSON result artifact. A single aggregation job should:

  - Run after all framework jobs, including failed jobs.
  - Validate and aggregate the result artifacts.
  - Preserve failed, timed-out, skipped, and infrastructure-failed runs.
  - Compare each measurement with:
      - The previous comparable successful run.
      - The median of the previous seven comparable successful runs.

  - Publish a GitHub Actions summary.
  - Store scheduled default-branch results on a data-only e2e-benchmark-history branch.

  Comparisons should only combine compatible environments, including the same framework image, model, GPU
  model, storage/cache configuration, and benchmark definition.

  PR runs must not modify durable history. Matrix jobs should remain read-only; only the aggregation job
  receives history-write permission.

  ## Dashboard

  Publish a static GitHub Pages dashboard showing:

  - Checkpoint duration over time.
  - Restore-to-traffic duration over time.
  - Total test duration over time.
  - Separate, consistently colored series for vLLM, SGLang, and TensorRT-LLM.
  - Date-range, framework, GPU, outcome, and test-suite filters.
  - Failed and timed-out runs as explicit markers or gaps rather than zero values.
  - Point details including comparison deltas, GPU, commit, workflow link, framework image, and outcome.
  - A latest-runs table with exact values.

[...]

PR #268, opened 2026-09-09
ci(e2e): persist framework benchmark history

oleg-kushniriov:
Part of #242. 

## Summary

Persist framework benchmark results on a data-only git branch and compare every run against its compatible history.

- `e2e/snapshot_e2e/benchmark_history.py` validates the versioned result envelope, synthesizes an explicit `infrastructure_failed` record for any expected framework without a valid artifact, compares each completed measurement with the previous compatible successful run and the median of the previous seven, and writes immutable raw results plus derived monthly NDJSON indexes and a manifest. It is idempotent per `(runId, runAttempt, suite, case, test)`.
- `.github/workflows/e2e-frameworks.yaml` adds two post-matrix jobs. `benchmark-history-read-only` runs for pull-request mirrors, ordinary branches, and manual dispatches with `contents: read` and only writes a step summary and a comparison artifact. `benchmark-history-publish` runs only for scheduled `main` runs, is the sole job with `contents: write`, serializes through a concurrency group, and pushes one commit per run to the orphan `e2e-benchmark-history` branch, bootstrapping it on first use.
- `.github/actions/fetch-benchmark-history` is a composite action shared by both jobs (and by the dashboard deployment in #281) that checks the history branch out with the job token and `GITHUB_SERVER_URL`.
- `e2e/README.md` documents the raw and index layout, the comparison dimensions, the emitted `comparisonKey`, the re-run semantics, and the rebuild and recovery procedure.
- `ci.yml` runs `test_benchmark_history.py` in the cluster-free `e2e-static` job.

## Behavior worth knowing

- Aggregation reads the artifacts of every attempt of the run and keeps the newest attempt per framework, so "Re-run failed jobs" does not record the frameworks that already passed as missing.
- Compatible baselines share suite, case, test, schema and benchmark versions, GPU models, framework image (resolved digest when recorded, otherwise the tag), model, storage configuration, model-cache mode, image-cache state, and GPU-monitoring mode. Commit and Snapshot tag are diagnostic only.
- Failed and timed-out results stay visible in the indexes but never contribute to a numeric baseline.
- Every index line carries `rawPath`, `comparisonKey`, and `result`; `e2e/tests/data/comparison-key.json` pins the key for other readers.

## Testing

- `uv run --locked --project e2e pytest e2e/tests/test_benchmark.py e2e/tests/test_benchmark_outcomes.py e2e/tests/test_benchmark_history.py e2e/tests/test_framework_manifests.py e2e/tests/test_workload_scripts.py -q` (105 passed)
- `actionlint .github/workflows/e2e-frameworks.yaml .github/workflows/ci.yml` (only the pre-existing custom runner-label warnings)

## Summary by CodeRabbit

- **New Features**
  - Added benchmark history tracking for snapshot end-to-end results.
  - Added comparison reports against compatible historical benchmarks.
  - Added automated aggregation, deduplication, index rebuilding, and publishing of benchmark data.
  - Added handling for missing, failed, retried, and incompatible benchmark results.
  - Added workflow support for scheduled publishing and read-only comparisons.

- **Documentation**
  - Documented benchmark history storage, comparison rules, result formats, and index rebuilding.

- **Tests**
  - Added comprehensive coverage for validation, persistence, comparisons, reporting, and workflow permissions.


e2e/snapshot_e2e/benchmark_history.py
```diff
@@ -0,0 +1,965 @@
...
+    for measurement in current["measurements"]:
+        comparison: dict[str, Any] = {
+            "name": measurement["name"],
+            "displayName": measurement["displayName"],
+            "unit": measurement["unit"],
+            "current": measurement["value"],
+            "previous": None,
+            "median7": None,
+        }
+        if measurement["status"] != "complete":
```
shayasoolin, line 506: `test.total.duration` is finalized even when the test fails or times out, so it remains `complete` with the elapsed time until the early exit. This loop then compares it with successful historical runs, which can make an early failure look like a large performance improvement. Could we skip comparison of `test.total.duration` unless the current outcome is `passed`, while retaining comparisons for checkpoint/restore phases that genuinely completed?

shayasoolin (dismissed):

danbar2 (approved):

PR #282, opened 2026-09-10
feat(benchmarks): add reproducible checkpoint/restore benchmark tool

eli-runai:
## Summary

Adds a repository-owned `benchmarks/` tool so users can reproduce the cold-start-vs-restore numbers published in `docs/development/benchmarks.md` against their own hardware and models, without hand-driving `kubectl` through every stage of the vLLM guide:

- new `benchmarks/` Python package (managed with `uv`), path-depending on `e2e/`'s `snapshot-e2e` package for k8s/polling plumbing instead of duplicating it
- drives one model's full cycle: cold start → checkpoint → restore, timing each stage from real cluster signals (pod conditions, the `PodSnapshot` CR's own timestamps)
- separates "Snapshot restore" from "vLLM wake and copy-to-GPU" using the node agent's `nvidia.com/Restored` pod condition as the boundary -- no changes to any guide's `app.py`
- records checkpoint size and checkpoint duration alongside the existing cold-start/restore timings
- `cli.py` exposes `run` (single model), `sweep` (all models in `models.yaml`), `metadata` (environment/cluster info dump), and `report` (renders a standalone Markdown summary from a results directory -- never auto-edits `benchmarks.md`)
- `--image-pull-policy` and `--toleration` flags for clusters whose vLLM image isn't in a registry the cluster can reach, or whose GPU nodes carry taints the guide's own manifests don't tolerate
- `models.yaml` ships the 7 models from the published table, with per-model `VLLM_GPU_MEMORY_UTILIZATION`/`VLLM_MAX_MODEL_LEN` overrides for the ones whose weights don't fit under the guide's small-single-GPU defaults
- adds `docs/development/benchmarks-guide.md` and links it from `benchmarks.md`

Related work: DEP-1185.
Issue: https://github.com/ai-dynamo/snapshot/issues/285

## Verification

- Live end-to-end runs on two real clusters: a single-GPU A10G k3s box, and an isolated namespace on a shared 8x B200 cluster -- both produced complete, sane result JSONs (cold start, checkpoint, and restore timings; checkpoint artifact size) for `Qwen/Qwen3-0.6B`
- No automated test suite is included in this PR; `snapshot-benchmarks` has no CI-run tests today, only the manual verification above (a fresh cluster + real GPU + a downloadable model are required to exercise it end to end)

## API impact

None. This is a new standalone tool with no changes to the Snapshot CRDs, agent, or Helm chart -- it only calls the existing `PodSnapshot`/`PodSnapshotContent` APIs as a client, the same way `e2e/` already does.

<h2 data-heading="Example output">Example output</h2>
<p>Run a single model:</p>
<pre><code class="language-bash">uv run --project benchmarks python -m snapshot_benchmarks run \
  --model-label "Qwen3 0.6B" --image &#x3C;registry>/vllm-snapshot:&#x3C;tag>
</code></pre>
<p>This deploys the model, checkpoints it, restores it, and writes one JSON file per model to <code>benchmarks/results/&#x3C;date>-&#x3C;git-sha>/</code> (gitignored — see <a href="vscode-webview://1tvokr2l1l1jvit5p08bml338actg718qjjhhghfjlrli3barhg1/docs/development/benchmarks-guide.md" class="external-link" target="_blank" rel="noopener nofollow" aria-label="vscode-webview://1tvokr2l1l1jvit5p08bml338actg718qjjhhghfjlrli3barhg1/docs/development/benchmarks-guide.md" data-tooltip-position="top">docs/development/benchmarks-guide.md</a> for the full walkthrough, including <code>sweep</code> for running the whole default model set).</p>
<p>Render a Markdown report from a results directory:</p>
<pre><code class="language-bash">uv run --project benchmarks python -m snapshot_benchmarks report \
  --results-dir benchmarks/results/&#x3C;date>-&#x3C;git-sha>
</code></pre>
<p>Which produces a table like this (trimmed to two models here):</p>
<h2 data-heading="Cold start vs. Snapshot restore">Cold start vs. Snapshot restore</h2>

Model | Weights (reported) | Checkpoint size (measured) | Checkpoint (s) | Cold start (s) | Restore, agent-internal (s) | Restore total, incl. vLLM wake (s) | Storage | Placement
-- | -- | -- | -- | -- | -- | -- | -- | --
Qwen3 0.6B | - | 8.45 GB | 28.000 | 110.000 | 4.721 | 7.000 | vast | same_node
Qwen3 32B | - | 82.19 GB | 113.000 | 502.000 | 18.474 | 23.000 | vast | same_node

<p>(GPU/driver columns and the Warnings section omitted here for brevity — see <a href="vscode-webview://1tvokr2l1l1jvit5p08bml338actg718qjjhhghfjlrli3barhg1/docs/development/benchmarks-guide.md#interpreting-results" class="external-link" target="_blank" rel="noopener nofollow" aria-label="vscode-webview://1tvokr2l1l1jvit5p08bml338actg718qjjhhghfjlrli3barhg1/docs/development/benchmarks-guide.md#interpreting-results" data-tooltip-position="top">docs/development/benchmarks-guide.md</a> for the full column reference and how to interpret a hardware mismatch.)</p>

## Summary by CodeRabbit

* **New Features**
  * Added a Kubernetes-based snapshot benchmark suite for cold-start and snapshot-restore measurements.
  * Added commands to collect environment metadata, run individual benchmarks or model sweeps, and generate reports.
  * Added a seven-model benchmark configuration with structured JSON result artifacts.
  * Added vLLM support with configurable images, timeouts, tolerations, cleanup, and resource retention.
  * Added detailed timing, hardware, storage, placement, and warning information to benchmark results.
* **Documentation**
  * Added setup, usage, reproducibility, timing, result interpretation, and comparison guidance for benchmark users.


benchmarks/snapshot_benchmarks/logs.py
```diff
@@ -0,0 +1,122 @@
...
+        total += float(amount) * _GO_DURATION_UNIT_SECONDS[unit]
+    return total
+
+
+def find_restore_summary_json(log_text: str) -> dict[str, Any] | None:
+    """Returns the parsed trailing JSON object of the last "Restore timing
+    summary" line in `log_text`, or None if no such line is present."""
+    summary_line = None
+    for line in log_text.splitlines():
+        if RESTORE_SUMMARY_MARKER in line:
```
oleg-kushniriov: find_restore_summary_json picks the last Restore timing summary line on the node without checking the payload's "pod" field, so on any node with concurrent or recent restores the headline agent_log_phases.duration and all phase columns can belong to someone else's pod.

You can pass the restore pod key (f"{namespace}/{name}") and/or the PodSnapshot name into parse_agent_log_phases, and select the last summary line whose parsed payload has payload.get("pod") == pod_key (fall back to matching "snapshot"). Emit a parse_warnings entry when a summary line exists but none match, so the report shows why the agent-log columns are empty rather than silently showing another run's numbers.

benchmarks/snapshot_benchmarks/run.py
```diff
@@ -0,0 +1,360 @@
...
+        schema_version=SCHEMA_VERSION,
+        run_id=run_id,
+        timestamp_utc=datetime.datetime.now(datetime.timezone.utc).isoformat(),
+        mode=mode,
+        environment=metadata.BenchmarkEnvironment(),
+        engine=BenchmarkEngine(name=engine.name),
+        model=ModelInfo(label=model.label, hf_id_or_path=model.hf_id_or_path),
+    )
+
+    # Deliberately no try/except here: on failure this raises straight
```
oleg-kushniriov: Because run_benchmark never cleans up on failure and cmd_sweep continues past the exception, a single failed model leaves a GPU pod and PodSnapshot/PodSnapshotContent behind, which on a single-GPU node makes every remaining model in the sweep time out.

If leaving a failed run's pods behind is intended as the debugging default for run, keep it there but make sweep clean up unless `--keep`. If it isn't intended, do the cleanup in run_benchmark itself via try/finally gated on keep.

benchmarks/snapshot_benchmarks/run.py
```diff
@@ -0,0 +1,360 @@
...
+def _custom_object_condition_time(obj: dict, condition_type: str) -> datetime.datetime | None:
+    cond = lifecycle.condition(obj, condition_type)
+    return _parse_iso(cond.get("lastTransitionTime")) if cond else None
+
+
+def _checkpoint_size(cfg: BenchmarkConfig, capture_node: str | None, content_uid: str) -> int | None:
+    if not capture_node:
+        return None
+    snapshot_cfg = cfg.snapshot_e2e_config()
+    agent_pod = lifecycle.checkpoint_agent_pod(snapshot_cfg, capture_node)
```
oleg-kushniriov: _checkpoint_size lets checkpoint_agent_pod's AssertionError abort the whole run after cold start and checkpoint have already succeeded, and the most common trigger is simply omitting `--snapshot-namespace`. The size should be best-effort and the agent lookup should be validated before anything is deployed.

As a solution you can add a pre-flight step at the top of run_benchmark (or in _benchmark_config) that lists snapshot agent pods in snapshot_namespace for release and fails immediately with a message naming both flags if none are found. Then make _checkpoint_size catch AssertionError and exec failures, return None, and append the reason to result.warnings so the report's Warnings section explains the missing size.

benchmarks/snapshot_benchmarks/report.py
```diff
@@ -0,0 +1,133 @@
...
+    for r in results:
+        model = r.get("model", {})
+        env = r.get("environment", {})
+        cold_start = r.get("cold_start") or {}
+        checkpoint = r.get("checkpoint") or {}
+        restore = r.get("restore") or {}
+        agent_log = r.get("agent_log_phases") or {}
+        restore_total = agent_log.get("duration")
+        if restore_total is None:
+            restore_total = restore.get("restore_total_seconds")
```
oleg-kushniriov: restore_total in report.render falls back from agent_log_phases.duration (agent-internal, excludes vLLM wake) to restore.restore_total_seconds (container start to Ready, includes wake) inside one column, so rows are not comparable and the reader cannot tell which they got.

Consider rendering both as separate columns

benchmarks/snapshot_benchmarks/cli.py
```diff
@@ -0,0 +1,276 @@
...
+
+
+def _load_models(path: Path) -> list[ModelSpec]:
+    data = yaml.safe_load(path.read_text())
+    return [
+        ModelSpec(
+            label=entry["label"],
+            hf_id_or_path=entry["hf_id_or_path"],
+            reported_weights_bytes=entry.get("reported_weights_bytes"),
+            env=entry.get("env") or {},
```
oleg-kushniriov: _load_models does not stringify env values, so an unquoted 0.75 or 2048 in models.yaml reaches the Pod spec as a number and fails at create_pod rather than at load time.

benchmarks/snapshot_benchmarks/logs.py
```diff
@@ -0,0 +1,122 @@
...
+            summary_line = line
```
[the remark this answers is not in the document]
eli-runai: @coderabbitai resolve

benchmarks/snapshot_benchmarks/metadata.py
```diff
@@ -0,0 +1,126 @@
...
+
+def gpu_identity(namespace: str, pod: str, container: str) -> tuple[str | None, str | None]:
+    """Returns (gpu_product, driver_version) queried live from inside the pod,
+    which already has GPU access via `runtimeClassName: nvidia`."""
+    output = k8s.exec_command(
+        namespace,
+        pod,
+        "nvidia-smi --query-gpu=name,driver_version --format=csv,noheader",
+        container=container,
+    )
```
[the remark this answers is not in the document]
eli-runai: @coderabbitai resolve

benchmarks/snapshot_benchmarks/results.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+def _slug(label: str) -> str:
+    return re.sub(r"[^a-z0-9]+", "-", label.lower()).strip("-")
+
+
+def invocation_dir(root: Path, *, git_sha: str | None) -> Path:
+    date = datetime.datetime.now(datetime.timezone.utc).strftime("%Y%m%d")
+    sha = git_sha or "nogit"
+    directory = root / f"{date}-{sha}"
```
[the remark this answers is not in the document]
eli-runai: @coderabbitai resolve

benchmarks/snapshot_benchmarks/run.py
```diff
@@ -0,0 +1,360 @@
...
+        model=ModelInfo(label=model.label, hf_id_or_path=model.hf_id_or_path),
+    )
+
+    # Deliberately no try/except here: on failure this raises straight
+    # through to the caller rather than returning a partially-populated
+    # `RunResult`. `cli.py`'s `sweep` command catches per-model, logs the
+    # error, and moves to the next model without writing a result file for
+    # it -- a half-filled JSON masquerading as a complete run would be worse
+    # than no file at all. Pods from a failed run are left in place (unless
+    # `keep=False` reached a cleanup step before the failure) for debugging.
```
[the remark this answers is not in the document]
eli-runai: @coderabbitai resolve

docs/development/benchmarks-guide.md
```diff
@@ -0,0 +1,131 @@
...
+- A pre-built snapshot-ready vLLM image — follow
+  [Build and deploy a vLLM replica](../guides/vllm.md) steps 1–2 first; the
+  benchmark deploys that image, it does not build it
+- A Hugging Face token/cache if the model you're benchmarking is gated
+
+## Quick start
+
+Sanity-check the cluster before committing to a multi-minute sweep — this
+prints the same environment bundle every run records, without deploying
+anything:
```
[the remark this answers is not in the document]
eli-runai: @coderabbitai resolve

benchmarks/snapshot_benchmarks/logs.py
eli-runai: thanks, already found by coderabbit

benchmarks/snapshot_benchmarks/run.py
eli-runai: thanks, already found by coderabbit

benchmarks/snapshot_benchmarks/run.py
eli-runai: Added a pre-flight check at the top of run_benchmark that verifies a snapshot-agent pod exists for the given namespace/release before deploying anything, failing fast with both flags named. Also made _checkpoint_size itself best-effort: it now catches the agent-lookup and du exec failures, returns None instead of raising, and records the reason in result.warnings (rendered by report.py's existing Warnings section)

benchmarks/snapshot_benchmarks/report.py
eli-runai: Split into two explicit columns, "Restore, agent-internal (s)" and "Restore total, incl. vLLM wake (s)", so every row always shows both and it's clear which is which.

benchmarks/snapshot_benchmarks/cli.py
eli-runai: Added _stringify_env, which coerces scalar values to str and raises a clear ValueError naming the model and key for anything else (list/dict/bool).

benchmarks/snapshot_benchmarks/run.py
```diff
@@ -0,0 +1,433 @@
...
+                _try_cleanup(lambda: k8s.delete_pod(cfg.workload_namespace, restore_name))
+            if snapshot_name is not None:
+                _try_cleanup(
+                    lambda: lifecycle.delete_podsnapshot(cfg.workload_namespace, snapshot_name)
+                )
+            if content_name is not None:
+                _try_cleanup(lambda: lifecycle.delete_podsnapshotcontent(content_name))
+            if source_created and not source_deleted:
+                progress(f"[{run_id}] cleaning up source pod")
+                _try_cleanup(lambda: k8s.delete_pod(cfg.workload_namespace, source_name))
```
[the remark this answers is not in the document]
eli-runai: @coderabbitai resolve

benchmarks/snapshot_benchmarks/cli.py
```diff
@@ -0,0 +1,295 @@
...
+    malformed value reach `create_pod` and fail deep inside a Kubernetes API
+    JSON-unmarshal error."""
+    result: dict[str, str] = {}
+    for key, value in (env or {}).items():
+        if isinstance(value, bool) or not isinstance(value, (str, int, float)):
+            raise ValueError(
+                f"model {label!r}: env[{key!r}] must be a plain string/number, got "
+                f"{type(value).__name__} ({value!r}) -- quote it in models.yaml"
+            )
+        result[key] = str(value)
```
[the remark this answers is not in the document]
eli-runai: @coderabbitai resolve

benchmarks/snapshot_benchmarks/run.py
```diff
@@ -0,0 +1,439 @@
...
+        )
+
+        capture_node = source_pod.spec.node_name
+        result.environment = metadata.collect_environment(
+            namespace=cfg.workload_namespace,
+            pvc_name=cfg.pvc_name,
+            gpu_pod=source_name,
+            gpu_container=engine.container_name,
+            capture_node=capture_node,
+        )
```
hhzhang16, line 248: The benchmark supports separate workload and Snapshot namespaces, but `collect_environment` receives `cfg.workload_namespace` and uses that same namespace for the checkpoint PVC lookup. With the documented default/snapshot split, `storage_backend` would silently return null metadata for the PVC mounted by the Snapshot release. Would be best to split the workload namespace used for pod exec from the Snapshot namespace used for PVC metadata.

benchmarks/snapshot_benchmarks/run.py
```diff
@@ -0,0 +1,439 @@
...
+            result.restore = RestoreTiming(
+                restore_pod_created_at=restore_created_pod.metadata.creation_timestamp,
+                restore_container_started_at=restore_container_started_at,
+                restored_condition_at=_pod_condition_time(restored_pod, "nvidia.com/Restored"),
+                restore_ready_at=_pod_condition_time(restore_ready_pod, "Ready"),
+            )
+
+            restore_node = restore_ready_pod.spec.node_name
+            result.environment.restore_node = restore_node
+            result.environment.placement = metadata.placement(capture_node, restore_node)
```
hhzhang16, line 327: Only `restore_node` and `placement` are updated after restoration which means that `gpu_product`, `gpu_driver_version`, and `cuda_driver_major_label` still describe the capture pod. For a different-node/heterogenous restore, the report can therefore attribute restore timing to the wrong hardware, so capture and restore identities should be collected + serialized separately

benchmarks/snapshot_benchmarks/run.py
```diff
@@ -0,0 +1,439 @@
...
+
+def _try_cleanup(action: Callable[[], None]) -> None:
+    """Runs a best-effort cleanup step, swallowing and printing any exception
+    rather than letting it propagate -- a cleanup failure (e.g. the pod was
+    already gone) must never shadow the run's own real exception when both
+    happen inside the same `finally` block."""
+    try:
+        action()
+    except Exception as exc:  # noqa: BLE001 - deliberately broad, see docstring
+        print(f"warning: cleanup step failed: {exc}")
```
hhzhang16: Preserve the primary benchmark exception, but propagate or otherwise signal cleanup failure so a sweep does not start another model on contaminated cluster state

benchmarks/snapshot_benchmarks/results.py
```diff
@@ -0,0 +1,57 @@
...
+    sha = git_sha or "nogit"
+    directory = root / f"{date}-{sha}-{uuid.uuid4().hex[:8]}"
+    directory.mkdir(parents=True)
+    return directory
+
+
+def write_result(result: RunResult, directory: Path) -> Path:
+    path = directory / f"{_slug(result.model.label)}-{result.mode}.json"
+    path.write_text(json.dumps(result.to_json_dict(), indent=2, sort_keys=False) + "\n")
+    return path
```
hhzhang16, line 55: Filenames use only a normalized label and mode. I reproduced "Foo Bar" and "Foo-Bar" producing the same path; the second result replaced the first. Include `run_id` or reject duplicate/colliding labels.

benchmarks/snapshot_benchmarks/run.py
eli-runai: Fixed — collect_environment now takes pvc_namespace explicitly , and both call sites (run.py, cli.py's cmd_metadata) now pass cfg.snapshot_namespace, matching where the checkpoint PVC actually lives per the Snapshot Helm chart's storage.pvc.name.

benchmarks/snapshot_benchmarks/run.py
eli-runai: fixed

benchmarks/snapshot_benchmarks/run.py
eli-runai: try_cleanup now returns the caught exception instead of only printing it.
run_benchmark's finally block collects all cleanup exceptions and, if any occurred, raises a new CleanupError — chained via raise ... from primary_exc so the original benchmark exception (if any) is preserved both in the message text and the traceback (__cause__), never silently masked.

benchmarks/snapshot_benchmarks/results.py
eli-runai: Fixed. write_result now appends run_id to the filename

benchmarks/snapshot_benchmarks/report.py
```diff
@@ -0,0 +1,150 @@
...
+        "| Model | Weights (reported) | Checkpoint size (measured) | Checkpoint (s) | "
+        "Cold start (s) | Restore, agent-internal (s) | Restore total, incl. vLLM wake (s) | "
+        "GPU (capture) | Driver (capture) | GPU (restore) | Driver (restore) | Storage | Placement |"
+    )
+    lines.append("|---|---:|---:|---:|---:|---:|---:|---|---|---|---|---|---|")
+    for r in results:
+        model = r.get("model", {})
+        env = r.get("environment", {})
+        capture_gpu = env.get("capture") or {}
+        restore_gpu = env.get("restore") or {}
```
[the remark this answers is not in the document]
eli-runai: @coderabbitai resolve

benchmarks/snapshot_benchmarks/run.py
```diff
@@ -0,0 +1,512 @@
...
+    if capture.gpu_product and restore.gpu_product and capture.gpu_product != restore.gpu_product:
+        mismatches.append(f"gpu_product {capture.gpu_product!r} -> {restore.gpu_product!r}")
+    if (
+        capture.gpu_driver_version
+        and restore.gpu_driver_version
+        and capture.gpu_driver_version != restore.gpu_driver_version
+    ):
+        mismatches.append(
+            f"gpu_driver_version {capture.gpu_driver_version!r} -> {restore.gpu_driver_version!r}"
+        )
```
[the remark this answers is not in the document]
eli-runai: @coderabbitai resolve

benchmarks/snapshot_benchmarks/report.py
```diff
@@ -0,0 +1,211 @@
...
+def _normalize_result(result: dict) -> dict:
+    """Upgrades a raw `RunResult` dict to the shape `render` expects, or raises
+    for a `schema_version` this report renderer doesn't know how to read --
+    silently rendering an unsupported/unrecognized shape would display missing
+    GPU metadata as "no data" rather than "can't read this file", masking a
+    real schema mismatch in a directory of mixed-version results."""
+    version = result.get("schema_version")
+    if version == 2:
+        return result
+    if version == 1:
```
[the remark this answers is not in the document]
eli-runai: @coderabbitai resolve

docs/development/benchmarks-guide.md
```diff
@@ -0,0 +1,133 @@
...
+Only vLLM is implemented today ("the agreed day-one vLLM comparison" scope).
+SGLang and TensorRT-LLM can be added later without redesigning the tool — see
+`benchmarks/snapshot_benchmarks/engines/__init__.py`.
+
+## Prerequisites
+
+- [`uv`](https://docs.astral.sh/uv/)
+- `kubectl`
+- A Snapshot Helm release already [installed](../operations/install.md) and
+  reachable via `KUBECONFIG`
```
dfeigin-nv, line 18: is pvc part of the prereqs or is it part of the helm installation?

docs/development/benchmarks-guide.md
```diff
@@ -0,0 +1,133 @@
...
+- A GPU-schedulable node with `RuntimeClass/nvidia` (see the repo
+  [Prerequisites](../../README.md#prerequisites))
+- A pre-built snapshot-ready vLLM image — follow
+  [Build and deploy a vLLM replica](../guides/vllm.md) steps 1–2 first; the
+  benchmark deploys that image, it does not build it
+- A Hugging Face token/cache if the model you're benchmarking is gated
```
dfeigin-nv: `- A Hugging Face token/cache if the model you're benchmarking is gated`
unnecessary. Some folks load it from disk

docs/development/benchmarks-guide.md
```diff
@@ -0,0 +1,133 @@
...
+prints/writes a separate summary for you to read or paste from by hand.
+
+## What gets measured
+
+Every run separates two things the published doc reports together as "wake /
+remap": **Snapshot's own restore work** and **the engine's own wake and
+copy-to-GPU work**. The boundary is the node agent's `nvidia.com/Restored` pod
+condition, which fires the moment Snapshot believes the restore is done —
+before vLLM's own `wake_up()`/`resume_generation()`/warmup sequence runs. So,
+in each run's JSON:
```
dfeigin-nv: what JSON?

docs/development/benchmarks-guide.md
```diff
@@ -0,0 +1,133 @@
...
+These come from Kubernetes condition timestamps, which have one-second
+resolution — a large relative error for the smallest models. For the headline
+total, `agent_log_phases.duration` is parsed from the node agent's own restore
+log, which has sub-second precision and is what actually produced the
+published numbers. `agent_log_phases` also includes a best-effort remap of the
+agent's own phase names to the published doc's 4-stage vocabulary
+(`agent_setup_approx`, `criu_restore_approx`, `cuda_restore_approx`) — these
+are approximate by construction; `cuda_restore_approx` is the one exact 1:1
+mapping. `wake_remap_approx` is always `null`: that published stage isn't
+represented in the agent log at all, use `vllm_wake_and_copy_seconds` instead.
```
dfeigin-nv: who is the 'use' directed at?

docs/development/benchmarks-guide.md
```diff
@@ -0,0 +1,133 @@
...
+it can be excluded to match the published doc's "Cold start" column.
+
+Checkpoint size (`model.checkpoint_artifact_bytes`) is measured with `du -sb`
+on the capture node — neither `PodSnapshot` nor `PodSnapshotContent` status
+exposes a size field.
+
+## Interpreting results
+
+The published numbers were measured on a single B200 (driver 595) with a VAST
+NFS PVC — see [Storage throughput sets the floor](benchmarks.md#storage-throughput-sets-the-floor)
```
dfeigin-nv: with direct_io enabled on the pvc. It will save some back and forths on speed

benchmarks/snapshot_benchmarks/run.py
```diff
@@ -0,0 +1,527 @@
...
+    source_name = f"bench-source-{run_id}"
+
+    result = RunResult(
+        schema_version=SCHEMA_VERSION,
+        run_id=run_id,
+        timestamp_utc=datetime.datetime.now(datetime.timezone.utc).isoformat(),
+        mode=mode,
+        environment=metadata.BenchmarkEnvironment(),
+        engine=BenchmarkEngine(name=engine.name),
+        model=ModelInfo(label=model.label, hf_id_or_path=model.hf_id_or_path),
```
dfeigin-nv: Reported weights disappear: ModelInfo does not copy model.reported_weights_bytes, so JSON contains null and reports show -.

docs/development/benchmarks-guide.md
```diff
@@ -0,0 +1,133 @@
...
+
+Every run separates two things the published doc reports together as "wake /
+remap": **Snapshot's own restore work** and **the engine's own wake and
+copy-to-GPU work**. The boundary is the node agent's `nvidia.com/Restored` pod
+condition, which fires the moment Snapshot believes the restore is done —
+before vLLM's own `wake_up()`/`resume_generation()`/warmup sequence runs. So,
+in each run's JSON:
+
+- `restore.snapshot_restore_seconds` — container start → `nvidia.com/Restored`
+- `restore.vllm_wake_and_copy_seconds` — `nvidia.com/Restored` → pod `Ready`
```
dfeigin-nv: i think it's called wake_and_copy_seconds in the code

benchmarks/README.md
```diff
@@ -0,0 +1,54 @@
+# Snapshot Benchmarks
+
+This directory contains the Python tool that reproduces the numbers published in
+[Performance Benchmarks](../docs/development/benchmarks.md): deploy a
+snapshot-ready vLLM replica, time cold start, checkpoint it, restore it, time
+the restore, and record everything (including per-run environment metadata —
```
dfeigin-nv, line 6: do you want to add an automated test to the benchmark so we'll know when it breaks?

benchmarks/snapshot_benchmarks/schema.py
```diff
@@ -0,0 +1,266 @@
...
+    if isinstance(value, datetime.datetime):
+        return _iso(value)
+    if isinstance(value, list):
+        return [_to_json_value(item) for item in value]
+    if isinstance(value, dict):
+        return {key: _to_json_value(item) for key, item in value.items()}
+    return value
+
+
+SCHEMA_VERSION = 2
```
dfeigin-nv: we're already at 2? and what do we need a schema here? where's the contract?

benchmarks/snapshot_benchmarks/cli.py
```diff
@@ -0,0 +1,306 @@
...
+        image_pull_policy=args.image_pull_policy,
+        tolerations=args.toleration,
+        mode=args.mode,
+        keep=args.keep,
+        timeout=args.timeout,
+    )
+
+
+def cmd_sweep(args: argparse.Namespace) -> int:
+    cfg = _benchmark_config(args)
```
dfeigin-nv: cmd run and sweep are a bit duplicated no?

benchmarks/snapshot_benchmarks/run.py
```diff
@@ -0,0 +1,527 @@
...
+            )
+            if heterogeneous_warning:
+                result.warnings.append(heterogeneous_warning)
+
+            result.agent_log_phases = _collect_agent_log_phases(
+                cfg, restore_node, restore_name=restore_name, snapshot_name=snapshot_name
+            )
+
+        return result
+    finally:
```
dfeigin-nv, line 347: a suggestion: whenever a resource is created, immediately register how to clean it up.

Conceptually:

```
cleanup_actions = []
 
create_source_pod()
cleanup_actions.append(delete_and_wait_for_source_pod)
 
create_snapshot()
cleanup_actions.append(delete_snapshot_and_content)
 
create_restore_pod()
cleanup_actions.append(delete_and_wait_for_restore_pod)
 
finally:
    for cleanup in reversed(cleanup_actions):
        cleanup()
```

dfeigin-nv:
Can you add an exmaple table in the pr description or point to one?

docs/development/benchmarks-guide.md
eli-runai: The checkpoint PVC is part of the Helm installation — the chart provisions a ReadWriteMany PVC by default (snapshot-pvc, matching this tool's default pvc_name). It's not a separate prereq for the benchmarks tool; see [install.md](vscode-webview://1tvokr2l1l1jvit5p08bml338actg718qjjhhghfjlrli3barhg1/operations/install.md#install-from-a-release). I've updated the Requirements section to call that out explicitly.

docs/development/benchmarks-guide.md
eli-runai: Fair point, dropped it — not everyone needs a token since some load models from local disk/cache instead

docs/development/benchmarks-guide.md
eli-runai: Clarified - it's the per-model result JSON written to benchmarks/results/ (mentioned in Quick start above), added an explicit reference.

docs/development/benchmarks-guide.md
eli-runai: Yes, it was ambiguous - clarified it's for anyone trying to reproduce the published 'wake / remap' number: they should read vllm_wake_and_copy_seconds instead.

docs/development/benchmarks-guide.md
eli-runai: Added - direct_io affects storage throughput enough that it's worth calling out upfront

benchmarks/snapshot_benchmarks/run.py
eli-runai: Yeah fixed - it was silently dropped when building ModelInfo.

docs/development/benchmarks-guide.md
eli-runai: Right, fixed both spots - it's wake_and_copy_seconds in the actual schema.

benchmarks/README.md
eli-runai: good idea. we'll deal with that in another pr.

benchmarks/snapshot_benchmarks/schema.py
eli-runai: oh, a mistake. fixed

benchmarks/snapshot_benchmarks/cli.py
eli-runai: Extracted the shared config/engine setup into a _setup() helper used by both run and sweep. Left the result-writing step separate since sweep needs to memoize out_dir across iterations while run only writes once; merging those would add indirection without removing real duplication.

benchmarks/snapshot_benchmarks/run.py
eli-runai: good suggestion. will defer this to a nother pr

dfeigin-nv (dismissed):

benchmarks/snapshot_benchmarks/logs.py
```diff
@@ -0,0 +1,135 @@
...
+    diagnostic parse, never a required value."""
+    if not value:
+        return None
+    matches = _GO_DURATION_TERM.findall(value)
+    if not matches:
+        return None
+    total = 0.0
+    for amount, unit in matches:
+        total += float(amount) * _GO_DURATION_UNIT_SECONDS[unit]
+    return total
```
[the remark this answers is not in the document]
eli-runai: @coderabbitai resolve

benchmarks/snapshot_benchmarks/schema.py
```diff
@@ -0,0 +1,266 @@
...
+    criu_prepare: float | None = None
+    criu_restore: float | None = None
+    cuda_restore: float | None = None
+    unaccounted: float | None = None
+
+    @property
+    def agent_setup_approx(self) -> float | None:
+        parts = [self.gpu_device_map, self.pagebroker_stage, self.pagebroker_mount, self.pagebroker_commit]
+        present = [p for p in parts if p is not None]
+        return sum(present) if present else None
```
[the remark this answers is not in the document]
eli-runai: @coderabbitai resolve

docs/development/benchmarks-guide.md
```diff
@@ -0,0 +1,134 @@
...
+the comments in that file about tuning `VLLM_MAX_MODEL_LEN` /
+`VLLM_GPU_MEMORY_UTILIZATION` per model for your GPU's VRAM):
+
+```bash
+uv run --project benchmarks python -m snapshot_benchmarks sweep \
+  --snapshot-namespace <namespace-snapshot-is-installed-in> \
+  --image <registry>/vllm-snapshot:<tag>
+```
+
+Each run writes one JSON file per model to `benchmarks/results/<date>-<git-sha>/`
```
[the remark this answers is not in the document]
eli-runai: @coderabbitai resolve

dfeigin-nv (approved):

danbar2 (approved):

PR #284, opened 2026-09-10
build: pin the pagebroker base images and cover the e2e Python deps

### LINKED ISSUE #270: Harden CI supply chain: SHA-pinned actions, CodeQL, Dependabot coverage
...

danbar2:
Two gaps found by running the OpenSSF Scorecard CLI against this repo locally, with **nothing published** — `publish_results` was never involved, the scan ran from a workstation.

Fixes https://github.com/ai-dynamo/snapshot/issues/270

## `agent/pagebroker/Dockerfile` was missed

There is a **third** Dockerfile in the tree, and #277 only pinned the two I knew about. Both of its `ubuntu:24.04` stages were on a mutable tag. They now carry the same digest as the CRIU builder in `agent/Dockerfile`, which is also the digest Scorecard itself suggested.

That takes `Pinned-Dependencies` from **9/10 to 10/10**. The only unpinned `FROM` left is the placeholder stage's `${BASE_IMAGE}`, which is caller-supplied and cannot be pinned here.

## `e2e/` had no Dependabot coverage at all

It is a uv-managed Python project with its own lockfile, and no ecosystem entry covered it — so nothing has ever proposed an update for it. Scorecard reports a known **pytest** advisory (`PYSEC-2026-1845`) sitting there.

Worth being explicit about why nothing caught this: `govulncheck` only covers Go, and only *reachable* symbols. A vulnerable Python test dependency is outside everything this repo currently runs.

## Validation

`make check` run in `golang:1.27.1` on amd64 — exit 0, tree clean.

## Summary by CodeRabbit

- **Chores**
  - Added weekly dependency update monitoring for the end-to-end testing project, including grouped minor and patch updates with a seven-day cooldown.
  - Updated Go dependencies across agent and operator components.
  - Pinned container build stages to an immutable Ubuntu 24.04 image digest for more consistent, reproducible builds.

- **Documentation**
  - Added guidance on recommended development patterns, logging, generated resources, image pinning, and toolchain updates.


Ronkahn21 (dismissed):

danbar2:
### Added: the known vulnerabilities, and concrete examples for agents

**Vulnerabilities.** The Scorecard run reported six advisories (`Vulnerabilities 4/10` — one point above the threshold at which `SIG-SEC-06` flips to a failure). All five Go ones are fixed:

| Module | From → To | Advisories cleared |
| --- | --- | --- |
| `google.golang.org/grpc` | 1.82.1 → **1.83.2** | GHSA-vp52-pcj8-j9qc **HIGH**, GHSA-2v4p-qf9q-27wj **HIGH**, GHSA-qc2q-p7wx-3px3 MODERATE |
| `github.com/containerd/containerd/v2` | 2.0.11 → **2.0.12** | GHSA-7jxh-36q5-gcqv MODERATE |
| `github.com/klauspost/compress` | 1.18.0 → **1.18.7** | GO-2026-5841 |

Verified with `osv-scanner`: **api, agent and operator now report no issues.**

**Two deliberate non-fixes**, both flagged rather than forced:

- `containerd` v1.7.23 is below its fixed 1.7.35, reached only via `Microsoft/hcsshim`. No package from it compiles into anything here — `go list -deps` returns none, and `go mod tidy` drops an explicit require because nothing imports it. Forcing it means carrying a require the tooling deletes on every tidy, for a module that is not built.
- `pytest` 8.4.2 (`PYSEC-2026-1845`, **MEDIUM**, 6.8) needs ≥ 9.0.3, but `e2e/pyproject.toml` pins `pytest>=8.4.0,<9`. Clearing it means breaking a deliberate major-version ceiling on a test-only dependency. The `uv` ecosystem added in this PR means Dependabot will now propose it properly, for a human to weigh.

Worth noting **why `make check` never caught any of this**: `govulncheck` reports only *reachable* symbols, and every one of these sits in the dependency graph without being called. That is the gap between what CI enforces and what a manifest-level scanner sees.


**Validation:** `make check`, `make build` and `make test` all exit 0 in `golang:1.27.1` on amd64; tree clean.

Ronkahn21 (approved):

Issue #245, opened 2026-09-08
Simplify agent capture flow using a work queue


danbar2:
remove all the agent usage of lease for locking and sync

PR #178, opened 2026-09-02
fix(agent): bound snapshot content reconciliation

galletas1712:
Signed-off-by: Schwinn Saereesitthipitak <schwinns@nvidia.com>

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

PR #291, opened 2026-09-10
fix(agent): prevent snapshot content informer blocking

### LINKED ISSUE #302: agent: a stalled PodSnapshotContent reconciliation blocks all later content events (informer head-of-line blocking)
### What happened?

In the node-agent, the dynamic PodSnapshotContent informer ran `reconcilePodSnapshotContent` synchronously inside its `Add`/`Update` event handlers (`agent/internal/controller/controller.go` on current `main`). That reconciliation performs live typed reads and writes through an uncached controller-runtime client, so the handlers block on API-server I/O.

As a result, a single PodSnapshotContent reconciliation that stalls on a live request blocks the informer goroutine that delivers content events. While one work order's reconciliation is stalled:

- later PodSnapshotContent events are not delivered at all,
- later work orders on the same node never enter reconciliation, and
- their captures are stuck behind the stalled one (head-of-line blocking).

This reproduces deterministically in the controller unit tests with a controlled live-Get barrier: before the fix, one stalled content reconciliation prevented later content from entering reconciliation and capture.

We suspect a variant of this stalled capture progress in a vCluster-based [Dynamo CI job](https://github.com/ai-dynamo/dynamo/actions/runs/34444478883/job/102771685192?pr=14105) where live content reads kept hitting object-modified conflicts, although the local reproduction does not prove that specific conflict was the production trigger.

### What did you expect to happen?

A stalled or slow reconciliation of one PodSnapshotContent must not prevent unrelated PodSnapshotContent work orders on the same node from being reconciled and captured.

### How can we reproduce it?

Reproduced in node-agent controller unit tests; no cluster, GPU, or CRIU needed. The reproduction (test `TestNodeController_ContentInformerDoesNotBlockLaterWork` in `agent/internal/controller/controller_test.go`) installs a barrier that stalls the live Get for one content's reconciliation and then asserts a second content still gets reconciled:

- With the synchronous informer handlers on `main`, the second content never enters reconciliation while the first is stalled — the test fails.
- With informer handlers that only enqueue onto a work queue drained by multiple workers, the second content reconciles while the first is stalled — the test passes.

### Environment

- Snapshot version: `main` at 6725e1b (agent built from source)
- Reproduced in `agent/internal/controller` unit tests on Linux/arm64, Go 1.26
- Suspected production sighting: Dynamo CI on vCluster (PodSnapshotContent object-modified conflicts during live reads)

### Relevant logs, events, or resource YAML

See the linked Dynamo CI job above; the unit-test reproduction needs no logs.

Fix proposed in #291: move PodSnapshotContent informer delivery onto a deduplicating work queue drained by two workers, keeping informer callbacks limited to cache-to-queue work.

carlory:
## Summary

Fixes #302

This change deterministically reproduces node-agent PodSnapshotContent informer head-of-line blocking with a controlled live-Get barrier: before the fix, one stalled content reconciliation prevented later content from entering reconciliation and capture. The minimal production fix moves PodSnapshotContent informer events onto a deduplicating queue drained by two workers, while retaining live reads as the authoritative reconciliation state.

This also contributes to #245 (simplify the agent capture flow using a work queue): PodSnapshotContent event delivery now goes through a deduplicating work queue drained by two workers instead of synchronous informer handlers. The rest of #245's scope — replacing lease-based locking/sync and the synchronous source-pod informer — is untouched.

The local evidence proves the one-stall head-of-line blocking behavior and its correction. It does not prove that the vCluster object-modified conflict seen in the [failing Dynamo CI job](https://github.com/ai-dynamo/dynamo/actions/runs/34444478883/job/102771685192?pr=14105) was the exact production trigger. This work does not claim a full vCluster, GPU, or CRIU end-to-end reproduction.

[ai-dynamo/snapshot PR 178](https://github.com/ai-dynamo/snapshot/pull/178) was a closed draft measurement-infrastructure follow-up, not an already merged fix.

The fix has deliberately bounded scope:

- Two workers guarantee progress for the reproduced one-stall case.
- When all workers are stalled, later work orders wait in the queue until a worker frees.
- The typed client has no controller-specific request timeout.
- The separate source-pod informer remains synchronous and is outside this content-informer fix.

## Validation

Validated on the exact final tree using the running Colima Linux/aarch64 runtime, Go 1.26.6, a persistent build cache, and the host module cache mounted read-only:

- `gofmt` on all three changed files and `git diff --check`: passed with no formatting change or whitespace error.
- Native Linux/arm64 `go test -mod=readonly -count=1 ./internal/controller`: passed in 0.678s.
- `TestNodeController_ContentInformerDoesNotBlockLaterWork -count=100`: passed in 32.675s.
- The same focused test with `-race -count=10`: passed in 5.491s with no race.
- Linux/amd64 `go test -mod=readonly -c ./internal/controller` cross-compilation: passed; output was written outside the repository.
- no-mistakes AXI run `01M2606X3XM0WMMPY8MTN1JBEY`: reached `checks-passed`. Automated review reported only the two independently accepted informational no-op scope limitations above; test, documentation, and lint stages passed with zero findings and no pipeline-authored repository or commit change; all visible CI checks passed.

### DIFF
--- a/agent/internal/controller/controller.go
+++ b/agent/internal/controller/controller.go
@@ -53,9 +53,9 @@ import (
 
 // NodeController watches local-node pods with checkpoint metadata and reconciles
 // snapshot execution for checkpoint and restore requests. The restore path is
-// driven by a client-go pod informer; the capture path is driven by a dynamic
-// informer over PodSnapshotContent work orders filtered to this node, with typed
-// reads/writes via an uncached controller-runtime client.
+// driven by a client-go pod informer; PodSnapshotContent events feed content
+// workqueue workers, which perform typed reads/writes via an uncached controller-runtime
+// client.
 type NodeController struct {
 	config                  *types.AgentConfig
 	clientset               kubernetes.Interface
@@ -70,6 +70,7 @@ type NodeController struct {
 	writeControlSentinelFn  func(int, string) error
 	controlSentinelExistsFn func(int, string) (bool, error)
 	sendSignalFn            func(logr.Logger, int, syscall.Signal, string) error
+	contentQueue            workqueue.TypedDelayingInterface[string]
 	restoreQueue            workqueue.TypedDelayingInterface[client.ObjectKey]
 	restorePodLister        corev1listers.PodLister
 
@@ -153,6 +154,10 @@ const (
 	// snapshotContentResyncInterval re-drives every PodSnapshotContent work order so a
 	// not-yet-Ready source pod is re-checked for quiesce without a busy loop.
 	snapshotContentResyncInterval = 10 * time.Second
+	// contentWorkerCount keeps informer delivery independent of live API calls while allowing
+	// unrelated work orders to make progress when one reconciliation is stalled.
+	// When all workers are stalled, later work orders wait in the queue until a worker frees.
+	contentWorkerCount = 2
 )
 
 // podSnapshotContentGVR is the cluster-scoped resource the capture informer watches.
@@ -212,6 +217,9 @@ func newDefaultController(
 		holderID:  "snapshot-agent/" + uuid.NewString(),
 		inFlight:  make(map[string]struct{}),
 		stopCh:    make(chan struct{}),
+		contentQueue: workqueue.NewTypedDelayingQueueWithConfig(
+			workqueue.TypedDelayingQueueConfig[string]{Name: "snapshot-contents"},
+		),
 		restoreQueue: workqueue.NewTypedDelayingQueueWithConfig(
 			workqueue.TypedDelayingQueueConfig[client.ObjectKey]{Name: "restore-pods"},
 		),
@@ -227,6 +235,7 @@ func newDefaultController(
 
 // Run starts the local pod informers and processes checkpoint/restore events.
 func (w *NodeController) Run(ctx context.Context) error {
+	defer w.contentQueue.ShutDown()
 	defer w.restoreQueue.ShutDown()
 	// Seed the agent logger onto ctx so the capture path resolves it via log.FromContext.
 	ctx = logr.NewContext(ctx, w.log)
@@ -280,15 +289,9 @@ func (w *NodeController) Run(ctx context.Context) error {
 	}
 	w.contentIndexer = contentInformer.GetIndexer()
 	if _, err := contentInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
-		AddFunc: func(obj interface{}) {
-			if name, ok := contentNameFromInformerObj(obj); ok {
-				w.reconcilePodSnapshotContent(ctx, name)
-			}
-		},
+		AddFunc: w.enqueuePodSnapshotContent,
 		UpdateFunc: func(_, newObj interface{}) {
-			if name, ok := contentNameFromInformerObj(newObj); ok {
-				w.reconcilePodSnapshotContent(ctx, name)
-			}
+			w.enqueuePodSnapshotContent(newObj)
 		},
 	}); err != nil {
 		return fmt.Errorf("failed to add snapshot-content informer handler: %w", err)
@@ -337,6 +340,7 @@ func (w *NodeController) Run(ctx context.Context) error {
 	var stopOnce sync.Once
 	go func() {
 		<-ctx.Done()
+		w.contentQueue.ShutDown()
 		w.restoreQueue.ShutDown()
 		stopOnce.Do(func() { close(w.stopCh) })
 	}()
@@ -345,11 +349,56 @@ func (w *NodeController) Run(ctx context.Context) error {
 		return fmt.Errorf("failed to sync informer caches")
 	}
 
+	for i := 0; i < contentWorkerCount; i++ {
+		go w.runContentQueue(ctx)
+	}
 	go w.runRestoreQueue(ctx)
 	w.log.Info("PodSnapshot node controller started and caches synced")
 	<-ctx.Done()
 	stopOnce.Do(func() { close(w.stopCh) })
 	return nil
+}
+
+// enqueuePodSnapshotContent keeps informer callbacks limited to cache-to-queue work. The live
+// client calls happen in runContentQueue workers, so one stalled reconciliation cannot block later
+// informer notifications. Filtering here also keeps terminal or foreign work out of the queue.
+func (w *NodeController) enqueuePodSnapshotContent(obj interface{}) {
+	content, ok := contentFromInformerObj(obj)
+	if !ok {
+		if name, nameOK := contentNameFromInformerObj(obj); nameOK {
+			w.log.Error(errors.New("failed to convert PodSnapshotContent informer object"),
+				"Queuing PodSnapshotContent after informer conversion failure", "content", name)
+			w.contentQueue.Add(name)
+		} else {
+			w.log.Error(errors.New("failed to identify PodSnapshotContent informer object"),
+				"Dropping unidentified PodSnapshotContent informer object")
+		}
+		return
+	}
+	if content.Spec.Source.NodeName != w.config.NodeName || !contentActionable(content) {
+		return
+	}
+	w.contentQueue.Add(content.Name)
+}
+
+// contentActionable reports whether a PodSnapshotContent can still be reconciled.
+func contentActionable(content *snapshotv1alpha1.PodSnapshotContent) bool {
+	return content.DeletionTimestamp.IsZero() && !isContentTerminal(content)
+}
+
+func (w *NodeController) runContentQueue(ctx context.Context) {
+	for {
+		name, shutdown := w.contentQueue.Get()
+		if shutdown {
+			return
+		}
+		w.processContentQueueItem(ctx, name)
+	}
+}
+
+func (w *NodeController) processContentQueueItem(ctx context.Context, name string) {
+	defer w.contentQueue.Done(name)
+	w.reconcilePodSnapshotContent(ctx, name)
 }
 
 func tweakNodePodListOptions(nodeName string) func(*metav1.ListOptions) {
@@ -1111,6 +1160,19 @@ func contentFromInformerObj(obj interface{}) (*snapshotv1alpha1.PodSnapshotConte
 	return content, true
 }
 
+// contentNameFromInformerObj extracts the name used to re-drive a work order when its cached
+// representation cannot be converted. The live typed read in the worker remains authoritative.
+func contentNameFromInformerObj(obj interface{}) (string, bool) {
+	if tombstone, ok := obj.(cache.DeletedFinalStateUnknown); ok {
+		obj = tombstone.Obj
+	}
+	accessor, err := meta.Accessor(obj)
+	if err != nil || accessor.GetName() == "" {
+		return "", false
+	}
+	return accessor.GetName(), true
+}
+
 // chooseActiveContent returns the name of the oldest non-terminal PodSnapshotContent among the indexed
 // objects (oldest first by CreationTimestamp, ties broken by Name), or "" when none are active.
 // Driving the oldest until it finishes gives deterministic, stable selection across pod events.
@@ -1118,7 +1180,7 @@ func chooseActiveContent(objs []interface{}) string {
 	var chosen *snapshotv1alpha1.PodSnapshotContent
 	for _, obj := range objs {
 		content, ok := contentFromInformerObj(obj)
-		if !ok || !content.DeletionTimestamp.IsZero() || isContentTerminal(content) {
+		if !ok || !contentActionable(content) {
 			continue
 		}
 		if chosen == nil ||
--- a/agent/internal/controller/controller_test.go
+++ b/agent/internal/controller/controller_test.go
@@ -11,6 +11,7 @@ import (
 	"reflect"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"syscall"
 	"testing"
 	"time"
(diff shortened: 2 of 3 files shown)


agent/internal/controller/controller_test.go
```diff
@@ -112,12 +116,363 @@ func TestNewDefaultControllerSetsDefaultOperations(t *testing.T) {
...
+	t.Run("released reconciliation allows later content", func(t *testing.T) {
+		h := newContentInformerBarrierHarness(t, false)
+		h.deliver(informerRaceContent("content-a", "snapshot-a"))
+		h.requireBarrier(h.contentAGet, "content A did not enter live reconciliation")
+		closeBarrier(h.releaseContentA)
+
+		h.deliver(informerRaceContent("content-b", "snapshot-b"))
+		h.requireBarrier(h.contentBGet,
+			"content B did not reach the live Get boundary after content A reconciliation was released")
+		h.assertNoCaptureEvidence("content-b")
```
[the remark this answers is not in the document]
carlory: This is a false positive for the current harness.

After `releaseContentA` is closed, the intercepted initial `PodSnapshotContent` Get for `content-a` falls through and returns `errors.New("injected live content Get failure")`. `reconcilePodSnapshotContent` returns immediately on that non-NotFound content Get error, before it can look up the source pod, call `failContentFromGate`, access a lease, or invoke the checkpoint seam. Retries follow the same interceptor path. Meanwhile, `content-b` is held at its first live Get while `assertNoCaptureEvidence("content-b")` runs.

Therefore `content-a` cannot increment `contentStatusWrites` or `checkpointCalls`, or add lease actions in these non-capture subtests, so per-content counters are not needed for the claimed race. The exact final test also passed `-count=100` and `-race -count=10`.

hhzhang16 (approved):

carlory:
The `Validate Issue Reference` check was failing because the description's only `#NNN` reference pointed at PR 178 (a closed draft PR), not an issue. I've filed the tracking issue for the bug this PR fixes and linked it from the description (`Fixes #302`), and reworded the PR 178 mention so it no longer parses as an issue reference.

The check can now go green once a maintainer either labels #302 `approved` or waives this PR with `skip-issue-check` — it re-runs on its own when the label lands. The latest workflow run from this description edit is also waiting on the usual fork-PR workflow approval.

carlory:
The `PR links an approved issue` check failed because the description originally referenced #178, which is a pull request. The description now links **#302** (the bug this fixes) and also references **#245** — the approved work-queue issue this change partially advances (content-informer delivery only; lease removal and the source-pod informer stay out of scope, so #245 is deliberately not marked fixed).

The re-run triggered by that edit (run 34877641578) is awaiting workflow approval for this fork PR. Maintainers: please approve it (or label #302 `approved` / add `skip-issue-check` if you prefer either of those routes).

agent/internal/controller/controller.go
```diff
@@ -345,13 +349,58 @@ func (w *NodeController) Run(ctx context.Context) error {
...
+	for {
+		name, shutdown := w.contentQueue.Get()
+		if shutdown {
+			return
+		}
+		w.processContentQueueItem(ctx, name)
+	}
+}
+
+func (w *NodeController) processContentQueueItem(ctx context.Context, name string) {
```
danbar2, line 399: **No per-item deadline — the wedge is still permanent.**

This passes `Run`'s ctx straight through, and the typed client is built from `rest.InClusterConfig()` with no `Timeout` (`controller.go:172-186`), so a stalled live Get has no bound at all. A stalled worker is consumed *forever*, and the item stays in the workqueue's `processing` set — so the 10s resync can only re-mark it dirty, never re-drive it. Two concurrent permanent stalls and the content path is dead for the lifetime of the pod.

#302's stated expectation is that "a stalled or slow reconciliation of one PodSnapshotContent must not prevent unrelated work orders from being reconciled and captured." With `contentWorkerCount = 2` and no deadline, that holds for N=1 and breaks for N=2.

Roughly three lines:

```go
func (w *NodeController) processContentQueueItem(ctx context.Context, name string) {
	defer w.contentQueue.Done(name)
	ctx, cancel := context.WithTimeout(ctx, contentReconcileTimeout)
	defer cancel()
	w.reconcilePodSnapshotContent(ctx, name)
}
```

The PR body lists "the typed client has no controller-specific request timeout" as deliberate scope, which is fair as scoping — but it's load-bearing here. With the timeout, `contentWorkerCount` becomes a throughput knob and the pool is self-healing. Without it, the worker count is the only thing standing between the agent and a permanent deadlock.

agent/internal/controller/controller.go
```diff
@@ -212,6 +217,9 @@ func newDefaultController(
 		holderID:  "snapshot-agent/" + uuid.NewString(),
 		inFlight:  make(map[string]struct{}),
 		stopCh:    make(chan struct{}),
+		contentQueue: workqueue.NewTypedDelayingQueueWithConfig(
```
danbar2, line 220: **Errors are swallowed, and the delaying queue is never used as one.**

`reconcilePodSnapshotContent` returns `void` and logs-and-drops every transient error — the content Get, the pod Get, and `labelCaptureEligible` — relying entirely on the 10s resync. The sibling `processRestoreQueueItem` (`controller.go:446-460`) handles this properly with `AddAfter`.

Since `contentQueue` is a `TypedDelayingInterface` and nothing ever calls `AddAfter`, I'd either use it (have reconcile return an error and requeue with a short backoff) or drop to a plain `workqueue.NewTyped[string]()` — the delaying queue starts a `waitingLoop` goroutine for capability the code doesn't exercise.

This matters most on the `labelCaptureEligible` path: that patch is the hand-off *into* the capture path, so a transient failure there costs a full resync interval of capture latency.

agent/internal/controller/controller.go
```diff
@@ -345,13 +349,58 @@ func (w *NodeController) Run(ctx context.Context) error {
...
+	}
+	w.contentQueue.Add(content.Name)
+}
+
+// contentActionable reports whether a PodSnapshotContent can still be reconciled.
+func contentActionable(content *snapshotv1alpha1.PodSnapshotContent) bool {
+	return content.DeletionTimestamp.IsZero() && !isContentTerminal(content)
+}
+
+func (w *NodeController) runContentQueue(ctx context.Context) {
```
danbar2, line 389: **Diverges from the sibling queue's concurrency model.**

`runRestoreQueue` (`controller.go:446`) spawns `go w.processRestoreQueueItem(...)` per item — unbounded, and therefore immune to head-of-line blocking by construction. The content path gets a fixed 2 instead. Same file, same "blocking live API calls" shape, opposite answer.

Worth a sentence on why they differ, or converging them.

agent/internal/controller/controller.go
```diff
@@ -1111,14 +1160,27 @@ func contentFromInformerObj(obj interface{}) (*snapshotv1alpha1.PodSnapshotConte
 	return content, true
 }
 
+// contentNameFromInformerObj extracts the name used to re-drive a work order when its cached
+// representation cannot be converted. The live typed read in the worker remains authoritative.
+func contentNameFromInformerObj(obj interface{}) (string, bool) {
```
danbar2, line 1165: **The tombstone branch is unreachable.**

The content informer registers only `AddFunc`/`UpdateFunc` (`controller.go:291-297`), and client-go delivers `DeletedFinalStateUnknown` exclusively through `OnDelete`. So this branch can never fire.

It was equally dead before, but the function was rewritten and moved here, so this is the moment to drop it. (No need to add a `DeleteFunc` — `enqueuePodSnapshotContent` already filters on `DeletionTimestamp`.)

danbar2 (commented):
Reviewed the PR head locally (`GOOS=linux go vet ./internal/controller` is clean).

The direction is right — informer callbacks should only do cache-to-queue work — and `TestNodeController_ContentInformerDoesNotBlockLaterWork` is a genuine regression test: pre-fix, content-B's Add is never delivered while A blocks the single handler goroutine, so the barrier times out.

My one substantive concern is inline on `processContentQueueItem`: without a per-item deadline this is a constant factor rather than a fix — it turns "1 stall wedges the agent forever" into "2 stalls wedge the agent forever," which still violates #302's stated expectation for N=2. The remaining comments are non-blocking.

Things I checked that are fine: two workers introduce no new race (workqueue guarantees no two workers hold the same key, so per-content serialization is preserved; cross-content shared state is the mutex-guarded `w.inFlight`, the read-only `contentIndexer`, and idempotent merge patches in `labelCaptureEligible`); the double `ShutDown()` (defer in `Run` plus the ctx.Done goroutine) is idempotent in client-go; enqueue-time filtering against the cached object is safe because terminal is sticky and the worker re-checks against a live read; and reordering `contentNameFromInformerObj` to check the tombstone before `meta.Accessor` is a real improvement, since `meta.Accessor` errors on `DeletedFinalStateUnknown` and the old first branch never helped it.

The PR description is unusually precise about what the evidence does and does not prove — worth keeping.

agent/internal/controller/controller.go
```diff
@@ -345,13 +349,58 @@ func (w *NodeController) Run(ctx context.Context) error {
...
+		return
+	}
+	if content.Spec.Source.NodeName != w.config.NodeName || !contentActionable(content) {
+		return
+	}
+	w.contentQueue.Add(content.Name)
+}
+
+// contentActionable reports whether a PodSnapshotContent can still be reconciled.
+func contentActionable(content *snapshotv1alpha1.PodSnapshotContent) bool {
```
danbar2, line 385: Nit: the dedup missed a site. `reconcileSourcePod` (`podsnapshotcontent.go:210-215`) still carries the inline `isContentTerminal(content)` + `!content.DeletionTimestamp.IsZero()` pair that this helper replaced at the other two call sites.

agent/internal/controller/controller_test.go
```diff
@@ -112,12 +116,363 @@ func TestNewDefaultControllerSetsDefaultOperations(t *testing.T) {
...
+func closeBarrier(barrier chan struct{}) {
+	select {
+	case <-barrier:
+	default:
+		close(barrier)
+	}
+}
+
+func TestNodeController_ContentInformerDoesNotBlockLaterWork(t *testing.T) {
+	t.Run("blocked reconciliation does not block later content", func(t *testing.T) {
```
danbar2, line 435: This subtest depends on `contentWorkerCount >= 2`, but never references the constant. If someone sets it to 1, the failure message ("content B did not reach the live Get boundary while content A reconciliation was blocked") won't point at the cause.

Cheap guard at the top of the subtest:

```go
require.GreaterOrEqual(t, contentWorkerCount, 2)
```

agent/internal/controller/controller_test.go
```diff
@@ -112,12 +116,363 @@ func TestNewDefaultControllerSetsDefaultOperations(t *testing.T) {
...
+
+func closeBarrier(barrier chan struct{}) {
+	select {
+	case <-barrier:
+	default:
+		close(barrier)
+	}
+}
+
+func TestNodeController_ContentInformerDoesNotBlockLaterWork(t *testing.T) {
```
danbar2, line 434: Nothing here covers the N+1-stall case — stalling *every* worker, which is the failure mode left open by the missing per-item deadline (see the comment on `processContentQueueItem`).

If you add that timeout, a subtest that stalls both workers and asserts a third content still reconciles once the deadline fires would lock in the self-healing property. That's the assertion that makes this a fix rather than a mitigation.

danbar2 (changes requested):
Thanks for digging into this one — the issue write-up and the deterministic repro are genuinely excellent, and moving the informer callbacks to cache-to-queue work is exactly the right shape for the fix.

Marking 'request changes' to make sure everything lands in a good shape - would be happy to re-review once it is ready

agent/internal/controller/controller.go
```diff
@@ -153,6 +154,10 @@ const (
 	// snapshotContentResyncInterval re-drives every PodSnapshotContent work order so a
 	// not-yet-Ready source pod is re-checked for quiesce without a busy loop.
 	snapshotContentResyncInterval = 10 * time.Second
+	// contentWorkerCount keeps informer delivery independent of live API calls while allowing
+	// unrelated work orders to make progress when one reconciliation is stalled.
+	// When all workers are stalled, later work orders wait in the queue until a worker frees.
+	contentWorkerCount = 2
```
Ronkahn21, line 160: Can we move to configuration instead of hard coded

agent/internal/controller/controller.go
Ronkahn21, line 220: also naming 
```suggestion
		contentCaptureQueue: workqueue.NewTypedDelayingQueueWithConfig(
```

agent/internal/controller/controller.go
```diff
@@ -345,13 +349,58 @@ func (w *NodeController) Run(ctx context.Context) error {
...
+// contentActionable reports whether a PodSnapshotContent can still be reconciled.
+func contentActionable(content *snapshotv1alpha1.PodSnapshotContent) bool {
+	return content.DeletionTimestamp.IsZero() && !isContentTerminal(content)
+}
+
+func (w *NodeController) runContentQueue(ctx context.Context) {
+	for {
+		name, shutdown := w.contentQueue.Get()
+		if shutdown {
+			return
```
Ronkahn21, line 393: add log here

Ronkahn21 (commented):
Thanks for contributing, added additional comments

agent/internal/controller/controller_test.go
```diff
@@ -112,12 +116,363 @@ func TestNewDefaultControllerSetsDefaultOperations(t *testing.T) {
 		noopInjector{},
 		testr.New(t),
 	)
+	t.Cleanup(w.contentQueue.ShutDown)
 	t.Cleanup(w.restoreQueue.ShutDown)
-	if w.checkpointFn == nil || w.restoreFn == nil || w.writeControlSentinelFn == nil || w.controlSentinelExistsFn == nil || w.sendSignalFn == nil || w.restoreQueue == nil {
+	if w.checkpointFn == nil || w.restoreFn == nil || w.writeControlSentinelFn == nil || w.controlSentinelExistsFn == nil || w.sendSignalFn == nil || w.contentQueue == nil || w.restoreQueue == nil {
```
oleg-kushniriov, line 121: Nit, pre-existing: when this chain trips, the single `t.Fatal` doesn't say which of the seven fields was nil. Since this PR extends it with `contentQueue`, consider iterating a name → value map with `require.NotNil(t, v, name)` instead, so the failure names the missing seam or queue and the next addition doesn't grow the boolean chain.

PR #218, opened 2026-09-04
feat(agent): checkpoint and restore POSIX CUDA VMM state

galletas1712:
## Summary

Layer 9 of the eleven-PR [cuinterpose stack #293](https://github.com/ai-dynamo/snapshot/stack/293), based on the host-carrier module in #292. Closed PR #214 is not in the active stack.

This PR checkpoints and restores actually shared POSIX-FD CUDA VMM allocations. It owns allocation eligibility and unicast topology; #292 owns the D2H/H2D host-carrier mechanics.

### Ownership boundary

| Allocation | Checkpoint owner |
| --- | --- |
| Private VMM allocation, including POSIX-capable memory never exported or bound into tracked multicast | Native CUDA checkpoint |
| Allocation actually exported/imported through the tracked POSIX sharing path | Interposer |
| Allocation bound into tracked multicast | Interposer; #219 marks the member shared |
| Legacy `cuIpc*` memory | Native CUDA checkpoint |

Export capability alone does not make an allocation shared. Private allocations remain mapped with their native handles intact when the interposer finishes preparation. The native driver remains responsible for saving and rebuilding their physical backing; this does not imply that GPU contents survive checkpoint automatically.

Only shared eligible creators advertise `ALLOCATION_CONTENT` and use host carriers. Imported allocations reference their canonical creator without duplicating content. Private records remain available for topology inspection, but no longer advertise `CONTENT_ONLY`; its protocol bit is reserved. Existing device-location, pinned-allocation, and exportability checks are retained. Unsupported exportable resources and live raw imports still fail preflight before destructive teardown.

### Lifecycle

```text
PREPARE_MULTICAST → SAVE_ALLOCATIONS → PREPARE_UNICAST
```

At this layer multicast preparation is a phase boundary; #219 supplies teardown. `SAVE_ALLOCATIONS` recovers handles only for shared creators with surviving mappings, copies their contents through #292, and advances all participants. `PREPARE_UNICAST` quiesces descriptor export, drains in-flight requests, and unmaps/releases only shared allocations. Private allocations are not temporarily retained, copied, unmapped, or released.

```text
LOAD_ALLOCATIONS → RESTORE_UNICAST
```

[...]

PR #292, opened 2026-09-11
refactor(agent): isolate cuinterpose host-carrier storage

galletas1712:
## Summary

Layer 8 of the eleven-PR [cuinterpose stack #293](https://github.com/ai-dynamo/snapshot/stack/293). Closed PR #214 is intentionally not in the active stack.

This PR isolates the current allocation-content transport in `host_carrier.c`/`.h`, with CUDA context scoping in `context.c`/`.h`. It does not add lifecycle operations, change allocation eligibility, introduce a backend registry, or add a save-all flag. #218 calls this concrete module from the generic `SAVE_ALLOCATIONS` and `LOAD_ALLOCATIONS` operations.

The public surface is deliberately small: save a batch, load a batch, release restored storage, and discard inherited CPU-side state after `fork`. The lifecycle caller supplies each allocation's context, properties, size, device-handle pointer, and host-address pointer; it remains responsible for eligibility, mappings, access grants, export caches, and importer topology.

On save, the module creates one anonymous host arena per shim process and registers it as portable pinned memory. It groups allocations by CUDA context, maps each group into a contiguous temporary device VA range, and copies D2H on one nonblocking stream with one synchronization per group. CRIU captures the arena as ordinary process memory.

On load, the module verifies or recreates host registration, creates fresh physical allocations with the original properties, stages each context group, copies H2D, and publishes fresh handles only after the complete batch succeeds. Save and load each use an explicit operation-state object with scoped cleanup and direct returns, so partial failures release fresh handles and partial host state without `goto`. The adjacent context helper restores the caller's prior context and can retain a device primary context when tracking never observed one.

## PageBroker boundary

This is a replacement seam, not a speculative pluggable framework. PageBroker currently implements only `TransferEngineType::POSIX_COPY` and registers only `PosixCopyEngine`; its protobuf reserves `DirectRestoreRequest` for a future GPU engine. There is no GPU engine to integrate with today.

When PageBroker defines a GPU/direct engine and a transaction contract for allocation IDs, content, and completion barriers, the two host-carrier lifecycle calls are the intended replacement point. Tickets, participant identity, topology records, and coordinator ordering should not need to become PageBroker-aware.

## Stack boundary

Based on #217. #218 owns which supported exportable allocations preserve content and composes this module with unicast teardown, creator publication, and importer reconstruction.

## Validation

The pinned CUDA 13.1 builder compiles `host_carrier.c` into the production shim with `-Werror`. The #218 lifecycle suite exercises successful save/load, context grouping, missing registration recovery, failure cleanup, and fork behavior through the fake CUDA driver; the final stack passes all 4 lifecycle tests under ASan/UBSan. `make test` passes in all Go modules. The final published stack head `0eae9c4f9b66093332f00a332afaf56048e1e64a` passed the full `make check` gate, the CUDA 13.1 production build, and all 73 sanitizer-backed native cuinterpose tests. Its physical-GPU suite passed all 3 tests with no skips on two DRA-assigned NVIDIA B200 GPUs. The unicast test kept an explicit allocation-ID ticket-backed peer mapping per worker live across capture and restore; the multicast test required a nonzero multicast VA and shim-logical handle, exercised `BindAddr`, and passed its collective and captured-graph replay. Detailed hardware evidence and measurements are in #220.

### DIFF
--- a/agent/cmd/cuinterpose/Makefile
+++ b/agent/cmd/cuinterpose/Makefile
@@ -120,7 +120,7 @@ PRELOAD_TESTS := $(patsubst tests/%.cc,$(TEST_DIR)/%,$(wildcard tests/*_preload_
 COORDINATOR_TEST := $(patsubst tests/%.cc,$(TEST_DIR)/%,$(wildcard tests/coordinator_test.cc))
 # Shim sources that need no CUDA driver at run time. symbols.c and context.c
 # resolve driver symbols and are exercised through the preload tests.
-UNIT_SOURCES := $(filter-out coordinator.c interpose.c multicast.c symbols.c context.c,$(wildcard *.c)) $(UTIL_SOURCES)
+UNIT_SOURCES := $(filter-out coordinator.c interpose.c multicast.c symbols.c context.c host_carrier.c,$(wildcard *.c)) $(UTIL_SOURCES)
 UNIT_OBJECTS := $(patsubst %.c,$(TEST_DIR)/%.o,$(UNIT_SOURCES))
 # When the shim under test is ASan-instrumented, the ASan runtime must be the
 # first library in the process, ahead of the shim itself.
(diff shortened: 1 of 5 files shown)


galletas1712:

Issue #278, opened 2026-09-10
Sign, attest, and publish verifiable release artifacts


danbar2:
## Problem

Snapshot publishes container images and a Helm chart to GHCR, but nothing about
those artifacts is verifiable by a consumer. There is no signature, no SBOM, no
checksum, and no provenance attestation. A user pulling
`ghcr.io/ai-dynamo/snapshot` has no way to establish that the image was built by
this repository's CI from the commit it claims.

Concretely, today:

- All four published releases (`v0.1.0`, `v0.1.0-rc2`, `v0.1.0-rc.1`,
  `v0.1.0-alpha.1`) carry **zero release assets** — everything goes to GHCR only.
- `git grep -i 'cosign|sigstore|sbom|syft|spdx|cyclonedx'` over `.github/`,
  `Makefile`, and `hack/` returns nothing.
- `push-artifacts.yaml` builds and pushes with buildx but passes neither
  `--sbom` nor `--provenance`.

## Proposed implementation

All of this lands in `.github/workflows/push-artifacts.yaml`, which already has
the release context (`version`, `head_sha`) and the GHCR login.

**1. Keyless signing with cosign.** Add `id-token: write` to the image and chart
jobs and sign each pushed artifact by digest with Sigstore keyless signing —
no key material to manage, and the resulting Rekor entry ties the artifact to
the workflow identity:

```
cosign sign --yes "ghcr.io/ai-dynamo/snapshot/operator@${DIGEST}"
```

Sign the operator image, the agent image, and the Helm chart OCI artifact. Note
the build steps currently go through `make docker-build-*` with `--push`, so the
pushed digest needs to be captured (`--metadata-file` or
`docker buildx imagetools inspect`) rather than assumed from the tag.

**2. Build-time provenance and SBOM attestations.** Add `--provenance=mode=max
--sbom=true` to the buildx invocations. This attaches in-toto SLSA provenance
and an SPDX SBOM to each image as OCI attestations, which is what OpenSSF's
SLSA check looks for.

**3. SBOM and checksums as release assets.** The image attestation above does
not satisfy `SIG-SEC-17`, which specifically wants an asset on the GitHub
release. Generate an SPDX SBOM with syft for each image, write a `SHA256SUMS`
covering the SBOMs and the chart tarball, and upload all of it to the release:

```
gh release upload "${TAG}" sbom-operator.spdx.json sbom-agent.spdx.json \
  snapshot-${VERSION}.tgz SHA256SUMS
```

[...]

PR #303, opened 2026-09-14
docs/ci: cosign v3 verification, SECURITY.md signing section, and GHCR pruning

danbar2:
Part of #278

`RELEASE.md` tells consumers to verify releases with cosign but never states a
minimum version — and the version most people have installed reports correctly
signed artifacts as unsigned.

## The problem

CI signs with **cosign v3.0.6** (the default of the pinned
`sigstore/cosign-installer@v4.1.2`). Cosign v3 stores signatures in Sigstore's
bundle format, attached via the OCI 1.1 referrers fallback tag
`sha256-<digest>`.

Cosign **v2** looks for the older `sha256-<digest>.sig` tag, does not find one,
and reports:

```
Error: no signatures found
```

That is not a version error — it is a *missing signature* error. Someone
following our own instructions on cosign v2 concludes the release is unsigned.
A verification document that produces a false negative on the exact property it
exists to prove is worse than no document.

This is not a corner case: `brew install cosign` currently yields v2.4.1.

## Evidence

Verified by hand against the `v0.0.0-geb58c630` dev build published by #299, for
all three signed artifacts:

| cosign | operator | agent | chart |
| --- | --- | --- | --- |
| v3.0.6 | verifies | verifies | verifies |
| v2.4.1 | `no signatures found` | `no signatures found` | `no signatures found` |

The v3 verification confirms the Rekor transparency-log entry and the Fulcio
certificate chain, so the signing itself is correct. Only the documentation was
wrong.

Registry state for the operator digest, for the record:

- `sha256-<digest>.sig` (legacy tag cosign v2 looks for) → **404**
- OCI 1.1 referrers API → **0 entries**
- `sha256-<digest>` (referrers fallback tag, where v3 actually wrote it) → **present**

## The fix

State the minimum version, explain the format split, and name the exact error
string so anyone who hits it checks `cosign version` before concluding their
artifact is unsigned.

Documentation only; no workflow or behaviour change.

## Summary by CodeRabbit

* **New Features**
  * Added automated weekly cleanup of untagged, unreachable container images, with manual and dry-run options.
  * Release artifacts from version 0.2.0 onward now include Sigstore bundles for checksum verification and provenance tracking.

* **Documentation**
  * Added guidance for verifying signed artifacts, transparency logs, attestations, checksums, and software bills of materials.
  * Clarified that releases before 0.2.0 are unsigned and include limited provenance, with backfilled SBOMs.
  * Updated verification requirements to use Cosign 3.0 or newer.


.github/workflows/prune-registry.yaml
```diff
@@ -0,0 +1,57 @@
...
+        with:
+          persist-credentials: false
+
+      # DRY_RUN defaults to true on the schedule as well: this deletes published
+      # artifacts, so it reports for a few cycles and a maintainer flips the
+      # default here once the output has been reviewed.
+      - name: Prune
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          DRY_RUN: ${{ inputs.dry_run == false && 'false' || 'true' }}
```
oleg-kushniriov: According to the GitHub's documented expression coercion table (null and false both coerce to 0 for ==), not just inferred. This is a known Actions footgun with boolean inputs compared via == across trigger types — inputs is only populated for workflow_dispatch/workflow_call, so any schedule-triggered comparison against a boolean input needs to route around it explicitly.

Suggestions:

Branch on github.event_name instead of relying on coercion, e.g.:

`DRY_RUN: ${{ github.event_name == 'schedule' && 'true' || (inputs.dry_run == false && 'false' || 'true') }}`

oleg-kushniriov (approved):

Issue #37, opened 2026-07-30
Handle GPU device ordinal mismatch between checkpoint source and restore target


shayasoolin:
## Summary

Restore can fail when the checkpoint source pod and restore target pod receive
different NVIDIA device ordinals on the same node.

During manual testing on a 4-GPU node, the source pod checkpointed with one
device path, but the restore pod was allocated a different one. CRIU then tried
to restore a mount for a device node that was not present in the restore target
namespace.

## Observed failure

```text
Error (criu/mount-v2.c:634): mnt-v2:
Can't stat mountpoint .../nvidia1: No such file or directory
```

The captured device path can be seen in the checkpoint artifact:

```bash
grep -oE '/dev/nvidia[0-9]+' /checkpoints/gpu-poc-1/versions/1/manifest.yaml | sort -u
```

## Why this matters

Kubernetes schedules GPU resources, but a restore target may not receive the
same `/dev/nvidiaX` ordinal that the source pod had at checkpoint time. CRIU
restore can then fail before Snapshot has a chance to complete CUDA remapping or
report a clearer user-facing cause.

shayasoolin:
Issue is resolved when using NVIDIA gpu-operator v26.3.3. CDI will be documented as a prereq.

PR #306, opened 2026-09-15
fix(agent): restore UUID-mapped legacy GPU device paths

### LINKED ISSUE #305: Support UUID-mapped GPU device paths for legacy NVIDIA_VISIBLE_DEVICES restore
## What happened?

Cross-GPU restore with legacy NVIDIA device injection can fail before CUDA restore because CRIU replays the checkpoint-time `/dev/nvidiaN` external mount, but the destination container exposes its allocated GPU at a different physical device path.

Related: #37 was closed after a CDI-based resolution. This issue tracks explicit legacy `NVIDIA_VISIBLE_DEVICES` selection and UUID-mapped device-path restoration, including multi-device ordering and overlapping paths. It does not claim to resolve driver-level `CUDA_ERROR_NOT_SUPPORTED` failures.

## What did you expect to happen?

For a compatible source and destination, Snapshot should resolve explicit NVIDIA GPU selections to canonical UUIDs, retain their physical device paths in the checkpoint manifest, and prepare the appropriate destination mount aliases before CRIU. Existing DRA/device discovery should continue working when no explicit list is supplied.

## How can we reproduce it?

1. Run a checkpoint-ready Qwen3-0.6B vLLM workload with legacy NVIDIA injection and an explicit `NVIDIA_VISIBLE_DEVICES` index or UUID.
2. Capture a fresh checkpoint and record the source physical `/dev/nvidiaN` mount.
3. Restore into a pod allocated a different GPU, exposing a different physical device path.
4. Observe CRIU attempting to resolve the source device path before the CUDA UUID migration step.

The original missing-path failure was reproduced in kind. The revised implementation subsequently completed actual cross-GPU restore locally and on two Kubernetes environments.

## Environment

- Investigated Snapshot base: `6725e1bcbc56b5d1ff2ae2d6c41c671180269cb1`.
- Local: Linux/amd64, kind/containerd, two RTX 5880 Ada GPUs, driver 615.02.
- Regression environments: A100 with device-plugin discovery; B200 with DRA discovery.
- Model: Qwen/Qwen3-0.6B, eager mode, maximum model length 2048.
- Used an existing prebuilt checkpoint-ready vLLM runtime; no runtime image was built locally.
- The independently reported B300/590.48.01/private-vLLM setup was not reproduced exactly.

## Relevant logs

```text
Can't stat mountpoint .../dev/nvidiaN: No such file or directory
```

This happens before `cuda-checkpoint-helper --action restore`. A later `CUDA_ERROR_NOT_SUPPORTED` is a separate failure class.

## Proposal tracking

Following the SEP process proposed in #253, this issue supplies the proposal number. The review branch will include `docs/proposals/<issue-number>-legacy-gpu-device-remap/README.md` with motivation, design, risks, alternatives, and validation.

Proposed changes:

- Resolve explicit `NVIDIA_VISIBLE_DEVICES` index/full-UUID lists against the host inventory.
- Persist the raw value and resolved UUID-to-physical-device-path map.
- Build CRIU mount aliases from the same UUID pairing used for CUDA migration.
- Pin all destination devices before changing mounts, including overlapping paths and swaps.
- Roll back partial preparations on failure.
- Preserve existing discovery for absent/non-list values; retain conservative handling of older manifests.

`CUDA_VISIBLE_DEVICES` is an application-level filter/reordering mechanism, not the source of host device ordinals for this resolution. This work does not rewrite it or add MIG support.

## Validation and remaining work

- All Go module unit tests and agent lint passed.
- Local real cross-GPU restore passed after a host reboot and kind-specific mount-propagation normalization before fresh capture.
- A100: capture 70.05s, cross-GPU restore 18.05s.
- B200/DRA: capture 29.19s, same-GPU restore 3.03s, explicitly different-UUID restore 3.10s.
- Each successful restored Qwen workload generated output.
- Multi-GPU overlaps/swaps have unit coverage; a multi-GPU workload remains an end-to-end validation gap.
- Request maintainer triage and `approved` before opening the implementation/proposal PR.

## Before submitting

- [x] Searched existing issues and PRs. #37 is related but closed with a CDI resolution; no open implementation of this explicit-selection/device-path fix was found.

AI assistance was used for implementation, investigation, and test execution.

galletas1712:
#### What type of PR is this?

- [x] bug
- [x] documentation

## Summary

CRIU reconstructs external NVIDIA device mounts before CUDA restore. With
legacy injection, the destination pod can expose a different `/dev/nvidiaN`
than the checkpoint recorded, so CUDA UUID remapping alone is too late.

- Resolve explicit `NVIDIA_VISIBLE_DEVICES` index/full-UUID lists against host
  inventory with one deadline covering the entire operation. Absent/`all` retains
  existing discovery; empty/`none`/`void` uses only actual container visibility,
  allowing independent CDI injection without allocation fallback.
- Persist the raw selection and UUID-to-physical-device-path metadata.
- Compute one mount-alias plan from the CUDA UUID pairing and use it for both
  compatibility inspection and CRIU preparation.
- Pin destination devices before creating aliases to handle overlapping paths
  and swaps; roll back partial preparation on failure.
- Require a fresh checkpoint for device-path remapping; older manifests keep
  their existing behavior without inferred mappings.
- Document the behavior in `docs/limitations.md`. The unnecessary SEP and
  old-checkpoint inference implementation were removed following review.

#### Which issue(s) this PR fixes:

Fixes #305

Related to #37, which was closed with a CDI-based resolution. This change
addresses the explicit legacy-selection/device-path case instead. No open
implementation of this fix was found during duplicate-work checks.

## Validation

At commit `93d9e59`:

- `make check` — passed, including lint, vulnerability checks, and clean-tree gate.
- `make build` — passed.
- `make test` — passed across all three Go modules.
- Regression tests cover disabled legacy/CDI visibility, deadline cancellation,
  duplicate selections, physical device validation, overlapping/swap mappings,
  unknown source UUIDs, old manifests, and resolving paths before compatibility.

Fresh local validation of the revised implementation: Qwen3-0.6B restored from
one RTX 5880 Ada GPU to the other with compatibility checks enabled, mapping
`/dev/nvidia0` to `/dev/nvidia1`. Restore took 2.83s and generated
`The answer is "restored"`. The final commit subsequently removed a redundant
post-gate guard and updated comments; no hardware behavior was changed.

Earlier hardware tests used Qwen/Qwen3-0.6B in eager mode with an existing
prebuilt checkpoint-ready vLLM runtime. Remote tests below predate the review
revision and have not been rerun on this revision:

| Environment | Validation | Result |
| --- | --- | --- |
| Local kind, two RTX 5880 Ada GPUs | Explicit selection, real cross-GPU restore | Passed; restored model generated output |
| A100 Kubernetes | Existing device discovery, different GPU UUID | Capture 70.05s; restore 18.05s |
| B200 Kubernetes/DRA | Same-GPU restore | Restore 3.03s |
| B200 Kubernetes/DRA | Selector explicitly excluding source UUID | Capture 29.19s; restore 3.10s |

Each successful restored workload generated output. Only Snapshot images were
built locally; no runtime images were built locally.

#### Special notes for your reviewer:

- Multi-GPU overlap/swap logic has unit coverage, but a multi-GPU workload has
  **not** been validated end to end.
- Local kind required private NVIDIA mounts before a fresh capture to avoid
  nested mount-propagation errors. Remote tests needed no such adjustment.
- Initial B200 destination pods were rejected for node `DiskPressure` before
  CRIU ran. Kubelet cleared this without host mutation before successful retries.
- This does not establish support for the reported B300/590.48.01/private-vLLM
  configuration or resolve its separate `CUDA_ERROR_NOT_SUPPORTED` failure.
- `CUDA_VISIBLE_DEVICES` is not interpreted or rewritten; application-level
  filtering/reordering remains a runtime qualification gap.
- AI assistance was used for implementation, investigation, and test execution.

#### Does this PR introduce an API change?

No Kubernetes CRD changes. Checkpoint CUDA manifests gain optional
`nvidiaVisibleDevices` and `devicePaths` fields. Older manifests retain their
existing restore behavior; fresh capture is required for the new remapping.

```release-note
Support explicit legacy NVIDIA_VISIBLE_DEVICES selections and UUID-mapped GPU device mount preparation before CRIU restore. CUDA checkpoint manifests include optional selection and device-path metadata.
```

#### Additional documentation, e.g. enhancement proposals, usage docs:

- [GPU selection and restore limitations](docs/limitations.md#gpu-device-selection-and-restore-paths)
- Tracking issue: #305 (approved)

#### Checklist

- [x] Commits are signed off (`git commit -s`).
- [x] `make check test` passes locally.
- [x] Documentation is updated where behavior changed.

## Summary by CodeRabbit

- **New Features**
  - Added GPU selection based on container visibility settings, including explicit IDs and UUIDs.
  - Checkpoints now preserve GPU device paths and visibility settings for more accurate restores.
  - Added configurable GPU device-path remapping during restore.

- **Bug Fixes**
  - Improved compatibility checks for remapped GPU devices and ordinary mounts.
  - Added validation for invalid, duplicate, mismatched, or physically inconsistent GPU selections.

- **Documentation**
  - Documented GPU selection precedence and checkpoint restore behavior.


galletas1712:
Resolved conflicts with current main (`eb58c63`) in merge commit `d068683`, without rewriting branch history. Preserved the new compatibility gate and GPU model/driver manifest metadata alongside the explicit NVIDIA_VISIBLE_DEVICES and device-path mapping changes. The inspect gate validates mapped destination GPU paths while retaining ordinary mount checks; regression tests cover both interactions.

Validation: `make test build` and `make check` passed. GPU end-to-end tests have not been rerun after this merge; earlier hardware evidence remains scoped to the pre-merge implementation.

agent/internal/cuda/devices.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+	return resolveVisibleDevices(ctx, *value)
+}
+
+// VisibleDevicesValue distinguishes an absent selection from an explicit empty value.
+func VisibleDevicesValue(env []string) *string {
+	var value *string
+	for _, entry := range env {
+		if v, ok := strings.CutPrefix(entry, "NVIDIA_VISIBLE_DEVICES="); ok {
+			value = &v
```
leeZardNav: should the function just return at this point?

agent/internal/cuda/devices.go
```diff
@@ -0,0 +1,132 @@
...
+		}
+	}
+	return value
+}
+
+func resolveVisibleDevices(ctx context.Context, value string) ([]string, error) {
+	switch value {
+	case "", "all", "none", "void":
```
leeZardNav: maybe and if statement would be easier?

agent/internal/cuda/devices.go
```diff
@@ -0,0 +1,132 @@
...
+		if !ok {
+			return nil, fmt.Errorf("no physical device minor for GPU %s", uuid)
+		}
+		path := fmt.Sprintf("/dev/nvidia%d", minor)
+		var stat unix.Stat_t
+		if err := unix.Stat(filepath.Join(hostProc, strconv.Itoa(pid), "root", path), &stat); err != nil {
+			return nil, fmt.Errorf("GPU %s is not exposed at %s: %w", uuid, path, err)
+		}
+		if stat.Mode&unix.S_IFMT != unix.S_IFCHR || unix.Major(stat.Rdev) != 195 || unix.Minor(stat.Rdev) != minor {
+			return nil, fmt.Errorf("GPU %s device mismatch at %s", uuid, path)
```
leeZardNav: it could be my lacking of knowledge in nvidia or GPU, but in my opinion this technical parts should get their own (very small) function, or const, with a bit deeper explanations

i'm not entirely sure i can verify/disqualify anything you say here regarding the files you are looking at or their paths

agent/internal/executor/compat.go
```diff
@@ -28,6 +29,25 @@ func inspectCompatibility(
...
+				continue
+			}
+			mappedPaths[sourcePath] = targetPath
+		}
+		for j, path := range sourceEnv.ExternalizedMounts {
+			if targetPath, ok := mappedPaths[path]; ok {
+				sourceEnv.ExternalizedMounts[j] = targetPath
+			}
+		}
+	}
```
leeZardNav: this code which is specific for the paths seems our of place here. i suggest it should be in the CompatEnvironment function

agent/internal/executor/checkpoint.go
```diff
@@ -259,23 +259,40 @@ func inspectContainer(ctx context.Context, rt snapshotruntime.Runtime, log logr.
...
-			ctx,
-			req.Clientset,
-			req.PodName,
-			req.PodNamespace,
-			req.ContainerName,
-			snapshotruntime.HostProcPath,
-			pid,
-			log,
-		)
+		if ociSpec != nil && ociSpec.Process != nil {
```
leeZardNav, line 267: is this the same code in the restore? if so should it have its own helper function or be part of one as part of ResolveVisibleGPUs?

agent/internal/criu/restore.go
```diff
@@ -195,6 +224,260 @@ func buildRestoreExtMounts(m *types.CheckpointManifest) ([]*criurpc.ExtMountMap,
...
+
+func gpuMountAliases(m *types.CheckpointManifest, deviceMap string, targets map[string]string) (map[string]string, error) {
+	mapping := map[string]string{}
+	for _, uuid := range m.CUDA.SourceGPUUUIDs {
+		mapping[uuid] = uuid
+	}
+	if deviceMap != "" {
+		for _, pair := range strings.Split(deviceMap, ",") {
+			source, target, ok := strings.Cut(pair, "=")
+			if !ok || mapping[source] == "" || target == "" {
```
leeZardNav: when should `mapping[source] == "" ` happen?
it is initiated as `mapping[uuid] = uuid` and then set to `mapping[source] = target` which is checked for empty string

agent/internal/cuda/devices.go
```diff
@@ -0,0 +1,132 @@
...
+	"golang.org/x/sys/unix"
+
+	"github.com/ai-dynamo/snapshot/api/compat"
+)
+
+// ResolveVisibleGPUs preserves explicit selection order and includes the model
+// and driver metadata used by the restore compatibility gate.
+func ResolveVisibleGPUs(ctx context.Context, env []string) (compat.GPUInfo, error) {
+	uuids, err := ResolveVisibleDevices(ctx, env)
+	if err != nil || len(uuids) == 0 {
```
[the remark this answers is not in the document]
galletas1712: Addressed in 93d9e59 without treating disabled legacy injection as disabled CDI: absent/all retains allocation discovery; empty/none/void queries only actual container visibility and never falls back to allocation UUIDs. This permits independently injected CDI devices but cannot substitute GPUs from host/allocation discovery. Added tests for all three values with both empty visibility and CDI-visible GPUs; documented the distinction.

agent/internal/cuda/devices.go
```diff
@@ -0,0 +1,132 @@
...
+		return compat.GPUInfo{}, err
+	}
+	ctx, cancel := context.WithTimeout(ctx, nvidiaSMITimeout)
```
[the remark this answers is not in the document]
galletas1712: Fixed in 93d9e59. resolveSelectedGPUs creates the shared timeout before any NVIDIA command, and each selection query now returns UUID/model/driver together rather than issuing a second metadata pass. Added a stalled-command deadline test.

agent/internal/executor/restore.go
```diff
@@ -371,7 +382,7 @@ func inspectRestore(
 
 	// Ahead of BuildDeviceMap, whose positional pairing turns a GPU difference
 	// into a device-map error that names neither GPU.
-	if err := inspectCompatibility(log, manifest, targetGPUs, targetRoot, targetImageID, req.SkipCompatCheck); err != nil {
+	if err := inspectCompatibility(log, manifest, targetGPUs, gpuDevicePaths, targetRoot, targetImageID, req.SkipCompatCheck); err != nil {
```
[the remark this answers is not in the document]
galletas1712: Confirmed: the published merge commit still passed a nil path map at the gate; my earlier summary was incorrect. Fixed in 93d9e59 by resolving paths and building one alias plan before inspectCompatibility, then passing that exact plan to nsrestore. The regression test exercises this orchestration, not just the gate helper, and still rejects missing ordinary mounts. A fresh local cross-GPU Qwen restore passed with compatibility checks enabled (2.83s, generated output).

agent/internal/cuda/devices.go
galletas1712: Changed in 93d9e59: scan backwards and return immediately. This keeps last-assignment-wins behavior for duplicate environment entries while eliminating the temporary pointer state. Added explicit duplicate/empty/absent coverage.

agent/internal/cuda/devices.go
galletas1712: Simplified in 93d9e59. Removed this switch and the extra resolution wrapper; selection precedence now lives in DiscoverGPUs, with a small if for disabled legacy values and a distinct explicit-list path.

agent/internal/cuda/devices.go
galletas1712: Addressed in 93d9e59: extracted validateGPUDevice, named nvidiaDeviceMajor, linked the Linux device-number documentation, and explained that the NVIDIA procfs information files provide UUID/physical minor associations, not CUDA ordinals. Validation checks the character device identity in the workload root rather than trusting its filename.

agent/internal/executor/compat.go
galletas1712: Agreed that this mapping logic was out of place. In 93d9e59 it is removed from compat.go. A single CRIU alias plan is built before inspection and reused unchanged by nsrestore; existingMountPaths follows that plan while reporting source mount names. I did not put destination-dependent mapping into CompatEnvironment(), which describes the captured source and should not need restore-target state.

agent/internal/executor/checkpoint.go
galletas1712: Agreed and addressed in 93d9e59. Both callers now pass the workload environment to DiscoverGPUs; explicit selection, disabled-legacy visibility, and allocation fallback are handled there. Removed the duplicated selection/fallback branches and combined UUID/model/driver lookup per selection.

agent/internal/criu/restore.go
galletas1712: This was detecting a source UUID absent from the initialized map, rather than one mapped to an empty target. Replaced it with an explicit membership lookup in 93d9e59 and added an unknown-source regression case, making that intent clear.

galletas1712:
Review update pushed in `93d9e59`: removed the SEP and old-checkpoint inference path; reduced additions from 1,000 to 636 (including regression tests); consolidated discovery and reused one alias plan for inspection and CRIU. Replied to all nine review threads; leaving them open for reviewer confirmation.

`make check test build` passed. Fresh local Qwen3-0.6B cross-GPU restore passed with compatibility checking enabled, mapping /dev/nvidia0 to /dev/nvidia1 and generating output (2.83s). As previously, nested kind needed private NVIDIA mounts before capture. No runtime image was built. Remote A100/B200 results remain from the earlier revision, not this update; general multi-GPU workload validation is still outstanding.

agent/internal/executor/restore.go
```diff
@@ -37,6 +38,44 @@ type RestoreMounter interface {
...
+			if err != nil {
+				return "", nil, err
+			}
+			aliases, err = criu.GPUMountAliases(manifest, deviceMap, paths)
+			if err != nil {
+				return "", nil, err
+			}
+		}
+	}
+	if err := inspectCompatibility(log, manifest, target, aliases, targetRoot, imageID, skip); err != nil {
```
leeZardNav: inspectGPUCompatibility calls inspectCompatibility seems a bit not in line with the compatibility check. the compatibility check contain several checks in a mechanism that allow to add more and prepare the structure in the restore.go. i suggest to add the inspectGPUCompatibility to the same structure

agent/internal/executor/compat.go
leeZardNav: resolving this as i opened a new comment on the revised code

agent/internal/executor/nsrestore.go
```diff
@@ -191,6 +207,7 @@ func executeRestore(
 	if err != nil {
 		return nil, 0, nil, err
 	}
+	gpuMountsCommitted = true
```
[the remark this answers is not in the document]
galletas1712: Fixed in 093c87d: gpuMountsCommitted is set only immediately before successful return, after process-table validation and CUDA restore/unlock. All intermediate errors keep deferred alias cleanup enabled. make check, make test, and make build pass. Post-CRIU failure injection has not been exercised on hardware for this revision.

agent/internal/executor/restore.go
galletas1712: Agreed; addressed in 093c87d. Removed inspectGPUCompatibility. prepareGPUMapping only prepares the CUDA map/aliases; inspectRestore then calls the existing inspectCompatibility gate directly. GPU count/model/driver and mount policy remains in compat.Compare with its registered checks, rather than adding a parallel GPU gate. Added coverage showing count mismatch is reported by the registered inspect check without attempting path resolution. make check/test/build pass.

agent/internal/cuda/devices.go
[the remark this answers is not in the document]
galletas1712: The visibility code is unchanged by the latest follow-up. Locally verified again with make test: TestDisabledLegacySelectionUsesOnlyContainerVisibility covers empty/none/void with both no devices and independently injected CDI devices. These values call DiscoverVisibleGPUs directly and cannot reach allocation fallback. The reported clone failure is review infrastructure, not a new code finding; please retry verification when repository access is available.

agent/internal/executor/restore.go
```diff
@@ -37,6 +38,30 @@ type RestoreMounter interface {
 	MountPageBroker(ctx context.Context, namespaceMount nsmount.MountPoint, stagingPath string) (nsmount.MountPoint, error)
 }
 
+// prepareGPUMapping produces the plan used by both mount inspection and nsrestore.
+// Compatibility policy remains in the registered checks, not in this preparation.
+func prepareGPUMapping(log logr.Logger, manifest *types.CheckpointManifest, uuids []string,
```
leeZardNav: perfectttt

leeZardNav (approved):

dfeigin-nv:
Is host inventory calculated once upon initialization of the agent? @galletas1712

Ronkahn21 (approved):

galletas1712:
@dfeigin-nv No, currently it’s resolved per checkpoint/restore operation. We could cache the host inventory at agent initialization while keeping container device-node validation per operation. This might reduce restore times. I can maybe put this as an enhancement.

PR #313, opened 2026-09-15
feat(runtime): terminate containers by runtime identity

hhzhang16:
#### What type of PR is this?

- [x] enhancement

#### What this PR does / why we need it:

Adds an internal runtime primitive for terminating a container through CRI rather than thru host PID (which could be reused). 

#### Which issue(s) this PR fixes:

N/ A

#### How was this tested?

- Runtime termination tests covering containerd, CRI-O, and bare CRI identifiers
- Exact normalized identifier and zero-second grace verification
- Empty, cross-runtime, unknown, and nested scheme rejection with no backend call
- Context-cancellation and wrapped backend-error coverage
- Linux runtime, controller, and executor package validation
- Linux cross-compilation
- `git diff --check`

#### Special notes for your reviewer:

This slice intentionally adds no production caller. The subsequent recovery-policy MR will use it to terminate an identity-bound workload before recording restore failure, and will retain the finalizer when termination cannot be confirmed.

#### Does this PR introduce an API change?

```release-note
NONE
```

#### Additional documentation, e.g. enhancement proposals, usage docs:
```docs
NONE
```

#### Checklist

- [x] Commits are signed off (`git commit -s`), per [CONTRIBUTING.md](../CONTRIBUTING.md)
- [ ] `make check test` passes locally
- [x] Documentation is updated where behavior changed

## Summary by CodeRabbit

* **New Features**
  * Added support for stopping containers through containerd and CRI-O runtimes.
  * Containers can be stopped using recognized runtime identifiers or bare container IDs.
  * Added validation for empty, malformed, nested, unknown, or mismatched runtime identifiers.
  * Stop requests now use zero-grace termination and handle already-stopped containers safely.
  * Context cancellation and runtime errors are preserved and reported clearly.


agent/internal/runtime/oci_containerd.go
```diff
@@ -46,6 +46,17 @@ func (r *ContainerdRuntime) ResolveContainerImageID(ctx context.Context, contain
 	return resolveContainerImageID(ctx, r.cri, containerID)
 }
 
+func (r *ContainerdRuntime) TerminateContainer(ctx context.Context, containerID string) error {
+	id, err := containerIDForRuntime(containerID, "containerd://")
+	if err != nil {
+		return fmt.Errorf("invalid containerd container ID %q: %w", containerID, err)
```
dfeigin-nv, line 50: There's no timeout here

agent/internal/runtime/oci_containerd.go
```diff
@@ -46,6 +46,17 @@ func (r *ContainerdRuntime) ResolveContainerImageID(ctx context.Context, contain
...
+	}
+	if err := r.cri.StopContainer(ctx, id, 0); err != nil {
```
dfeigin-nv: These 2 impls are not idempotent. Meaning - if we killed a container and say we have a finalizer that calls this function to make sure a container did terminate, it can fail simply because we treat NotFound as an error.

agent/internal/runtime/runtime.go
```diff
@@ -30,6 +31,10 @@ type Runtime interface {
 	ResolveContainerIDByPod(ctx context.Context, pod, ns, ctr string) (string, error)
 	ResolveContainerByPod(ctx context.Context, pod, ns, ctr string) (int, *specs.Spec, error)
 	ResolveContainerImageID(ctx context.Context, id string) (string, error)
+	// TerminateContainer stops the container identified by its runtime ID with
+	// zero grace. Unlike signaling a resolved PID, the runtime performs the
+	// final identity check and cannot target a process that reused the PID.
```
dfeigin-nv: We check the scheme, not actual container identity, so let's consider a different wording here.

leeZardNav (dismissed):

dfeigin-nv (approved):

Issue #232, opened 2026-09-08
Restore compatibility: GPU identity precision, host OS equality, CRIU cpuinfo comparison, memory usage measurement, MIG restore target, application command and environment


danbar2:
## Summary

Define and enforce the concrete compatibility rules that gate a restore. This is
the rule-set issue under #231; #233 supplies the recorded provenance these rules
compare against.

Six distinct checks are bundled here because they share one call site and one
result shape — a single pre-restore validation pass that either admits the restore
or rejects it with a named property and both values. They are listed separately
below because each has its own semantics and its own strictness decision.

## Motivation

There is no compatibility gate today. A restore onto a mismatched node is attempted
and then fails somewhere inside CRIU, tar, or the CUDA runtime, producing errors
that describe a symptom rather than the cause (#36 has concrete transcripts).
Every check below corresponds to a way a restore can be silently wrong or
confusingly broken.

## The checks

**1. GPU identity precision.** `CUDAManifest.SourceGPUUUIDs`
(`agent/internal/types/manifest.go:133`) records source GPU UUIDs, and
`BuildDeviceMap` (`agent/internal/cuda/cuda.go:266`) maps source→target. It
identity-maps any UUID present on both sides, then pairs the remainder
**positionally** — the only guard is an equal-count check. UUIDs are node-unique,
so on any cross-node restore the identity pass matches nothing and every GPU is
paired by position. Nothing verifies the paired devices are the same model, and
nothing records what model the source GPUs even were. Decide what identity means:
UUID, model/SKU, compute capability, or driver ABI version — and record enough at
capture time to evaluate it. (The narrower model-ordering consequence is tracked
separately as #246.)

**2. Host OS equality.** Kernel version and distro affect CRIU's ability to restore
namespaces, mounts, and seccomp state. Define what must match exactly, what must
be within a range, and what is advisory.

**3. CRIU cpuinfo comparison.** CRIU can dump and compare CPU feature sets
(`criu cpuinfo dump` / `check`). A process restored onto a CPU missing an
instruction set it was compiled or JIT-optimized for will fault at an arbitrary
later point, not at restore. Wire this into capture and validate it at restore.

**4. Memory usage measurement.** Record the resident footprint — host and device —
at capture time, and refuse a restore onto a target whose limits or free device
memory cannot hold it. Today nothing is measured, so this fails as an OOM during
or after restore.

**5. MIG restore target.** A snapshot taken on a full GPU and restored onto a MIG
[...]

PR #318, opened 2026-09-16
✨ refuse a restore onto a different MIG partitioning or shape

### LINKED ISSUE #317: Restore compatibility: refuse a restore onto a different MIG profile
## Summary

Record which MIG profile each captured GPU was, and refuse a restore onto a
device with a different one. This is the MIG restore target check bundled in
 #232, carved out on its own because it is the one of the six whose rule needs no
further decision: a checkpoint is valid only on the partitioning it was taken on.

## Motivation

`gpu-model` compares the product name `nvidia-smi` reports, and for a MIG device
that name is the **parent** GPU's, not the slice's. The profile appears only in
`nvidia-smi -L`:

```
GPU 0: NVIDIA H100 80GB HBM3 (UUID: GPU-b...)
  MIG 3g.40gb     Device  0: (UUID: MIG-7089d0f3-293f-58c9-8f8c-5ea666eedbde)
  MIG 2g.20gb     Device  1: (UUID: MIG-56c30729-3...)
  MIG 1g.10gb     Device  2: (UUID: MIG-9d14fb21-4ae1-546f-a636-011582899c39)
```

Discovery queries `--query-gpu=gpu_uuid,name,driver_version`
(`agent/internal/cuda/cuda.go`), and `--query-gpu` exposes no profile field at
all, so the profile is never read and never recorded. Two invalid restores
therefore pass every gate we have:

- a checkpoint taken on a **whole** H100 restored onto a `1g.10gb` slice of an
  H100 — `gpu-model` sees the same product name on both sides, and `gpu-count`
  sees one device on both sides
- a checkpoint taken on `3g.40gb` restored onto `1g.10gb` — identical, for the
  same reason

Both reach CRIU and the CUDA runtime before failing on memory or SM capacity,
which is the failure mode the compatibility gates exist to prevent: an error
that describes a symptom rather than the cause.

## Scope

Two checks, not one, because the two rejections rest on different recorded data
and one of them can be enforced on artifacts captured before this change.

1. **Whole GPU against MIG slice**, from data already recorded. NVIDIA spells a
   MIG device's UUID with a `MIG-` prefix — in both the current `MIG-<uuid>` form
   and the older `MIG-GPU-<parent>/<gi>/<ci>` form — and `SourceGPUUUIDs` has
   been recorded since the first release. So this rejection needs no new capture
   field and applies retroactively.
2. **Slice shape against slice shape**, which does need a new field: read each
   visible device's profile from `nvidia-smi -L` at capture and store it per
   device beside the product name.

An artifact with no recorded profile leaves the shape comparison *unknown*, and
an unknown value admits, matching the policy `mustMatch` already documents in
`api/compat/checks.go` — a checkpoint captured before a value was ever recorded
has to stay restorable. Absence therefore means "not known", never "whole GPU";
the whole-versus-slice question is answered from the UUID instead, which is
always present.

3. **Publish** the profile on
   `PodSnapshotContent.status.source.devices.nvidia.instances[]`, so what a user
   reads is the value the gate compares.

## Goals

1. A whole-GPU checkpoint is refused on a MIG slice, and a MIG checkpoint on a
   whole GPU, by name and with both values — including for artifacts captured
   before this change.
2. A checkpoint is refused across differently shaped profiles.
3. An artifact carrying no recorded profile still restores rather than failing
   closed on a value it could never have carried.
4. Unit testable from recorded manifest data, with no MIG hardware and no GPU.

## Non-goals

- The remaining checks in #232 — host OS distro equality, CRIU cpuinfo
  comparison, memory footprint measurement, and command and environment
  comparison. Each needs a decision this one does not.
- Re-pairing devices to repair an incompatibility (#246).
- Reconfiguring MIG geometry on the target, or selecting a node by profile. This
  refuses a wrong restore; it does not arrange a right one.

## Acceptance criteria

- Both checks have unit tests proving the admit and the reject path, including
  whole GPU against slice in both directions.
- An artifact with a recorded UUID but no recorded profile is still refused
  across the whole-versus-slice boundary, and admits on shape.
- The published status value equals the value the shape check compares.
- No test requires live GPU or MIG hardware.

## Open question for validation

Whether `--query-gpu=gpu_uuid` reports the slice UUID or the parent GPU's inside
a container holding only a MIG device decides whether the profile read from `-L`
joins onto the discovered devices. The join is keyed on UUID, so if the two
disagree the profile stays unknown and the shape check admits rather than
wrongly refusing — safe, but it means the shape half needs confirming on real
MIG hardware before it can be called enforced.

leeZardNav:
Closes #317, the first check carved out of the rule-set in #232.

## What was wrong

`nvidia-smi` reports a MIG slice under its **parent card's** product name, and
`--query-gpu` exposes no MIG field at all. So a checkpoint taken on a whole H100
and restored onto a `1g.10gb` slice of an H100 saw the same model on both sides
and one device on both sides: `gpu-model` passed, `gpu-count` passed, and the
restore failed later inside CRIU or the CUDA runtime on memory or SM capacity.
The same held across differently shaped slices, `3g.40gb` onto `1g.10gb`.

That is the failure mode the gates added in #140 exist to prevent — an error
describing a symptom rather than the cause.

## What this does

Two named checks rather than one, because they rest on different recorded data.

| Check | Refuses | Applies to |
| --- | --- | --- |
| `mig-partitioning` | a whole GPU against a slice, either direction | checkpoints from **any** version |
| `mig-profile` | one slice shape against another | checkpoints captured after this change |

`mig-partitioning` reads the GPU UUIDs, which have been recorded since the first
release and carry a `MIG-` prefix only for a slice — in both the current
`MIG-<uuid>` form and the older `MIG-GPU-<parent>/<gi>/<ci>` one. So it needs no
new capture field and reaches artifacts captured before this change.

`mig-profile` needs the shape, which only `nvidia-smi -L` publishes, so capture
makes a second call and joins the profiles onto the devices discovery already
found, keyed on UUID rather than position. An artifact carrying no profile
leaves the shape unknown, and an unknown value admits — the policy `mustMatch`
already documents, so a checkpoint captured before a value was recorded stays
restorable.

Both checks compare *which* values are present rather than how many devices
share each, so losing a GPU is reported once by `gpu-count` instead of twice
under two names.

The profile is also published on
`status.source.devices.nvidia.instances[].migProfile`, so what a user reads is
the value the gate compares. CRD change, and the field is absent both for a
whole-GPU capture and for a slice captured before it existed.

## Reading order

Three commits, each building and passing its own tests standalone:

1. `feat(compat)` — the two rules and their tests, against synthetic data
2. `feat(cuda)` — reading the profile at capture and recording it
3. `feat(api)` — publishing it, plus the CRD, API reference and changelog

## Test plan

- `api` and `operator`: lint clean, tests pass
- `agent`: full suite and lint clean inside the `golang:1.27.1` container
  (`internal/runtime` is Linux-only, so neither runs natively on macOS)
- `make generate` reproduces the CRDs with the chart copy in sync;
  `verify-crds`, `helm-lint`, license headers, `fmt` and `tidy` all clean
- New coverage: both checks admit and reject, including whole GPU against slice
  in both directions and the legacy-artifact paths; the `nvidia-smi -L` parser
  against real multi-slice output, a MIG-disabled node and an empty listing;
  discovery surviving a failed listing; and the status projection

No unit test needs GPU or MIG hardware, per the acceptance criteria.

## Validated on a real GPU cluster

Validated at `35a3f3e`, published as
`v0.0.0-lirane-mig-profile-restore-gat-g35a3f3ed` and installed as the running
agent, operator and CRDs on a single-GPU cluster: **Tesla T4**, driver
`595.71.05`, kernel `7.0.0-1012-aws`, GPU Operator `v26.3.3`.

`8624180` landed after that run, and changes only how the partitioning value is
held in memory — the rendered strings the gate compares and reports are
unchanged, byte for byte, which the unmodified tests and a replay of the
comparison against these same T4 values both confirm.

**The new `nvidia-smi -L` call reaches the GPU.** Run through `nsenter` into a
live container's mount and PID namespaces, exactly as capture does:

```
$ nsenter --mount=/host/proc/28561/ns/mnt --pid=/host/proc/28561/ns/pid -- \
    nvidia-smi --query-gpu=gpu_uuid,name,driver_version --format=csv,noheader
GPU-698b3416-207f-8e8f-99b8-533f00103ca9, Tesla T4, 595.71.05

$ nsenter --mount=/host/proc/28561/ns/mnt --pid=/host/proc/28561/ns/pid -- \
    nvidia-smi -L
GPU 0: Tesla T4 (UUID: GPU-698b3416-207f-8e8f-99b8-533f00103ca9)
```

Both calls report the **same** UUID, which is the key the profiles are joined
on.

**A whole-GPU capture records and publishes no profile.** The manifest on the
volume, and the status the operator published from it:

```yaml
# cudaRestore.sourceGpus
- uuid: GPU-698b3416-207f-8e8f-99b8-533f00103ca9
  productName: Tesla T4

# status.source.devices
nvidia:
  driverVersion: 595.71.05
  instances:
  - productName: Tesla T4
```

`migProfile` is absent on both sides, which is what a whole GPU should record.
Restoring that artifact unmodified gives `Restored=True/RestoreSucceeded`, so
neither new check refuses a restore it should admit.

**`mig-partitioning` refuses a slice against a whole GPU.** Flipping only the
prefix of the recorded UUID, leaving `productName` and the `sourceGpuUuids`
list that the device map is built from untouched, so the restore would have
succeeded had the check not read it:

```
sourceGpus[0].uuid  before: GPU-698b3416-207f-8e8f-99b8-533f00103ca9
sourceGpus[0].uuid  after:  MIG-698b3416-207f-8e8f-99b8-533f00103ca9

Restored=False  reason=RestoreIncompatible
message: Refused restore; this node cannot run the checkpoint:
         mig-partitioning: source MIG slice, target whole GPU
```

with a matching `RestoreIncompatible` event on the pod. That is the
retroactive path: an artifact carrying a recorded UUID and no profile field at
all is still refused across the whole-versus-slice boundary, and `mig-profile`
stays quiet because the target offers no slice to compare a shape against.

**The e2e suite** is 23 of 24. `test_orphan_scanner_reclaims_uid_root` plants an
orphan artifact root and asserts it exists before the operator's cleanup scan
removes it, which races the 5-second scan interval the e2e setup configures; it
passed 3 of 4 reruns. It is a CPU-only test, and nothing in this change touches
the operator or the cleanup path.

## Not yet confirmed on hardware

No MIG-capable GPU was reachable — the only GPU node is a T4, which the GPU
Operator labels `nvidia.com/mig.capable=false`, and the real-GPU pools offer T4
or A10G, neither of which partitions. So `mig-profile`'s **shape against shape**
refusal is covered by unit tests alone; `mig-partitioning` is confirmed above in
both the admit and the refuse direction.

The open question narrows rather than closes. On a whole GPU the two
`nvidia-smi` calls agree on the UUID, as shown above, so the join works. What
remains unknown is whether `--query-gpu=gpu_uuid` reports the slice UUID or the
parent card's inside a container holding only a MIG device. If the two disagree
there, the profile stays unknown and `mig-profile` admits rather than wrongly
refusing — safe in the conservative direction, but it means the shape half wants
confirming on a real MIG node before it is called enforced. `mig-partitioning`
does not depend on this, since it reads UUIDs that are already recorded and
already correct.

## Review ownership

Touches `api/compat` (the rules), `agent/internal/cuda` (capture), and the
`PodSnapshotContent` status contract. The compat and CRD changes are the ones
worth the most attention.

## Summary by CodeRabbit

- **New Features**
  - Added MIG partitioning and profile compatibility checks for GPU checkpoint restoration.
  - Recorded MIG slice profiles in NVIDIA checkpoint metadata and GPU manifests.
  - Added MIG profile discovery during GPU capture.
  - Preserved compatibility with older artifacts lacking profile information.

- **Bug Fixes**
  - Rejects incompatible whole-GPU/sliced-GPU reversals and known MIG profile mismatches.
  - GPU discovery continues when optional MIG profile listing is unavailable.

- **Documentation**
  - Documented `migProfile` checkpoint metadata and availability rules.


api/compat/checks.go
```diff
@@ -317,6 +354,78 @@ func gpuModels(devices []GPUDevice) (string, bool) {
...
+// ignores allocation order, which is #246's concern rather than a mismatch.
+func distinctSorted(values []string) string {
+	seen := make(map[string]struct{}, len(values))
+	distinct := make([]string, 0, len(values))
+	for _, value := range values {
+		if _, ok := seen[value]; ok {
+			continue
+		}
+		seen[value] = struct{}{}
+		distinct = append(distinct, value)
```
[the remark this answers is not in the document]
leeZardNav: done in 35a3f3ed01

Confirmed and fixed. `gpu-count` compares only the total, so with the counts dropped `[1g, 1g, 3g]` against `[1g, 3g, 3g]` was three devices either way, one parent model either way, and the same set of shapes either way — admitted. `mig-profile` counts the shapes again, and `the same shapes in different proportions` covers exactly that case.

I kept the `mig-partitioning` summary categorical rather than counted. Counting there would refuse every plain change in GPU count on a cluster with no MIG at all, reporting `whole GPU x2` against `whole GPU x1` — which restates `gpu-count` under a name that mentions MIG to someone not using it. The mixed-set case you raised is still caught, because adding a slice changes the *set* of kinds: `one of two devices sliced instead of whole` asserts that, with the total held at two.

api/compat/compat.go
```diff
@@ -69,6 +69,11 @@ type GPUInfo struct {
 type GPUDevice struct {
 	UUID        string
 	ProductName string
+
+	// MIGProfile is the slice shape, such as "1g.10gb", and is empty both for a
+	// whole GPU and for a slice captured before it was recorded. Whether a
+	// device is a slice at all comes from UUID instead, which is always present.
```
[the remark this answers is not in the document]
leeZardNav: Declining this one. `compat.GPUDevice` is not a Kubernetes API type — it is a plain internal comparison struct with no JSON tags and no kubebuilder markers, and its sibling fields `UUID` and `ProductName` carry no `// +optional` either. The marker has no effect there, and adding it would imply the struct is part of a CRD schema when it is not.

The field that does reach the CRD is `NvidiaCheckpointSourceInstance.MIGProfile` in `api/v1alpha1/checkpoint_source_types.go`, and that one is marked `// +optional` with a `json:"migProfile,omitempty"` tag. The generated schema reflects it.

agent/internal/cuda/cuda.go
```diff
@@ -107,21 +107,71 @@ func DiscoverVisibleGPUs(ctx context.Context, hostProcPath string, pid int, time
...
+	if err != nil {
+		return compat.GPUInfo{}, fmt.Errorf("nvidia-smi via nsenter (pid %d) failed: %w", pid, err)
+	}
+	env := parseNvidiaSmiGPUs(string(output))
+
+	// --query-gpu exposes no MIG field at all, so the slice shape has to come
+	// from a second call. A failure here costs the shape, not the checkpoint:
+	// the profile stays unknown, and an unknown value admits a restore rather
+	// than refusing one.
+	listed, err := nsenterNvidiaSMI(ctx, hostProcPath, pid, "-L")
```
leeZardNav: we don't want the checkpoint will fail on any fact retrieval. probable in this point the restore as well

agent/internal/cuda/cuda.go
```diff
@@ -107,21 +107,91 @@ func DiscoverVisibleGPUs(ctx context.Context, hostProcPath string, pid int, time
...
+	if err != nil {
+		return env, nil
+	}
+	return withMIGProfiles(env, parseNvidiaSmiMIGProfiles(string(listed))), nil
```
danbar2: **The profile join can't match: `--query-gpu` returns the parent's UUID, `-L` returns the slice's.**

`withMIGProfiles` keys on `device.UUID`, which came from `--query-gpu=gpu_uuid`. That query enumerates *GPUs*, and a MIG slice isn't one — inside a container holding a slice, the parent card is still what gets enumerated. The container toolkit docs show exactly this for a container given a single MIG device (`--gpus '"device=0:0"'`):

```
GPU 0: A100-SXM4-40GB (UUID: GPU-e86cb44c-6756-fd30-cd4a-1e6da3caf9b0)
  MIG 3g.20gb Device 0: (UUID: MIG-c7384736-a75d-5afc-978f-d2f1294409fd)
```

So `--query-gpu=gpu_uuid` yields `GPU-e86cb44c-…` while `parseNvidiaSmiMIGProfiles` keys its map on `MIG-c7384736-…`. The two sets never intersect: `MIGProfile` stays empty on every device, `gpuMIGProfiles` returns `!ok`, and `mig-profile` admits every shape change — including the `3g.40gb` → `1g.10gb` case in the description. Neither spelling of the MIG UUID helps, since both the modern `MIG-<uuid>` and the legacy `MIG-GPU-<parent>/<gi>/<ci>` differ from the parent's `GPU-<uuid>`.

This is the same fact the description already states: "`--query-gpu` exposes no MIG field at all" is precisely why the second call exists — but the join then reaches back to that MIG-blind query for its key.

`TestDiscoverVisibleGPUsRecordsTheMIGProfile` passes because its fake `--query-gpu` prints `MIG-7089d0f3-…`, which real nvidia-smi does not do; the same fixture's `-L` output names `GPU-b1c4` as the parent, so the two halves contradict each other. Worth fixing that fixture first — it should fail against the current code.

`-L` already carries the parent UUID, each slice UUID and the profile in one output, so the pairing can be derived from it alone. That also settles which UUID gets recorded, which is the subject of the `checks.go` comment.

Refs: [Specialized Configurations with Docker](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/docker-specialized.html) · [MIG Device Names](https://docs.nvidia.com/datacenter/tesla/mig-user-guide/mig-device-names.html)

api/compat/checks.go
```diff
@@ -286,35 +329,148 @@ var driverMinimumCheck = check{
...
+		return wholeGPU
+	default:
+		return ""
+	}
+}
+
+// gpuPartitioning reads UUID rather than MIGProfile so that it also holds for
+// an artifact captured before any profile was recorded: the UUIDs have been
+// recorded since the first release, and only a slice carries the MIG- prefix.
+func gpuPartitioning(devices []GPUDevice) (partitioning, bool) {
```
danbar2, line 391: **The `MIG-` prefix only ever appears on one of the three discovery paths.**

`gpuPartitioning` infers slice-vs-whole-GPU from the UUID prefix, but which producer supplies that UUID varies:

- **PodResources, device plugin `single` strategy** — device IDs are MIG UUIDs. Prefix present, check works.
- **PodResources, `mixed` strategy** — `GetPodGPUUUIDs` (cuda.go:86) matches only `nvidia.com/gpu`, so slices advertised as `nvidia.com/mig-1g.10gb` yield nothing and discovery falls back to nvidia-smi, recording the parent `GPU-…`. A slice is then read as a whole GPU.
- **DRA** — `dra.go:207` filters through `gpuUUIDPattern`, which matches `GPU-<uuid>` only, so a MIG UUID is dropped before it can be recorded at all.

Both error directions are live: a mixed-strategy source against a whole-GPU target passes the gate — the headline case this PR exists to refuse — and a whole-GPU source against a mixed-strategy target fails it, a terminal `GateInspect` refusal of a valid restore that only the skip annotation can clear.

agent/internal/cuda/cuda.go
```diff
@@ -107,21 +107,91 @@ func DiscoverVisibleGPUs(ctx context.Context, hostProcPath string, pid int, time
...
+	}
+	env := parseNvidiaSmiGPUs(string(output))
+
+	// --query-gpu exposes no MIG field at all, so the slice shape has to come
+	// from a second call. A failure here costs the shape, not the checkpoint:
+	// the profile stays unknown, and an unknown value admits a restore rather
+	// than refusing one.
+	listed, err := nsenterNvidiaSMI(ctx, hostProcPath, pid, "-L")
+	if err != nil {
+		return env, nil
```
danbar2, line 129: Could you add a log line here? The `-L` failure is dropped silently, so a node where the listing fails consistently records no profile on every capture with nothing to say why. A V(1) line with pid and error, matching the sibling path at cuda.go:264, would cover it — and would also make the case where the first nvidia-smi call eats most of the shared 30s deadline visible rather than silent.

danbar2:
We need to test actual support for mig devices, and update docs accordingly

agent/internal/cuda/cuda.go
leeZardNav: done in 3dc1e4025a

agent/internal/cuda/cuda.go
leeZardNav: done in b7a066d759

Issue #324, opened 2026-09-16
SNEP: Coordinated multi-Pod snapshot and restore


julienmancuso:
## Summary

Track the Snapshot Enhancement Proposal (SNEP) for coordinated checkpoint and
restore of a fixed group of mutually dependent Pods, including multi-node
inference replicas.

Proposal: #325

## Motivation

Snapshot currently captures and restores each Pod independently. Distributed
workloads additionally require immutable group membership, group-wide preflight
and release barriers, a complete source-to-target network identity map, atomic
checkpoint publication, and failure handling that never exposes a partial
group as ready.

The first qualification target is one TensorRT-LLM dense TP=2 replica across
two nodes with NCCL Socket and one GPU per Pod.

## Proposed scope

- Add `PodGroupSnapshot`, `PodGroupSnapshotContent`, and `PodGroupRestore`.
- Reuse existing `PodSnapshot` and `PodSnapshotContent` objects as subordinate
  per-Pod artifacts.
- Aggregate compatibility preflight before any member reconstructs.
- Hold every reconstructed member behind a durable group release barrier.
- Preserve and remap qualified in-group TCP sessions.
- Compose with Grove gang and topology-aware scheduling without moving
  scheduling policy into Snapshot.
- Define failure, retry, cleanup, observability, test, and graduation contracts.

The full design is published in #325 under
`docs/proposals/324-multi-pod-snapshot-restore/README.md`.

PR #325, opened 2026-09-16
docs: propose coordinated multi-pod snapshot restore

julienmancuso:
## Summary

- add SNEP-324 for coordinated multi-Pod checkpoint and restore
- define additive `PodGroupSnapshot`, `PodGroupSnapshotContent`, and
  `PodGroupRestore` APIs
- specify checkpoint, preflight, reconstruction, release, validation, failure,
  scheduling, monitoring, and qualification contracts

Tracking issue: #324

## Validation

- `make update-toc`
- `make verify-toc`
- `git diff --check`


docs/proposals/324-multi-pod-snapshot-restore/README.md
```diff
@@ -0,0 +1,802 @@
...
+
+The durable member phases are:
+
+```text
+Pending -> Restoring -> Restored
+```
+
+![PodGroupRestore state machine](diagrams/restore-state.svg)
+
+Group restore deliberately retains the standalone restore completion semantic.
```
oleg-kushniriov: No component coordinates resume across members

The Motivation and requirement 31 promise that established member-to-member TCP sessions survive restore. That only works if every member's sockets are rebuilt via `TCP_REPAIR` before any member resumes and sends a segment.

This paragraph and requirement 22 let each node agent resume its restored process and report `nvidia.com/Restored=True` as soon as its own CRIU restore succeeds. Nothing in the restore flow, the member phases, or the normative requirements names a component that waits for all members to be rebuilt and then releases them together. If rank 0 resumes while rank 1 is still restoring, rank 1's placeholder namespace answers with RST and the connection is lost on both sides.

Please name the component that owns the group-wide resume, and what the agent does while it waits, or drop the preserved-session claim.

docs/proposals/324-multi-pod-snapshot-restore/README.md
```diff
@@ -0,0 +1,802 @@
...
+
+1. Snapshot loads the complete group checkpoint record.
+2. The workload owner asks Grove or another workload manager to create and
+   schedule one restore-shaped target Pod for every logical member.
+3. After every target Pod exists, the caller creates `PodGroupRestore` with the
+   exact Pod names and UIDs.
+4. Snapshot waits until all targets are scheduled and their placeholders are
+   available, validates the one-to-one mapping, and constructs the complete
+   source-to-target network identity map.
+5. Snapshot creates or activates durable per-member restore work approximately
```
oleg-kushniriov: The document has the owner create every target Pod first and create PodGroupRestore afterwards. If those Pods carry the annotation, the single-Pod restore has already started before the group exists. If they do not, the Design does not say what the agent keys on instead. Step 5 says Snapshot "creates or activates durable per-member restore work", but there is no leaf restore object in the API, and nothing on a PodSnapshot marks it as group-owned, so "reject standalone restore" has nowhere to run. The document also does not say how the complete identity map reaches the agent.

docs/proposals/324-multi-pod-snapshot-restore/README.md
```diff
@@ -0,0 +1,802 @@
...
+6. Per-member success **MUST NOT** be exposed as group success.
+7. Snapshot **MUST** fan out subordinate member operations without
+   intentionally serializing one member behind another.
+8. Snapshot **MUST NOT** claim atomic checkpoint, atomic restore, or rollback
+   of work that has crossed a per-Pod destructive boundary.
+9. Every applicable CRIU restore **MUST** receive the complete source-to-target
+   network identity map.
+10. Missing or unsupported identity mappings **MUST** fail before process
+    reconstruction.
+11. Controller-observable group identity, readiness, node-agent, storage,
```
oleg-kushniriov: As written, one member can pass its local check and go ahead while a sibling fails its own. On restore that is one rank resumed and one refused as RestoreIncompatible, which the test plan lists as an expected outcome. On checkpoint it is worse: one source process has already been dumped and killed when the sibling refuses, and the document accepts this and hands recovery to the owner. In both cases a check that changes nothing could have stopped the group before anyone did anything irreversible.

docs/proposals/324-multi-pod-snapshot-restore/README.md
```diff
@@ -0,0 +1,802 @@
...
+that target and reports `RestoreFailed`. A successfully restored sibling stays
+`nvidia.com/Restored=True`; Snapshot does not delete, restart, terminate, or
+fence it solely because another member failed. `PodGroupRestore` reports
+Failed with every member outcome, and the workload owner decides how to repair
+or recreate the distributed replica.
+
+Group operation state survives controller restart. Reconciliation does not
+duplicate leaf creation or already completed per-Pod work. Operation identity,
+member identity, observed generation, leaf references, and durable per-member
+outcomes fence repeated work. Cancellation or deletion stops work that has not
```
oleg-kushniriov: It does not say whether deleting PodGroupSnapshot deletes the cluster-scoped content and the leaf objects, whether a `PodGroupRestore` may reference a group snapshot whose namespaced object is gone, how finalizers are ordered, or what a caller sees after deleting an operation that is in flight. CodeRabbit's cancellation comment is one part of this.

docs/proposals/324-multi-pod-snapshot-restore/diagrams/restore-state.mmd
```diff
@@ -0,0 +1,17 @@
+%% SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+%% SPDX-License-Identifier: Apache-2.0
+
+stateDiagram-v2
+    [*] --> Pending
+    Pending --> TargetsAvailable: every target name and UID resolves
```
oleg-kushniriov: It is not clear whether the diagram shows the member or the group. The waiting condition also differs: the text says restore waits until every target is scheduled and its placeholder is running, while the diagram's transition only requires names and UIDs to resolve, which happens well before that. The restore sequence diagram has no wait at all.

docs/proposals/324-multi-pod-snapshot-restore/README.md
```diff
@@ -0,0 +1,802 @@
...
+metadata:
+  name: trtllm-replica-a
+spec:
+  members:
+    - id: rank-0
+      source:
+        podRef:
+          name: trtllm-rank-0
+          uid: "<source-pod-uid>"
+          containers: [engine]
```
oleg-kushniriov: The design does not say whether the one-container limit carries over or is being lifted, and does not say how containers are mapped on restore. The existing single-Pod API allows exactly one container per snapshot, and the single-Pod restore contract already has a source-to-destination container map that group targets could reuse.

docs/proposals/324-multi-pod-snapshot-restore/README.md
```diff
@@ -0,0 +1,802 @@
...
+| A complete IP map does not by itself preserve established cross-Pod sockets. | Distinguish in-group from external peers, retain qualified in-group connections, remap both endpoints, and fail closed when membership is ambiguous. |
+| One target may restore while another fails. | Preserve each node agent's existing restore behavior, fail the aggregate group attempt, expose every member outcome, and let the workload owner recreate or repair the group. |
+| A controller or node agent can restart during fan-out. | Persist operation identity, member identity, observed generation, leaf references, and per-member outcomes. Make reconciliation idempotent. |
+| Grove `startsAfter` currently waits for Kubernetes Pod Ready and can prevent all restore placeholders from existing. | Omit or rewrite inter-member `startsAfter` dependencies for restore-shaped Pods, or add a Grove milestone that distinguishes placeholder availability from workload readiness. |
+| Successful process restore does not imply serving readiness. | Complete `PodGroupRestore` from the existing `nvidia.com/Restored` conditions. The workload owner separately observes Pod readiness and controls serving registration. |
```
oleg-kushniriov: The completion sentinel that drives the startup probe should be written only once the whole group is released, so all member Pods turn Ready together rather than one by one.

docs/proposals/324-multi-pod-snapshot-restore/README.md
```diff
@@ -0,0 +1,802 @@
...
+## Design Details
+
+### API
+
+The API adds three resources in `nvidia.com/v1alpha1`. The field names below
+are the proposed contract; implementation review may refine individual names
+without collapsing the separation between checkpoint artifact and restore
+attempt.
+
+#### `PodGroupSnapshot`
```
oleg-kushniriov: This might be read as a snapshot of a scheduler group.
Consider `PodSetSnapshot / PodSetRestore`

docs/proposals/324-multi-pod-snapshot-restore/README.md
```diff
@@ -0,0 +1,802 @@
...
+31. A preserved-session restore **MUST** distinguish in-group TCP peers from
+    external peers and **MUST NOT** apply the standalone external-peer
+    disconnection policy to a qualified in-group connection.
+32. `PodGroupRestore` **MUST** become Ready from successful per-Pod
+    `nvidia.com/Restored` outcomes and **MUST NOT** depend on Kubernetes Pod
+    Ready or serving readiness.
+
+### Configuration
+
+This proposal introduces no new user-facing runtime configuration and does not
```
oleg-kushniriov: The operator and agent have no feature gates. Features are enabled by installing CRDs and setting Helm values.

Consider saying the group APIs are enabled by installing their CRDs and, if needed, a Helm value, and drop the feature-gate wording.

docs/proposals/324-multi-pod-snapshot-restore/README.md
```diff
@@ -0,0 +1,802 @@
...
+5. If every leaf succeeds, Snapshot publishes the group checkpoint record as
+   Ready. If any leaf fails, Snapshot marks the group Failed and does not
+   publish a usable group checkpoint.
+
+The workload must complete any collective application quiescence or
+communicator preparation before its Pods become snapshot-ready. Snapshot's
+group validation reduces predictable partial failure but cannot prove that
+every CRIU or CUDA capture will succeed.
+
+Checkpoint capture is destructive: Snapshot terminates the captured source
```
oleg-kushniriov: Checkpointing a Pod kills its process at the end. The two ranks are checkpointed around the same time, but one always finishes first. For a moment, rank A is dead while rank B is still being copied and still has an open connection to A.

That is survivable only because A's Pod is still there, silently swallowing B's packets. If the owner deletes A's Pod too early, its IP can go to some new Pod, which answers B with "go away", and B's connection is broken before it is saved. Also, Kubernetes will normally restart A's killed container, and the new A will start dialing B while B is still being copied.

State that source Pods must not be deleted and their containers must not restart until the group checkpoint is published or failed, and say how the owner learns that moment.

docs/proposals/324-multi-pod-snapshot-restore/README.md
```diff
@@ -0,0 +1,802 @@
...
+target-specific map is a restore work order and belongs to
+`PodGroupRestore`, or to a durable leaf work object derived from it. Pod
+annotations are not the source of truth for the map or aggregate member state.
+
+Missing, duplicate, ambiguous, or unsupported mappings fail restore before
+process reconstruction begins.
+
+### Failure and Recovery
+
+A failure in any mandatory member fails the group operation.
```
oleg-kushniriov: With Pods pinned by UID, a deleted or preempted target can never come back under the same identity, and an unschedulable one may sit in Pending indefinitely, for example when the cluster is short of GPUs. The document gives the group no way to notice either case and no deadline, so the group restore waits forever while every other placeholder holds its GPU. The same applies on the checkpoint side to a source Pod that vanishes between validation and dump. The single-Pod job API already has an active deadline; the group objects have nothing comparable.

Define a terminal failure for a member whose pinned Pod no longer exists or cannot be scheduled, and add an optional deadline to both group objects.

julienmancuso:
Thanks @oleg-kushniriov — I pushed revision 36d788e. The review exposed a real contradiction: the proposal promised preservation of established in-group sessions while allowing every member to resume independently.

The revision makes network and release semantics capability-profile-specific:

- A communicator-reconstruction profile may complete members independently, but only after qualification proves that the backend safely reconstructs its communicator.
- A session-preservation profile requires coordinated network/session handling, a durable per-member `Prepared` state, and authorization from the `PodSetRestore` controller before any member is released.

The revision also:

- renames the APIs to `PodSetSnapshot`, `PodSetSnapshotContent`, and `PodSetRestore`;
- introduces a durable `PodRestore` leaf for target identity, exact ownership, container mapping, profile inputs, and per-member status;
- requires every member to complete admission before any member crosses its declared destructive boundary, while explicitly retaining non-atomic failure semantics after release;
- specifies exact UID binding, the V1 one-container-per-member limit, and source-to-destination container mappings;
- adds operation deadlines, missing-target behavior, deletion ordering, and retained failure evidence;
- states the source-Pod hold obligation during capture;
- corrects and relabels the checkpoint, restore, scheduling, and state diagrams.

Six decisions are intentionally still open: the initial qualified profile(s), the local leaf contract aligned with SNEP-295 (cc @galletas1712 ), prepared-sibling failure disposition, source-hold enforcement, cancellation semantics, and peer-mapping scope for communicator reconstruction.

I've replied on each thread with either the implemented correction or the remaining decision.

docs/proposals/324-multi-pod-snapshot-restore/README.md
julienmancuso: Agreed and implemented in 36d788e. The API kinds are now `PodSetSnapshot`, `PodSetSnapshotContent`, and `PodSetRestore`, with `PodRestore` as the per-member restore leaf. This avoids implying that the Snapshot object is itself a scheduler PodGroup.

docs/proposals/324-multi-pod-snapshot-restore/README.md
julienmancuso: Agreed and implemented in 36d788e. The feature-gate wording is removed. The rollout section now describes the actual mechanism: install the four additive CRDs and enable the controllers through chart/configuration values while leaving the standalone APIs unchanged.

docs/proposals/324-multi-pod-snapshot-restore/diagrams/restore-state.mmd
julienmancuso: Agreed and fixed in 36d788e. The state diagram now separates the per-member `PodRestore` state machine from the group-level `PodSetRestore` aggregation. Leaves may exist in `WaitingForPlaceholder`; admission does not begin until each pinned target is scheduled and its placeholder is available. The restore sequence and Grove diagram show the same ordering.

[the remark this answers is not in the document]
julienmancuso: Fixed in 36d788e. The diagrams now show `PodRestore` leaves being created first, waiting per member for placeholder availability, and only entering admission after all required placeholders are available. This matches the revised prose and gives the controller per-target waiting and deadline visibility.

docs/proposals/324-multi-pod-snapshot-restore/README.md
julienmancuso: Valid. The revision implements the proposed direction with a namespaced, durable `PodRestore` leaf. It carries the exact owning set UID, pinned target name/UID, explicit container mapping, capability profile, profile-required identity data, and per-member phases/reasons. Leaves are created before placeholders are ready, so they can expose `WaitingForPlaceholder`.

Group restore does not use `nvidia.com/restore-from` as its source of truth. The standalone annotation continues to name a `PodSnapshot`; controller and admission checks reject using that path to activate resolved content owned by a set.

The exact release field and restart-safe local contract remain part of the SNEP-295 alignment, so I’m leaving this thread open for review of the direction.

docs/proposals/324-multi-pod-snapshot-restore/README.md
julienmancuso: Valid. The revision now explicitly preserves the existing one-container limit for every V1 set member and adds an explicit source-to-destination `containerMapping` for each restore target. The set controller copies the validated mapping into the member's `PodRestore`; it does not rely on same-name defaults. Existing standalone restore semantics remain unchanged.

Multi-container set capture is still out of scope, so I’m leaving this open for agreement on the V1 boundary.

docs/proposals/324-multi-pod-snapshot-restore/README.md
julienmancuso: Valid. The revision now separates deletion from cancellation and specifies the V1 deletion behavior:

- deleting a set operation stops work that has not been activated but cannot undo work past a destructive boundary;
- controller-managed finalizers sequence leaf cleanup before content cleanup;
- deleting a `PodSetSnapshot` rejects new restores and waits for active restores before deleting leaves/content;
- deleting a `PodSetRestore` removes its `PodRestore` leaves but never deletes target Pods;
- V1 does not retain set content after its namespaced snapshot is deleted.

The independent cancellation model remains an explicit open decision, so I’m leaving the thread open.

docs/proposals/324-multi-pod-snapshot-restore/README.md
julienmancuso: Valid. Both set operation specs now include a single operation deadline. A pinned Pod UID that disappears is terminal immediately; an unscheduled or temporarily unschedulable target remains a visible waiting condition and becomes `DeadlineExceeded` only when the operation deadline expires. The same frozen-UID and deadline rules apply to capture members.

The `PodRestore` phases and group aggregation retain the per-member reason rather than collapsing it into an opaque group failure. Leaving this open for review of those semantics.

docs/proposals/324-multi-pod-snapshot-restore/README.md
julienmancuso: Valid. The revision adds a held admission phase: every leaf must finish its local admission checks before the set controller releases any leaf across its declared destructive boundary. If a member refuses before release, admitted siblings are abandoned without becoming unresumable. After release, failure remains explicitly non-atomic; the set is Failed and all member outcomes are retained, but rollback is not promised.

The leaf—not the group controller—must declare where its destructive boundary begins, since CUDA-interposition preparation may cross it earlier than CRIU dump-and-kill. The durable, restart-safe admission/abandon contract is still being aligned with SNEP-295, so this stays open.

docs/proposals/324-multi-pod-snapshot-restore/README.md
julienmancuso: This was the central contradiction, and the revision no longer makes one unconditional preserved-session promise.

Each qualified backend profile must choose one of two policies:

1. `CommunicatorReconstruction`: the backend is proven to reconstruct its communicator safely, so a group release barrier is not required.
2. `SessionPreservation`: the `PodSetRestore` controller owns the group barrier; each node agent reconstructs locally, records its `PodRestore` as `Prepared`, and remains held until every member is Prepared and the controller authorizes release.

The second path now names both the owner and the waiting state. What remains open is the restart-safe leaf release input and which profile each initial backend can actually qualify. SNEP-295 defines local reconstruction but not yet that external release contract. I’m leaving this thread open until the backend experiment and leaf contract land.

docs/proposals/324-multi-pod-snapshot-restore/README.md
julienmancuso: Agreed for the session-preservation profile. The revision now requires that no agent write `restore-complete` or report `nvidia.com/Restored=True` until every member is `Prepared` and the set controller has authorized release. A communicator-reconstruction profile retains the existing local-completion semantics after its own reconstruction succeeds.

The enforceable invariant is coordinated authorization, not literally simultaneous Pod readiness—agents observe release and startup probes asynchronously. The concrete release-input mechanism is still open with the SNEP-295 contract, so this thread remains open.

docs/proposals/324-multi-pod-snapshot-restore/README.md
julienmancuso: Valid. The revision now makes the constraint normative: from admission until the set operation terminates, the workload owner must keep the exact source Pod UIDs from restarting, being replaced, or resuming normal execution. Set status is the durable signal for when that hold may end; set publication alone is not a source-process-lifetime signal.

The enforcement mechanism is intentionally still open. A finalizer preserves the API record but does not stop kubelet termination or restart, and scheduling cannot enforce it. The remaining candidates are a workload-owner/Grove hold contract or a snapshot-aware wrapper with a local terminal signal. Transport timers may also require coordinated repair or isolation rather than process hold alone. Leaving this open until that contract is selected.

docs/proposals/324-multi-pod-snapshot-restore/README.md
```diff
@@ -0,0 +1,802 @@
...
+fence it solely because another member failed. `PodGroupRestore` reports
+Failed with every member outcome, and the workload owner decides how to repair
+or recreate the distributed replica.
+
+Group operation state survives controller restart. Reconciliation does not
+duplicate leaf creation or already completed per-Pod work. Operation identity,
+member identity, observed generation, leaf references, and durable per-member
+outcomes fence repeated work. Cancellation or deletion stops work that has not
+started and records outstanding artifact cleanup, but it cannot undo a capture
+or restore that crossed its existing per-Pod destructive boundary.
```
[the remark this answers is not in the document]
julienmancuso: Valid. The revision now specifies deletion independently: it stops non-activated work, cannot undo work past a destructive boundary, sequences cleanup with controller-managed finalizers, and retains per-member outcomes long enough to explain partial progress.

It does not yet claim a durable `Cancelled` outcome. Status on an object being deleted is not itself a durable record, so deletion-as-cancellation would require retained evidence. The alternative is an explicit request on a surviving operation object. Terminal cancellation versus reversible suspension remains Open Design Decision 5, and this thread stays open until that choice is made.

Issue #295, opened 2026-09-12
cuInterpose: checkpointing CUDA peer mappings


galletas1712:
### What would you like to be added?

Add **cuinterpose**, a library loaded into each workload process that intercepts selected CUDA calls and records how GPU memory is shared between processes. A coordinator uses those records to reconstruct the shared memory after checkpoint and restore. The proposed implementation is a **small C frontend, a Rust backend, and a static Rust coordinator**. An optional PageBroker allocation-transfer path stores shared allocation contents in explicit checkpoint files.

The initial scope is CUDA virtual memory management (VMM) allocations shared through POSIX file descriptors (FDs), including allocations bound into CUDA multicast objects. Participating processes run on one node. The complete captured process tree may be restored on another compatible node, where the processes share the reconstructed allocations locally.

### Terms used in this proposal

CUDA VMM separates allocating GPU memory from choosing the virtual addresses through which a process accesses it. For example, process A can create an allocation, export it, and map it at address A1. Process B can import that allocation and map the same GPU memory at address B1. Restore must recreate one shared allocation and reconnect both processes at their respective original addresses.

| Term | Meaning here |
| --- | --- |
| Allocation / backing | An allocation is a CUDA memory object. Its backing is the physical GPU memory holding its bytes. Multiple processes can hold references to the same backing. |
| Virtual mapping / access grant | A mapping associates a process's GPU virtual address range with a range of an allocation. Access grants specify which GPUs can read or write that mapping. |
| Creator / importer / peer | The creator is the process that originally created an allocation. An importer obtains a reference to that allocation through an exported handle. A peer is another participating workload process. |
| Driver handle / logical handle / allocation ID | A driver handle identifies a CUDA object within one process. The shim gives the application a logical handle and resolves it through its own table to the current driver handle. An allocation ID is the shim's persistent identifier for the same allocation across processes and restore. |
| Export FD / ticket | A CUDA export FD lets another process import a driver object. Cuinterpose gives the application a ticket FD: a sealed, memory-backed file containing the information needed to request an export FD from the creator. |
| Unicast / multicast | Unicast refers here to ordinary VMM allocations and mappings. A CUDA multicast object connects memory bindings on a team of GPUs and exposes multicast mappings. The device team is the set of attached GPUs; a binding associates a range of member memory with a range of the multicast object. |
| Topology | The recorded relationships between allocations, their creators and importers, virtual mappings, access grants, and multicast devices and bindings. |
[...]

PR #326, opened 2026-09-16
feat(cuinterpose): C preload frontend and private ABI

galletas1712:
## Summary

Refs #295 (approved).

Add the glibc-based C preload frontend, NVIDIA-typed private ABI, and explicit ELF export list. Intercept direct calls and CUDA symbol-query paths, including synchronous malloc and memory-IPC entry points.

Successful function lookup only substitutes compatible wrappers; it does not load the backend or start runtime services. The `cuInit` wrapper calls the real driver first, returns CUDA failures unchanged, and starts the shim only after CUDA initialization succeeds. Memory calls forward to the backend, which requires a current-process runtime.

Private ABI version 3 contains the resolver, initialization callback, and memory callbacks. Property-only allocation and multicast granularity queries go directly to CUDA. There are no fork callbacks or library-load PID in the ABI. Allocation tracking, process ownership checks, and CUDA IPC behavior are implemented in later slices.

Concurrent callers may load the backend independently; glibc serializes DSO construction. Atomic publication retains the winning callback table and library reference. A thread-local guard refuses constructor reentry without waiting under the loader lock. The small `pthread_once` only discovers glibc's real `dlsym`; it does not start runtime workers. Retained CUDA providers remain loaded for the lifetime of cached function pointers.

The Rust workspace uses `panic = "abort"` in debug and release builds, including the standalone coordinator. Ordinary returned errors retain their normal handling.

## Validation

Validated on the assembled stack at #338: `make -C agent/cmd/cuinterpose test`, `make check`, `make test`, and `make build`. The packaged suite includes formatting, strict Clippy, GNU/musl tests, C/Rust ABI checks, and headless regressions. Optional CUDA helper tests were skipped for unavailable development dependencies. Physical-GPU and cross-node tests were not run for this revision.

Intermediate PRs remain source layers, not independently buildable backends. Ship matching frontend/backend artifacts (private ABI version 3); the checkpoint wire/state format is version 3.

## Summary by CodeRabbit

- **New Features**
  - Added CUDA interposition for memory allocation, IPC, virtual memory, and multicast operations.
  - Added dynamic CUDA symbol and driver entry-point resolution.
  - Added backend initialization and lifecycle handling.
  - Added validation with clear responses for unsupported or invalid CUDA requests.
  - Added a shared interface for compatible frontend and backend operation handling.

- **Build & Validation**
  - Added automated build, formatting, linting, and test workflows.
  - Improved container build configuration and service usage validation.


agent/cmd/cuinterpose/frontend/frontend.c
```diff
@@ -0,0 +1,364 @@
...
+// Published nodes and their dlopen references live until exit. Readers need no
+// lock, including after quiescent fork; no loader call runs under a shim mutex.
+struct Provider {
+    void *handle;
+    struct Provider *next;
+};
+static _Atomic(struct Provider *) providers;
+
+static void initialize_resolver(void) {
+    real_dlsym = (void *(*)(void *, const char *))dlvsym(RTLD_NEXT, "dlsym", "GLIBC_2.34");
```
galletas1712: let's define the GLIBC string/version as a constant.

agent/cmd/cuinterpose/frontend/frontend.c
```diff
@@ -0,0 +1,364 @@
...
+            while (*suffix >= '0' && *suffix <= '9')
+                ++suffix;
+        }
+        if (!*suffix)
+            return (int)i + 1;
+    }
+    return 0;
+}
+
+// Qualify explicit CUDA handles before substitution; retain the actual provider
```
galletas1712: this comment is jargony and not clear.

galletas1712: like what does "qualify explicit cuda handles" even mean? what does "provider" mean? these are vague

agent/cmd/cuinterpose/frontend/frontend.c
```diff
@@ -0,0 +1,398 @@
...
+
+static const struct BackendAbi *backend(void) {
+    if (atomic_load(&failed) || atomic_load(&backend_unavailable))
+        return NULL;
+    const struct BackendAbi *api = atomic_load(&backend_api);
+    if (api)
+        return api;
+    // Refuse same-thread constructor reentry before dlopen has finished. Other
+    // threads may load independently: glibc serializes DSO construction, and
+    // the ABI handshake is idempotent. Never wait under a shim lock around dlopen.
```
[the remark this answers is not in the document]
galletas1712: Fixed in 92eab4a. `backend()` checks genuine `failed` first, then the published table, then `backend_unavailable`. A failed private load rereads the table once before latching unavailability. No new lock, waiting, or retry loop.

Deterministic regressions cover both publication orders and sticky failure without a winner. They live in #338 to keep this frontend PR small. Both race cases reproduce against the previous frontend and pass with this change. The loader suite passed 46 cases across glibc 2.36/2.41 on the cluster; qualification is CPU/fake-CUDA, not real-GPU/model validation.

agent/cmd/cuinterpose/frontend/frontend.c
```diff
@@ -0,0 +1,391 @@
...
+    ENTRY(cudaGetDriverEntryPointByVersion)
+    ENTRY(cudaGetDriverEntryPointByVersion_ptsz)
+#undef ENTRY
+    return NULL;
+}
+
+API void *dlsym(void *handle, const char *name) {
+    void *address = lookup(handle, name);
+    if (!address || !retain_cuda_library(handle, address))
+        return atomic_load(&failed) ? NULL : address;
```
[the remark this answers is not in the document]
galletas1712: Fixed in 92eab4a. The existing retention function now distinguishes out-of-scope addresses, retained CUDA addresses, and unavailable CUDA addresses. `dlsym` preserves unrelated/out-of-scope lookup results after shim failure, while eligible CUDA lookups and cached CUDA wrappers still fail closed. No CUDA passthrough or new process-termination policy was added.

The #338 regression injects a genuine retention failure after backend publication, checks unrelated libc/plugin lookup survival, verifies existing handle/namespace exclusions, and checks that CUDA remains blocked. The old frontend reproduces the unrelated-lookup failure; the new frontend passes on glibc 2.36/2.41.

agent/cmd/cuinterpose/frontend/frontend.c
```diff
@@ -0,0 +1,391 @@
...
+    if (atomic_load(&failed) || atomic_load(&backend_unavailable))
+        return NULL;
+    const struct BackendAbi *api = atomic_load(&backend_api);
+    if (api)
+        return api;
+    // A constructor may own the loader lock while another thread initializes.
+    // Reentry must return transient not-initialized rather than wait on it.
+    bool expected = false;
+    if (!atomic_compare_exchange_strong(&initializing, &expected, true))
+        return NULL;
```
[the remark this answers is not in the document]
galletas1712: The original global `initializing` CAS has been removed. Concurrent frontend callers can independently load/register the same backend and atomically publish the immutable ABI table; only same-thread loader reentry is refused. The handshake performs no runtime startup.

We deliberately did not add a waiting loop: a DSO constructor can hold glibc's loader lock while another initializer needs that lock for Rust thread/TLS setup. Waiting can deadlock. Runtime preparation/install coordination belongs to the later backend PRs, not this frontend. Constructor, concurrent-first-use, and quiescent post-fork regressions are in #338 and passed on the cluster, with constructor schedules exercised on glibc 2.36/2.41. Fork during active initialization is not a newly supported contract.

agent/Dockerfile
```diff
@@ -109,6 +96,27 @@ RUN g++ -std=c++20 -O2 -Wall -Wextra -Werror \
...
+       make gcc binutils \
+    && rm -rf /var/lib/apt/lists/* \
+    && cargo install cbindgen --version 0.29.4 --locked \
+    && rustup component add rustfmt clippy
+WORKDIR /cuinterpose
+COPY cmd/cuinterpose/ ./
+RUN make frontend
+
+FROM cuinterpose-builder AS cuinterpose-test
+RUN make test-native
```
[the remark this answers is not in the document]
galletas1712: Agreed this is a CI coverage gap. CI expansion is explicitly deferred from the current frontend cleanup; no claim that root `make test` exercises this suite. The behavioral suites are placed in #338, with the frontend PR retaining its basic build/ABI checks. We ran the relevant builds and cluster regressions directly for this change, but that is not a substitute for future CI wiring.

agent/cmd/cuinterpose/frontend/tests/fixtures/runtime.c
```diff
@@ -0,0 +1,45 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "cuda.h"
```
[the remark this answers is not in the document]
galletas1712: This fixture has moved out of #326 into #338. We also checked the claimed build failure: in the actual glibc 2.36 builder's `dlfcn.h`, `RTLD_DEFAULT` is defined outside the `__USE_GNU` guard. The fixture compiles unchanged with `gcc -std=gnu11 -Wall -Wextra -Werror -fsyntax-only`, without defining `_GNU_SOURCE`; the full fixture builds and loader runs also passed. We are therefore not adding a feature macro solely for this finding on the supported toolchain.

agent/cmd/cuinterpose/frontend/frontend.c
```diff
@@ -0,0 +1,381 @@
...
+
+API CUresult cuGetProcAddress_v2_ptsz(const char *name, void **out, int version,
+                                     cuuint64_t flags,
+                                     CUdriverProcAddressQueryResult *status) {
+    if (!(flags & 3))
+        flags |= 2;
+    return cuGetProcAddress_v2(name, out, version, flags, status);
+}
+
+#define RUNTIME_QUERY(function_name, version_parameter, version_argument) \
```
eli-runai, line 366: wrong error type returned from cudaGetDriverEntryPoint* 

finish_query returns driver-API CUresult codes. RUNTIME_QUERY just casts that to int and hands it back as the return value of cudaGetDriverEntryPoint*, which is supposed to return runtime-API cudaError_t codes. Those are different numbering schemes. Concretely: when finish_query rejects a name with CUDA_ERROR_NOT_SUPPORTED (801), callers of cudaGetDriverEntryPoint get 801 back instead of the runtime code they'd expect, cudaErrorNotSupported (71).

agent/cmd/cuinterpose/frontend/frontend.c
```diff
@@ -0,0 +1,381 @@
...
+static void *lookup(void *handle, const char *name) {
+    if (pthread_once(&resolver_once, initialize_resolver) != 0 || !real_dlsym)
+        return NULL;
+    return real_dlsym(handle, name);
+}
+
+static int cuda_library_family(const char *path) {
+    const char *base = strrchr(path, '/');
+    base = base ? base + 1 : path;
+    const char *names[] = {"libcuda.so", "libcudart.so"};
```
eli-runai, line 47: strings like these can be static constexpr or #defines (all across this file)

agent/cmd/cuinterpose/frontend/frontend.c
```diff
@@ -0,0 +1,381 @@
...
+        // A failed private load must not hide another caller's published table.
+        api = atomic_load(&backend_api);
+        if (!api)
+            atomic_store(&backend_unavailable, true);
+    }
+    loading_backend = false;
+    return api;
+}
+
+#define MEMORY_API(X) \
```
eli-runai, line 217: perhaps you can move it to the begining of the file for better readability

agent/cmd/cuinterpose/frontend/frontend.c
```diff
@@ -0,0 +1,381 @@
...
+        return CUDA_ERROR_INVALID_VALUE;
+    // Procedure lookup may return the modern ABI for an unversioned name.
+    // Direct dlsym of an old 32-bit allocation ABI must remain untouched.
+    if (replacement(name) || strcmp(name, "cuMemAlloc") == 0 ||
+        strcmp(name, "cuMemFree") == 0 || strcmp(name, "cuMemGetAddressRange") == 0) {
+        Dl_info info;
+        void *wrapper = NULL;
+        if (*output && dladdr(*output, &info) && info.dli_sname && info.dli_saddr == *output) {
+            const char *actual = info.dli_sname;
+            bool same = strcmp(name, actual) == 0 ||
```
eli-runai, line 305: maybe use a seperate function that checks same?

agent/cmd/cuinterpose/frontend/frontend.c
```diff
@@ -0,0 +1,381 @@
...
+    void *wrapper = replacement(name);
+    return wrapper ? wrapper : address;
+}
+
+static CUresult finish_query(const char *name, void **output) {
+    if (!name || !output)
+        return CUDA_ERROR_INVALID_VALUE;
+    // Procedure lookup may return the modern ABI for an unversioned name.
+    // Direct dlsym of an old 32-bit allocation ABI must remain untouched.
+    if (replacement(name) || strcmp(name, "cuMemAlloc") == 0 ||
```
tzachi-runai, line 299:
```
if (!replacement(name) &&
      strcmp(name, "cuMemAlloc") != 0 &&
      strcmp(name, "cuMemFree") != 0 &&
      strcmp(name, "cuMemGetAddressRange") != 0) {
      return CUDA_SUCCESS;
  }

  Then the nested pointer validation can also become a guard:

  Dl_info info;
  if (!*output || !dladdr(*output, &info) ||
      !info.dli_sname || info.dli_saddr != *output) {
      *output = NULL;
      return CUDA_ERROR_NOT_SUPPORTED;
  }
```

agent/cmd/cuinterpose/frontend/frontend.c
tzachi-runai: Building on this, could we keep the allowed name pairs in one table? The helper could check exact matches or allowed pairs, and the outer condition could check whether the requested name appears in that table, alongside replacement(name). That would avoid maintaining the three unversioned names separately at lines 299–300 and 306–308.

agent/cmd/cuinterpose/frontend/frontend.c
```diff
@@ -0,0 +1,381 @@
...
+    ENTRY(cudaGetDriverEntryPoint)
+    ENTRY(cudaGetDriverEntryPoint_ptsz)
+    ENTRY(cudaGetDriverEntryPointByVersion)
+    ENTRY(cudaGetDriverEntryPointByVersion_ptsz)
+#undef ENTRY
+    return NULL;
+}
+
+API void *dlsym(void *handle, const char *name) {
+    void *address = lookup(handle, name);
```
tzachi-runai, line 283: Can we preserve the original caller’s lookup scope for RTLD_DEFAULT? If a plugin is loaded with RTLD_LOCAL, it can normally use dlsym(RTLD_DEFAULT, ...) to find symbols in its dependencies. Forwarding the lookup from this library changes that scope and can make an unrelated lookup return NULL. Could we add a regression test for this case?

agent/cmd/cuinterpose/frontend/frontend.c
```diff
@@ -0,0 +1,381 @@
...
+    struct CudaLibrary *next;
+};
+static _Atomic(struct CudaLibrary *) cuda_libraries;
+
+static void initialize_resolver(void) {
+    real_dlsym =
+        (void *(*)(void *, const char *))dlvsym(RTLD_NEXT, "dlsym", GLIBC_DLSYM_VERSION);
+}
+
+static void *lookup(void *handle, const char *name) {
```
tzachi-runai, line 38: This can deadlock during concurrent first use. One thread enters pthread_once and calls dlvsym, waiting for glibc’s loader lock. Another thread already holds that lock while running a library constructor, which calls our dlsym and waits for the same pthread_once to finish. Could we avoid this lock ordering and add a regression test for this case?

PR #327, opened 2026-09-16
feat(cuinterpose): define checkpoint and sharing protocol

galletas1712:
## Summary

Refs #295 (approved).

Define the versioned CUDA sharing protocol and persisted topology records. `BeginCheckpoint` enters exclusive checkpoint mode and returns stable records. Version 3 omits copy-timing fields from completion replies; transferred byte counts remain part of correctness validation.

- Preserve namespace PID and allocation identity, bounded MessagePack framing, and optional FD transfer.
- Version 3 also applies to virtual shareable and IPC memory handles. Older formats are rejected; the matching private C ABI is version 3.

Unix peer connection setup applies the socket send timeout before connecting, bounding waits on a full listen queue. The shared helper is used by peer exports and the coordinator.

## Validation

## Summary by CodeRabbit

* **New Features**
  * Added a versioned protocol for exchanging CUDA inspection and state information.
  * Added support for tracking allocations, mappings, multicast resources, devices, and binding relationships.
  * Added Unix-socket communication with optional file-descriptor transfer.
  * Added serialization and deserialization with size, format, version, nesting, and trailing-data validation.
  * Added virtual shareable-handle encoding and decoding with integrity checks.
  * Added configurable operation timeouts with safe default values.
  * Added stable identifiers for allocations and their creating namespaces.


agent/cmd/cuinterpose/rust/protocol/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+#[doc(inline)]
+pub use ticket::{Resource, ResourceKind, TICKET_MAGIC, Ticket};
+#[doc(inline)]
+pub use transport::{receive, send};
+
+pub const VERSION: u16 = 4;
+pub const MAX_RECORDS: usize = 4096;
+pub const MAX_ACCESS: usize = 32;
+// A maximal inspection (4096 mappings, 32 named access grants each) fits here.
+// The same bound applies to a state document; tickets have a smaller bound.
```
galletas1712: what does "state document" mean?

agent/cmd/cuinterpose/rust/protocol/src/record.rs
```diff
@@ -0,0 +1,94 @@
...
+    Address {
+        address: u64,
+        tracked_member: Option<MemberRange>,
+    },
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum BindingVersion {
+    V1,
```
galletas1712, line 29: is this version something CUDA-specific or something we made up?

agent/cmd/cuinterpose/rust/protocol/src/transport.rs
```diff
@@ -0,0 +1,101 @@
...
+};
+use serde::{Serialize, de::DeserializeOwned};
+use std::io::{self, IoSlice, IoSliceMut, Read, Write};
+use std::mem::MaybeUninit;
+use std::os::fd::{AsFd, OwnedFd};
+use std::os::unix::net::UnixStream;
+
+/// Sends one bounded message, optionally transferring a borrowed descriptor.
+///
+/// # Errors
```
galletas1712, line 17: this is weird comment style

agent/cmd/cuinterpose/rust/protocol/src/identity.rs
```diff
@@ -0,0 +1,42 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
+// SPDX-License-Identifier: Apache-2.0
+
```
galletas1712, line 3: this kind of just seems like boilerplate. Is there no built-in or serde type for just some bytes? or rather, does this even have to be bytes? 

I was thinking if would be simpler to just have ParticipantId and AllocationId be type aliases to some bytes/number etc.

agent/cmd/cuinterpose/rust/protocol/src/lib.rs
galletas1712: This means `cuinterpose.state`: the checkpoint file containing the participant IDs and allocation/mapping records. Changed the comment to name the file explicitly in 79119ea. The design doc and issue #295 now also show a decoded example; allocation bytes are in the CRIU-captured host carriers, not this file.

agent/cmd/cuinterpose/rust/protocol/src/record.rs
galletas1712: The enum is ours, but it selects the CUDA binding API variant to replay: V1 uses cuMulticastBindMem/cuMulticastBindAddr, and V2 uses their _v2 entry points with an explicit device argument. It is not the cuinterpose protocol version. Added that explanation directly above the enum in 79119ea.

agent/cmd/cuinterpose/rust/protocol/src/transport.rs
galletas1712: `///` and the `# Errors` heading are standard Rust API-documentation syntax; rustdoc renders this as the function’s error section. Kept that convention but simplified the wording in 79119ea: after a failed send, close the stream rather than retrying, because part of the message or its descriptor may already have been sent.

agent/cmd/cuinterpose/rust/protocol/src/identity.rs
galletas1712: Agreed about simplifying the handwritten conversion. In 79119ea, formatting and parsing now use standard u128 formatting/from_str_radix and big-endian byte conversion instead of per-byte loops. Kept the two small newtypes: aliases to the same byte array would let a participant ID be passed where an allocation ID is expected, and peer requests contain both. serde_bytes already handles binary serialization. No new dependency or wire-format change. The existing identity test in the final test PR now checks byte order, leading zeroes, unchanged binary encoding, and malformed input.

agent/cmd/cuinterpose/rust/protocol/src/lib.rs
```diff
@@ -0,0 +1,202 @@
...
+pub use transport::{receive, send};
+
+pub const VERSION: u16 = 5;
+pub const MAX_RECORDS: usize = 4096;
+pub const MAX_ACCESS: usize = 32;
+// A maximal inspection (4096 mappings, 32 named access grants each) fits here.
+// Also limits cuinterpose.state, which stores every participant's records.
+// Tickets have a separate, smaller limit.
+pub const MAX_BYTES: usize = 32 * 1024 * 1024;
+pub const MAX_TICKET_BYTES: usize = 4096;
```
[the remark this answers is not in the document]
galletas1712: Obsolete in the current protocol: tickets and MAX_TICKET_BYTES have been removed. Virtual shareable handles now use a fixed-size byte representation with dedicated encode/decode functions, so there is no ticket serialization path to limit.

agent/cmd/cuinterpose/rust/protocol/src/lib.rs
```diff
@@ -0,0 +1,176 @@
...
+/// Returns serialization failures or an error when the encoded size exceeds
+/// `MAX_MESSAGE_BYTES`.
+pub fn encode<T: Serialize>(body: &T) -> Result<Vec<u8>> {
+    let bytes = rmp_serde::to_vec_named(&Envelope {
+        version: VERSION,
+        body,
+    })?;
+    if bytes.len() > MAX_MESSAGE_BYTES {
+        return Err(Error::Invalid("message exceeds size limit"));
+    }
```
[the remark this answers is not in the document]
galletas1712: Leaving this unchanged for this stack. MAX_MESSAGE_BYTES remains the encoded-message/frame size limit; outgoing values are locally constructed metadata. A bounded serialization writer would be separate memory hardening, and we agreed to defer that rather than add more defensive machinery. The same encoder also serves #332 manifest writing, so any future change belongs here once, not in per-caller guards.

PR #329, opened 2026-09-16
feat(cuinterpose): track VMM ownership through checkpoint and restore

galletas1712:
## Summary

Refs #295 (approved).

Build on the process runtime in #360 to implement supported POSIX VMM allocations, virtual handles, mappings, and sharing history. Unsupported exportable creates and foreign imports fail before acquiring CUDA backing, so the registry no longer carries `raw` or `unsupported` state.

- `BeginCheckpoint` requires an active process with no unlocked driver calls and returns records while closing application memory mutation under the same mutex.
- Derive teardown and reconstruction from stable ownership and one process phase; remove allocation/mapping progress flags and pin counts.
- Keep ordinary pre-mutation CUDA errors returnable; terminate when a successful mutation cannot be recorded or destructive lifecycle work fails.
- Merge access permissions per location without an arbitrary entry limit. Keep checkpoint-entry behavior coverage and omit the duplicate transition-table test.
- IPC, host carriers, and multicast follow in #342, #330, and #331.

Process initialization, runtime publication, and state locking live in #360. This PR keeps the VMM implementation together: CUDA driver/ABI plumbing, the memory registry and virtual handles, allocation/mapping operations, and checkpoint lifecycle.

Memory entry points require a runtime established by successful intercepted `cuInit`; they do not initialize it themselves. Missing or inherited runtimes return `CUDA_ERROR_NOT_INITIALIZED` before touching bookkeeping locks. Cached CUDA export descriptors are explicitly close-on-exec.

Rust ABI entry points call handlers directly, preserving ordinary error returns and runtime-failure checks. Panics terminate through the workspace build policy rather than an explicit catch boundary.

Resolve and publish the complete driver symbol table during runtime preparation, outside both installation and state locks. Later IPC, carrier, and multicast calls share this cache; unavailable optional symbols fail on use. Successful range operations update/remove every tracked mapping in the range, including ranges starting in native memory. Failed driver calls leave records unchanged. Redundant retained references propagate release failures into the existing fail-stop path.

## Validation

Validated on the assembled stack at #338: `make -C agent/cmd/cuinterpose test`, `make check`, `make test`, and `make build`. The packaged suite includes formatting, strict Clippy, GNU/musl tests, C/Rust ABI checks, and headless regressions. Optional CUDA helper tests were skipped for unavailable development dependencies. Physical-GPU and cross-node tests were not run for this revision.

Intermediate PRs remain source layers, not independently buildable backends. Ship matching frontend/backend artifacts (private ABI version 3); the checkpoint wire/state format is version 3.

## Summary by CodeRabbit

- **New Features**
  - Added CUDA virtual-memory allocation, mapping, access-control, and sharing support.
  - Added import and export of shareable memory through POSIX file descriptors.
  - Added process checkpoint and restoration workflows for shared and mapped GPU memory.
  - Added lifecycle management for CUDA contexts and primary contexts.
  - Added validated initialization for CUDA interposition services.
- **Bug Fixes**
  - Improved handling of invalid CUDA handles, unsupported sharing formats, failed allocations, and unavailable driver symbols.
  - Added safer cleanup, recovery, and error reporting across CUDA operations.
  - Added validation for checkpoint ordering, memory mappings, allocation capacity, and file descriptors.


agent/cmd/cuinterpose/rust/core/src/memory/checkpoint.rs
```diff
@@ -0,0 +1,370 @@
...
+            .map_err(|_| CUDA_ERROR_OUT_OF_MEMORY)?;
+        for allocation in self.memblocks.values().filter_map(Memblock::unicast) {
+            let virtual_allocation_handle_count = self
+                .virtual_allocation_handles
+                .values()
+                .filter(|id| **id == allocation.reference.id)
+                .count()
+                .try_into()
+                .map_err(|_| CUDA_ERROR_OUT_OF_MEMORY)?;
+            let record = Record::Allocation {
```
galletas1712, line 80: should we add a "from allocation" helper somewhere? or rather, actually why do we need all these type conversions/owns_content etc?

agent/cmd/cuinterpose/rust/core/src/memory/checkpoint.rs
```diff
@@ -0,0 +1,370 @@
...
+                location: (
+                    allocation.properties.location.type_ as u32,
+                    allocation.properties.location.id,
+                ),
+                virtual_allocation_handle_count,
+            };
+            records.push(record);
+        }
+        for mapping in self.mappings.values() {
+            let record = Record::Mapping {
```
galletas1712, line 95: should we add a helper for this as well? or is this just unnecesary abstraction?

agent/cmd/cuinterpose/rust/core/src/memory/checkpoint.rs
```diff
@@ -0,0 +1,370 @@
...
+        assert_eq!(state.phase, Phase::Active);
+        state.unlocked_driver_calls = 0;
+        assert!(state.begin_checkpoint().unwrap().is_empty());
+        assert_eq!(state.phase, Phase::Checkpointing);
+        assert_eq!(state.begin_checkpoint(), Err(CUDA_ERROR_NOT_READY.into()));
+        assert_eq!(state.new_reference(), Err(CUDA_ERROR_NOT_READY.into()));
+    }
+
+    #[test]
+    fn lifecycle_transitions_require_each_local_milestone() {
```
galletas1712: is this test really necessary? I feel like this is just a restatement of the FSM definition

agent/cmd/cuinterpose/rust/core/src/memory/mod.rs
```diff
@@ -0,0 +1,254 @@
...
+                }
+                cache()?.remove(&id)?;
+            }
+        }
+        self.memblocks.remove(&id);
+        Ok(())
+    }
+}
+
+pub(crate) fn random<const N: usize>() -> Result<[u8; N]> {
```
galletas1712, line 203: can we use standard library for this instead?

agent/cmd/cuinterpose/rust/core/src/memory/mod.rs
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+    /// Tracked id, native `None`, or `INVALID_HANDLE` for a stale tagged handle.
+    pub(crate) fn tracked(&self, handle: u64) -> Result<Option<AllocationId>> {
+        VirtualAllocationHandle::from_raw(handle)
+            .map(|handle| handle.id(self))
+            .transpose()
+    }
+
+    pub(crate) fn settle(&mut self, id: AllocationId) -> Result<()> {
```
galletas1712: is there a more specific name for this? `settle` is quite vague

agent/cmd/cuinterpose/rust/core/src/memory/mod.rs
```diff
@@ -0,0 +1,254 @@
...
+        let handle = VirtualAllocationHandle(
+            VirtualAllocationHandle::TAG | self.next_virtual_allocation_handle,
+        );
+        self.next_virtual_allocation_handle += 1;
+        self.virtual_allocation_handles.insert(handle, id);
+        Ok(handle.as_raw())
+    }
+
+    /// Tracked id, native `None`, or `INVALID_HANDLE` for a stale tagged handle.
+    pub(crate) fn tracked(&self, handle: u64) -> Result<Option<AllocationId>> {
```
galletas1712: this is also vague. is this "to tracked", or "from tracked", etc? "tracked" is also vague.

agent/cmd/cuinterpose/rust/core/src/memory/sharing.rs
```diff
@@ -182,6 +184,75 @@ impl ExportCache {
     }
 }
 
+impl Memblock {
+    /// Publish the creator's export without making peer service acquire ProcessState.
+    pub fn share(&mut self, namespace_pid: NamespacePid) -> Result<AllocationReference> {
```
galletas1712: let's call this export maybe?

agent/cmd/cuinterpose/rust/core/src/memory/vmm.rs
```diff
@@ -0,0 +1,205 @@
...
+                access: Vec::new(),
+                flags,
+            },
+        );
+        Ok(())
+    }
+}
+
+impl Mapping {
+    pub(crate) fn merged_access(
```
galletas1712: why is this needed?

agent/cmd/cuinterpose/rust/core/src/memory/checkpoint.rs
galletas1712: Addressed with the simplification we discussed: kept this one-use record construction inline, renamed `owns_content` to `needs_content_checkpoint`, removed the redundant handle-type condition, and simplified the handle count conversion. The remaining conversions form the boundary between CUDA types and protocol fields. The content predicate still distinguishes shared device memory owned by this creator from imported/private allocations, so that ownership decision remains necessary.

agent/cmd/cuinterpose/rust/core/src/memory/checkpoint.rs
galletas1712: Kept this construction inline, as discussed. It has one caller and directly shows the wire fields; a helper would mostly move the same literal elsewhere without hiding substantial logic.

agent/cmd/cuinterpose/rust/core/src/memory/checkpoint.rs
galletas1712: Addressed: removed `lifecycle_transitions_require_each_local_milestone`, which duplicated the transition table. The checkpoint-entry behavior tests remain; the later test layer covers the packaged lifecycle.

agent/cmd/cuinterpose/rust/core/src/memory/mod.rs
galletas1712: Addressed by replacing the manual unsafe `getrandom` syscall/EINTR/partial-read loop with `getrandom::fill`. The standard-library random API is still nightly-only ([Rust documentation](https://doc.rust-lang.org/std/random/fn.random.html)). The getrandom version was already present transitively; #333 declares it as a direct dependency for this code.

agent/cmd/cuinterpose/rust/core/src/memory/mod.rs
galletas1712: Addressed: renamed `settle` to `release_unused_memblock`, with a comment describing when the driver handle is released and when the registry entry is removed.

agent/cmd/cuinterpose/rust/core/src/memory/mod.rs
galletas1712: Addressed: renamed `tracked` to `resolve_virtual_handle`. The documentation now spells out the result: an allocation ID for a live virtual handle, `None` for a native handle, and an invalid-handle error for a stale tagged handle.

agent/cmd/cuinterpose/rust/core/src/memory/sharing.rs
galletas1712: Addressed: renamed `Memblock::share` to `Memblock::export` and updated its callers.

agent/cmd/cuinterpose/rust/core/src/memory/vmm.rs
galletas1712: Kept the merge for the reason we discussed: access updates apply to the specified locations, so granting access to GPU 1 must preserve an earlier GPU 0 grant for checkpoint/restore. The proposed metadata is computed before the CUDA call and committed only if that call succeeds. Simplified the helper to return a Vec directly, removed the arbitrary 32-entry limit, and extended the existing access-update test beyond that limit.

agent/cmd/cuinterpose/rust/core/src/memory/checkpoint.rs
```diff
@@ -0,0 +1,302 @@
...
+                    }
+                    allocation.driver = Some(Context::run(
+                        allocation.context,
+                        allocation.properties.location.id,
+                        || crate::driver::import_posix(raw.as_fd()),
+                    )?);
+                }
+                self.remap(false)?;
+            }
+            _ => return Err(CudaError::from(CUDA_ERROR_NOT_SUPPORTED)),
```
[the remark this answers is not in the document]
galletas1712: Addressed by the subsequent source layers in this stack: #330 implements SaveAllocations and LoadAllocations, and #331 adds multicast lifecycle work. The current assembled backend implements the mandatory phases. As the PR description states, #329 is a source layer, not an independently buildable/shippable backend. Adding temporary no-op transitions or feature gating here would introduce behavior that is immediately removed later in the stack.

agent/cmd/cuinterpose/rust/core/src/memory/mod.rs
```diff
@@ -0,0 +1,245 @@
...
+        self.memblocks.remove(&id);
+        Ok(())
+    }
+}
+
+pub(crate) fn random<const N: usize>() -> Result<[u8; N]> {
+    let mut bytes = [0; N];
+    getrandom::fill(&mut bytes).map_err(|_| CUDA_ERROR_NOT_INITIALIZED)?;
+    Ok(bytes)
+}
```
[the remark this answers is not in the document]
galletas1712: The core manifest and workspace/build integration intentionally arrive in #333, which declares getrandom.workspace = true for cuinterpose-core and includes core in the workspace. This PR is a source layer, as documented in its description; the assembled stack already has the dependency. No additional manifest or dependency declaration is needed in #329.

agent/cmd/cuinterpose/rust/core/src/driver.rs
```diff
@@ -0,0 +1,172 @@
...
+            $(
+                pub fn $name() -> Result<unsafe extern "C" fn($($ty),*) -> CUresult> {
+                    let address = crate::driver(
+                        std::ffi::CStr::from_bytes_with_nul(
+                            concat!(stringify!($name), "\0").as_bytes()).expect("static CUDA symbol"));
+                    if address.is_null() {
+                        return Err(CudaError::from(CUDA_ERROR_NOT_INITIALIZED));
+                    }
+                    // The frontend resolves this exact NVIDIA driver signature.
+                    Ok(unsafe { std::mem::transmute::<*mut c_void, unsafe extern "C" fn($($ty),*) -> CUresult>(address) })
```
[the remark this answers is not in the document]
galletas1712: Fixed in #329 at the shared driver boundary. Runtime preparation resolves a private typed symbol table before acquiring installation/state locks, then publishes it without executing the resolver inside a OnceLock initializer. Driver calls use cached pointers; missing optional symbols fail on use. This also covers #342 IPC, #330 carrier/context cleanup, and #331 multicast. #338 verifies that no resolver callback occurs after runtime publication and that an absent optional symbol does not prevent initialization. Existing constructor/race and reciprocal-import regressions pass.

agent/cmd/cuinterpose/rust/core/src/handlers.rs
```diff
@@ -0,0 +1,222 @@
...
+
+    state.map_unicast(id, address, size, offset, flags)
+}
+
+pub fn cuMemUnmap(address: u64, size: usize) -> Result<()> {
+    let mut state = active()?;
+    unsafe { crate::driver::cuMemUnmap(address, size) }?;
+    if let Some(mapping) = state.mappings.remove(&address) {
+        runtime::must_complete(state.release_unused_memblock(mapping.id));
+    }
```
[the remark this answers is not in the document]
galletas1712: Fixed the multi-mapping range handling in #329. After successful unmap, every tracked mapping in the range is removed and its backing is released as appropriate. Access updates prepare metadata for all tracked mappings in the range and commit only after CUDA succeeds. This also covers multicast through the shared handlers; #342 malloc/IPC retains its whole-allocation lookups. #338 tests a range starting in native memory and spanning two tracked allocations, plus unchanged metadata on failed access updates. We did not add partial-unmap support or extra admission guards: CUDA requires whole mapped ranges for unmap (https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__VA.html).

agent/cmd/cuinterpose/rust/core/src/memory/vmm.rs
```diff
@@ -0,0 +1,199 @@
...
+        let allocation = self
+            .memblocks
+            .get_mut(&id)
+            .and_then(Memblock::unicast_mut)
+            .ok_or(CUDA_ERROR_INVALID_HANDLE)?;
+        if allocation.driver.is_some() {
+            let _ = unsafe { crate::driver::cuMemRelease(driver) };
+        } else {
+            allocation.driver = Some(driver);
+        }
```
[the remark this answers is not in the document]
galletas1712: Fixed in #329 by propagating the redundant cuMemRelease result with ?. The existing caller already wraps retain_backing in must_complete, so a failed balancing release terminates through the established policy. #338 injects that failure and checks termination. I also checked #330 cleanup paths: secondary cleanup errors that preserve an existing fatal operation error are intentionally unchanged.

agent/cmd/cuinterpose/rust/core/src/handlers.rs
```diff
@@ -0,0 +1,235 @@
...
+    let descriptors = unsafe { std::slice::from_raw_parts(access, count) };
+    // Access applies to a fully mapped range, potentially spanning allocations.
+    // Prepare metadata before CUDA and publish it only after the call succeeds.
+    let updates: Vec<_> = state
+        .mappings
+        .range(address..)
+        .take_while(|(start, _)| **start - address < size as u64)
+        .map(|(start, mapping)| (*start, mapping.merged_access(descriptors)))
+        .collect();
```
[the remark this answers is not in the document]
galletas1712: The driver rejects access ranges that begin or end partway through a mapping; successful ranges can span multiple whole mappings. The existing whole-mapping metadata is sufficient, so no subrange tracking or duplicate application validation was added. The fake driver and regression coverage in [edf48c7f](https://github.com/ai-dynamo/snapshot/commit/edf48c7f66460ab6e1f94f0a8ff6062f201805be) now reject partial access ranges and overlapping mappings and verify that failed calls leave records unchanged. The packaged cuinterpose suite passed, including strict Clippy, GNU/musl tests, ABI checks, and headless regressions. Physical-GPU tests were not run.

agent/cmd/cuinterpose/rust/core/src/driver.rs
```diff
@@ -0,0 +1,195 @@
...
+    cuMemRetainAllocationHandle(handle: *mut CUmemGenericAllocationHandle, address: *mut c_void);
+    cuMemSetAccess(address: CUdeviceptr, size: usize, access: *const CUmemAccessDesc, count: usize);
+    cuMemUnmap(address: CUdeviceptr, size: usize);
+}
+
+pub(super) fn context() -> usize {
+    let mut context = std::ptr::null_mut::<c_void>();
+    if unsafe { crate::driver::cuCtxGetCurrent(&mut context) }.is_err() {
+        return 0;
+    }
```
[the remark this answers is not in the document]
galletas1712: Fixed in [fcd422cf](https://github.com/ai-dynamo/snapshot/commit/fcd422cf5cb6cc020de8f554a48840e734b100b4) and propagated through IPC and multicast. `context()` returns `Result<usize>`; callers capture fallible context metadata before creating/importing/mapping driver state or publishing an export, and propagate the original CUDA error. Zero now means a successful query found no current context. Regressions in [edf48c7f](https://github.com/ai-dynamo/snapshot/commit/edf48c7f66460ab6e1f94f0a8ff6062f201805be) inject context-query failures into creation, mapping, export, and IPC setup and check unchanged outputs and bookkeeping; the genuine no-current-context path remains covered. The packaged cuinterpose suite passed, including strict Clippy, GNU/musl tests, ABI checks, and headless regressions. Physical-GPU tests were not run.

PR #331, opened 2026-09-16
feat(cuinterpose): multicast tracking and reconstruction

galletas1712:
## Summary

Refs #295 (approved).

Track and reconstruct CUDA multicast objects in the common memblock and virtual-handle registry. Application calls release the process mutex around blocking driver work and use one process-wide unlocked-call counter; per-object pins and busy counters are removed.

- Require POSIX-FD multicast objects and reject unsupported creation before CUDA allocation. The application owns synchronization of object destruction.
- During restore, the application remains parked and peer FD service uses its separate cache lock. Reconstruction therefore holds the process mutex and updates records directly, without clone/unlock/merge bookkeeping or per-object progress flags.
- Preserve creator, importer, device, and binding barriers. Failed reconstruction terminates; no rollback or partial replay.

The CUDA binding entry points remain thin adapters over shared multicast validation and replay logic. Property-only granularity queries are not intercepted.

## Validation

## Summary by CodeRabbit

- **New Features**
  - Added CUDA multicast memory support, including object creation, device registration, address and memory binding, mapping, unbinding, and granularity queries.
  - Added support for sharing and importing multicast allocations.
  - Added checkpoint and restore support for multicast memory objects.
  - Improved validation and error reporting for unsupported inputs, invalid handles, ranges, and bindings.


agent/cmd/cuinterpose/rust/core/src/handlers.rs
```diff
@@ -234,6 +252,137 @@ pub fn cuMemGetAllocationPropertiesFromHandle(
...
+        offset,
+        size,
+        flags,
+        device,
+        BindingVersion::V2,
+        BindInput::Address(address),
+    )
+}
+
+pub fn cuMulticastGetGranularity(
```
galletas1712: I don't think we need to intercept this function then?

agent/cmd/cuinterpose/rust/core/src/handlers.rs
```diff
@@ -234,6 +252,137 @@ pub fn cuMemGetAllocationPropertiesFromHandle(
...
+    size: usize,
+    flags: u64,
+) -> Result<()> {
+    multicast::bind(
+        handle,
+        offset,
+        size,
+        flags,
+        0,
+        BindingVersion::V1,
```
galletas1712, line 341: do you think all the binding version specific stuff should belong here in handlers, or will that just bloat the code with repeated code?

agent/cmd/cuinterpose/rust/core/src/handlers.rs
galletas1712: Removed in [5c6f472a](https://github.com/ai-dynamo/snapshot/commit/5c6f472a2f8bda012bc00c2b15184a78c8559043), with the frontend/private-ABI changes in [0465668e](https://github.com/ai-dynamo/snapshot/commit/0465668e735516cb1226685b0e00814c9e64744d). The audit found another unnecessary interceptor: `cuMemGetAllocationGranularity`. Both property-only queries now go directly to CUDA. The backend retains its internal allocation-granularity lookup for IPC allocation sizing. The remaining intercepted APIs translate virtual handles, track state, or support initialization and symbol lookup.

Removed both callbacks and exports and advanced the private ABI to version 3. #338 covers direct calls, dlsym, driver proc-address queries, and runtime entry-point queries, including missing-backend cases. The packaged tests passed.

agent/cmd/cuinterpose/rust/core/src/handlers.rs
galletas1712: Kept the four thin CUDA entry-point adapters in `handlers.rs`. They only supply the binding source, version, and device to `multicast::bind`; shared validation and bookkeeping stay in `memory/multicast.rs`, and `Binding::apply` selects the driver entry point during replay. Moving that logic into the handlers would duplicate it. No additional abstraction or duplicated version-specific implementation was added. The packaged binding and multicast regressions passed.

agent/cmd/cuinterpose/rust/core/src/handlers.rs
```diff
@@ -234,6 +252,137 @@ pub fn cuMemGetAllocationPropertiesFromHandle(
...
+    if properties.handleTypes
+        != u64::from(CUmemAllocationHandleType::CU_MEM_HANDLE_TYPE_POSIX_FILE_DESCRIPTOR.0)
+    {
+        return Err(CUDA_ERROR_NOT_SUPPORTED.into());
+    }
+    let state = runtime::active()?;
+    let id = memory::random()?;
+    let (mut state, driver) = multicast::create_backing(state, &properties, out)?;
+    let driver = runtime::must_complete(VirtualAllocationHandle::from_driver(driver));
+    let handle = runtime::must_complete(state.adopt_multicast(id, driver, properties));
```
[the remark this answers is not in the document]
galletas1712: Removed the scattered virtual-handle capacity prechecks in [fcd422cf](https://github.com/ai-dynamo/snapshot/commit/fcd422cf5cb6cc020de8f554a48840e734b100b4) instead of adding another preflight check or reservation mechanism to multicast creation. Handle generation retains one check protecting the shim's tagged handle namespace; it is an internal invariant that CUDA cannot enforce. Application argument validation remains with CUDA. The packaged cuinterpose suite passed, including strict Clippy, GNU/musl tests, ABI checks, and headless regressions. Physical-GPU tests were not run.

agent/cmd/cuinterpose/rust/core/src/handlers.rs
```diff
@@ -74,8 +76,17 @@ pub fn cuMemRetainAllocationHandle(out: *mut u64, address: *mut c_void) -> Resul
...
-    if id.is_some() {
+    if let Some(id) = id {
         state.check_handle_capacity()?;
+        if let Some(object) = state.memblocks.get(&id).and_then(Memblock::multicast) {
+            if object.driver.is_none() {
+                return Err(CudaError::from(CUDA_ERROR_INVALID_HANDLE));
+            }
+            unsafe {
+                out.write(state.mint_virtual_allocation_handle(id)?);
+            }
```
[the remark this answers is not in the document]
galletas1712: Fixed for both unicast in [fcd422cf](https://github.com/ai-dynamo/snapshot/commit/fcd422cf5cb6cc020de8f554a48840e734b100b4) and multicast in [53212247](https://github.com/ai-dynamo/snapshot/commit/532122473fe4c93dac91f6497d2ea759e142faec). Each mapping keeps the application handle used to create it. Retain returns that value and increments its application reference count; release removes the handle entry only after the last application reference. A mapping can recover the same value after the original reference has been released. Regressions in [edf48c7f](https://github.com/ai-dynamo/snapshot/commit/edf48c7f66460ab6e1f94f0a8ff6062f201805be) cover repeated retains, balanced releases, interior addresses, released handles, and multicast/unicast handle identity across restore. The packaged cuinterpose suite passed, including strict Clippy, GNU/musl tests, ABI checks, and headless regressions. Physical-GPU tests were not run.

PR #332, opened 2026-09-16
feat(cuinterpose): coordinate checkpoint and restore across processes

galletas1712:
## Summary

Refs #295 (approved).

Coordinate capture and restore across the fixed set of interposed CUDA processes. Capture begins with `BeginCheckpoint` on each participant, validates the returned stable topology, and only then starts destructive phases.

- Inspection contains records only; unsupported-resource counts and deferred admission policy are removed.
- Dispatch each phase concurrently and wait for every participant before advancing. Preserve all allocation and multicast dependency barriers.
- Never retry a mutating request after a failure or lost reply. Publish checkpoint state only after successful preparation and compare restored records before completion.

The coordinator has no phase-report or throughput output. Transfer byte counts still validate completion, and topology failures identify the participant and allocation or conflicting properties.

## Validation

Validated on the assembled stack at #338:

- `make -C agent/cmd/cuinterpose test`: formatting, strict Clippy, GNU/musl builds, C/Rust ABI checks, Rust tests, and packaged fake-driver regressions.
- `make check`, `make test`, and `make build`. The optional CUDA helper tests were skipped because their C++20/OpenSSL development dependencies are unavailable.
- Physical-GPU and cross-node tests were not rerun for this revision. Earlier GPU results do not qualify these changes.

Intermediate PRs remain source layers; this does not claim independent buildability of every layer. Ship matching artifacts and recreate checkpoints made with an older format.

## Summary by CodeRabbit

- **New Features**
  - Added a CUDA checkpoint and restore coordinator for synchronized operations across participating processes.
  - Added checkpoint manifest persistence with atomic writes and secure file handling.
  - Added validation for allocation, multicast, device, and namespace configuration.
  - Restore operations now verify checkpoint participants and confirm that restored topology matches the saved state.

- **Bug Fixes**
  - Improved handling of invalid responses, transfer sizes, missing references, operation failures, and unexpected file descriptors.
  - Failures and unexpected worker errors now result in clear command-line failure reporting.

PR #333, opened 2026-09-16
build(cuinterpose): package frontend, Rust backend, and coordinator

galletas1712:
## Summary

Refs #295 (approved).

Build and package the GNU/glibc preload libraries and static-musl coordinator, consolidate Cargo integration, and include source/license information. The standalone artifact-verification script is removed.

## Review boundary

The shim build has no PageBroker schema or protobuf dependency. This retains the cleaned-up stack's existing build layout rather than copying the experiment branch's build changes.

## Stack and compatibility

| Order | PR | Scope |
| --- | --- | --- |

## Validation


agent/cmd/cuinterpose/verify_artifacts.py
```diff
@@ -0,0 +1,50 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
```
galletas1712: do we really need this file?

galletas1712: Deleted `verify_artifacts.py` entirely in [b40cc8a5](https://github.com/ai-dynamo/snapshot/commit/b40cc8a5f21dd05f083467116186cb779a286758), including its Makefile invocation. Updated the documentation and PR descriptions to remove claims about that verifier. The normal builds and existing ABI/behavioral tests remain. `make check`, `make test`, `make build`, and the packaged cuinterpose suite passed.

PR #334, opened 2026-09-16
feat(podcontract): deliver CUDA tools and enable opt-in preload

galletas1712:
## Summary

Refs #295 (approved).

Deliver CUDA tools at fixed paths, reserve the per-container control directory, and enable the shim through the cuinterpose annotation.

## Review boundary

Do not select a content backend or decide whether to wrap a command based on GPU count. Existing shared jobfile declarations survive only until their remaining callers are removed in the following slices.

Use `slices.EqualFunc` for the tool-mount comparison while preserving the exact fields compared.

## Stack and compatibility

| Order | PR | Scope |
| --- | --- | --- |

## Validation


api/podcontract/cuinterpose.go
```diff
@@ -0,0 +1,166 @@
...
+			Capabilities:             &corev1.Capabilities{Drop: []corev1.Capability{"ALL"}},
+			SeccompProfile:           &corev1.SeccompProfile{Type: corev1.SeccompProfileTypeRuntimeDefault},
+		},
+	})
+	*podTemplate = *shaped
+	return nil
+}
+
+// Reject adoption of a Job that would run without the requested shim.
+func VerifyCuinterposeCapture(spec *corev1.PodSpec, targetContainers []string) error {
```
galletas1712: do we really need to verify this?

api/podcontract/cuinterpose.go
```diff
@@ -0,0 +1,166 @@
...
+	for _, field := range fields {
+		if field != CuinterposeLibraryPath {
+			filtered = append(filtered, field)
+		}
+	}
+	env.Value = strings.Join(filtered, " ")
+	return nil
+}
+
+func preloadFields(value string) []string {
```
galletas1712: avoid small, generic helpers

api/podcontract/cuinterpose.go
```diff
@@ -0,0 +1,166 @@
...
+	return strings.FieldsFunc(value, func(r rune) bool {
+		return r == ':' || unicode.IsSpace(r)
+	})
+}
+
+func envValue(env []corev1.EnvVar, name string) string {
```
galletas1712: avoid small, generic helpers

api/podcontract/protocol.go
```diff
@@ -26,6 +26,10 @@ const (
 	// SkipCompatCheckAnnotation disables compatibility checks for one restore.
 	SkipCompatCheckAnnotation = "nvidia.com/snapshot-skip-compat-check"
 
+	// CuinterposeAnnotation enables the CUDA interposer for checkpoint targets.
+	CuinterposeAnnotation        = "nvidia.com/cuinterpose"
+	CuinterposeAnnotationEnabled = "enabled"
```
galletas1712: do we really need this? how are other boolean annotations handled? use that style. I don't think we need to introduce another constant here.

api/podcontract/cuinterpose.go
galletas1712: Removed VerifyCuinterposeCapture, its adoption check, and the verification tests. Existing ownership/UID checks remain; opted-in capture lets the coordinator report missing shim participants.

api/podcontract/cuinterpose.go
galletas1712: Inlined the colon/whitespace splitting into setCuinterposePreload and removed preloadFields. It had only one production caller after removing verification.

api/podcontract/cuinterpose.go
galletas1712: Removed envValue along with verification. Tests now inspect the generated container environment directly.

api/podcontract/protocol.go
galletas1712: Changed this to match SkipCompatCheckFromAnnotations: trim whitespace, use strconv.ParseBool, and treat absent or invalid values as false. Removed the value constant and parsing-error plumbing. The annotation is now nvidia.com/cuinterpose-enabled: "true", and the examples use that name. The earlier draft "enabled" value no longer enables it.

galletas1712:
Superseded by #336, which now includes the pod contract, agent checkpoint/restore integration, and operator/Helm wiring directly on #333. Review discussions remain here; continue reviewing the combined change in https://github.com/ai-dynamo/snapshot/pull/336.

PR #335, opened 2026-09-16
feat(agent): integrate cuinterpose into checkpoint and restore

galletas1712:
## Summary

Refs #295 (approved).

Run the cuinterpose coordinator in the workload namespaces before native CUDA capture and after native CUDA restore, while application threads remain parked.

- Require checkpoint artifact format 3 to match the exclusive-entry protocol and reject older prepared artifacts before CRIU.
- Remove obsolete raw-import and unsupported-creation progress fields.
- Preserve the source-termination contract after failed capture and publish restore completion only after all reconstruction succeeds.

Coordinator execution returns success or an error with stderr; phase-report parsing and logging are removed.

## Validation

Validated on the assembled stack at #338:

- `make -C agent/cmd/cuinterpose test`: formatting, strict Clippy, GNU/musl builds, C/Rust ABI checks, Rust tests, and packaged fake-driver regressions.
- `make check`, `make test`, and `make build`. The optional CUDA helper tests were skipped because their C++20/OpenSSL development dependencies are unavailable.
- Physical-GPU and cross-node tests were not rerun for this revision. Earlier GPU results do not qualify these changes.

Intermediate PRs remain source layers; this does not claim independent buildability of every layer. Ship matching artifacts and recreate checkpoints made with an older format.

PR #336, opened 2026-09-16
feat(operator): configure CUDA tool delivery and opt-in preload

galletas1712:
## Summary

Refs #295 (approved).

Configure tool delivery and opt-in preload from the operator and Helm chart. Validate the expected tool/control mounts when adopting source workloads.

## Review boundary

Remove launch-job command rewriting, the wrapper CLI flag, and DRA claim sizing/RBAC used only to choose wrapping. Keep generic DRA GPU discovery and device mapping. Finish removal of unused shared jobfile declarations after their last callers are gone.

## Stack and compatibility

| Order | PR | Scope |
| --- | --- | --- |

## Validation


operator/internal/controller/snapshotjob_job.go
```diff
@@ -15,17 +15,10 @@ import (
...
-// contract (control volume, readiness probe, labels, seccomp, sidecar opt-outs), not
-// Dynamo-specific code — and adds only the owner label so the PodSnapshot created
-// later (PR 4) can be mapped back to this SnapshotJob without an ownerReference.
-//
-// No storage is injected (spec §5.3: the agent falls back to its own config), and
-// WrapLaunchJob is always false (spec §5.4: PodSnapshotTemplate has no field to
-// source it from — a caller needing cuda-checkpoint --launch-job wrapping sets it
-// up themselves in spec.podTemplate).
-func buildSourceJob(sj *snapshotv1alpha1.SnapshotJob) (*batchv1.Job, error) {
+func buildShapedSourceJob(
```
galletas1712: why did you rename this function?

galletas1712 (commented):
in general, it seems like we are doing some refactoring here. Is this really necessary?

operator/internal/controller/snapshotjob_job.go
galletas1712: The rename was unnecessary. Restored buildSourceJob with its original signature and removed the test wrapper. Cuinterpose is applied once in the reconciler, immediately before creating the Job. classifyExistingSourceJob is also back to its original standalone form.

galletas1712:
Addressed [the refactoring feedback](https://github.com/ai-dynamo/snapshot/pull/336#pullrequestreview-5264160098): restored the original Job builder and adoption functions, removed the test wrapper and shim-specific adoption verification, and removed unused PodSnapshot annotation propagation. The reconciler now has one shim-injection call immediately before Job creation. #336 is down to 67 added lines across 6 files, including its opt-in creation test. `make check`, `make test`, and `make build` pass on the assembled stack; optional CUDA helper C++ tests skip locally.

charts/snapshot/templates/_helpers.tpl
```diff
@@ -71,6 +71,14 @@ Name of the operator service account.
 {{- printf "%s-operator" (include "snapshot.fullname" .) }}
 {{- end }}
 
+{{/*
```
galletas1712: why do we need these changes in templates/ at all?

operator/cmd/manager/main.go
```diff
@@ -24,6 +26,16 @@ func main() {
 	ctrl.SetLogger(zap.New(zap.UseDevMode(true)))
 
 	artifactCleanupConfig := bindArtifactCleanupFlags(flag.CommandLine)
+	agentImage := flag.String(
```
galletas1712, line 29: I thought for the init container delivery we would just use the standard busybox image? basically it's just a lightweight dependency injector, no?

agent/cmd/ns-bind-mount/main.c
```diff
@@ -2,15 +2,17 @@
  * SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  *
- * ns-bind-mount installs and removes the two mounts used by restore:
+ * ns-bind-mount installs and removes the fixed mounts used by restore:
  *
  *   mount-bundle-fd <namespace-fd>
+ *   mount-snapshot-cuda-fd <namespace-fd>
```
galletas1712, line 8: not sure why you need this new mount-snapshot-cuda-fd and unmount-snapshot-cuda-fd. What are we actually mounting/unmounting? is it the libcuinterpose.so and libcuinterpose-core.so, or is it cuda-checkpoint? Because if it's the latter, it shouldn't go in this PR.

charts/snapshot/templates/_helpers.tpl
galletas1712: Removed the shared image helper and restored the original DaemonSet template. The remaining template change passes the existing configured agent image and pull policy to the operator so it can construct the shim installer, including when Helm overrides the repository or tag. All 14 chart tests pass.

operator/cmd/manager/main.go
galletas1712: The init container only runs cp, but the source files live in the agent image: /usr/local/lib/snapshot/libcuinterpose.so and libcuinterpose_core.so. Stock BusyBox has cp but neither library. Keeping the agent image supplies those files without adding a download or host-path delivery mechanism. The installer uses the existing chart image settings and runs only for opted-in workloads; I removed the unrelated Helm helper/DaemonSet refactor.

agent/cmd/ns-bind-mount/main.c
galletas1712: This mounts /snapshot-binaries/snapshot-cuda read-only at /tmp/snapshot-cuda inside the restore target. That directory contains only libcuinterpose.so and libcuinterpose_core.so. CRIU needs those files at their capture-time paths, including when the placeholder has no shim init container. The paired unmount removes that bind mount during cleanup. These new commands do not mount cuda-checkpoint; its existing binary-bundle path is unchanged.

PR #354, opened 2026-09-17
fix(agent): skip unchanged restored condition writes

### LINKED ISSUE #353: Avoid no-op restored condition status writes

Skip the Pod status apply when the existing `nvidia.com/Restored` condition already has the requested status, reason, and message. Real condition transitions should continue through server-side apply and retain the existing `LastTransitionTime` behavior.

### Why is this needed?

Restore dependencies can remain pending across several reconciliations. Reapplying an identical condition creates a new Pod resource version and another informer event without changing user-visible state. Avoiding the no-op write reduces API server traffic and self-induced reconciliation.

### Acceptance criteria

- Identical restored conditions do not issue an API write.
- Changes to status, reason, or message are persisted normally.
- Existing transition-time behavior is preserved.
- Unit tests cover both paths.

agent

### Before submitting

hutm:
## Problem

Repeated restore reconciliations apply the same `nvidia.com/Restored` condition even when its status, reason, and message have not changed. Each no-op write creates a new Pod resource version and triggers another informer event.

## Why

Restore dependencies can remain pending across several reconciliations. Avoiding identical status writes reduces API server traffic and self-induced reconciliation, especially when many restores run concurrently.

## Change

Skip the status apply when the existing condition already matches. Real transitions continue through server-side apply and retain the existing `LastTransitionTime` behavior.

## Testing

- Added coverage for skipping an identical condition and persisting a changed condition.

Fixes https://github.com/ai-dynamo/snapshot/issues/353


julienmancuso:
This looks good to me. The change is small and directly addresses the redundant status writes without altering transition-time behavior. The tests cover both the skipped write and a real condition change.

I’d be comfortable marking this ready once issue #353 receives the required `approved` label.

PR #355, opened 2026-09-17
perf(agent): publish normal events off restore path

### LINKED ISSUE #351: Keep Normal Event writes off the restore critical path

Publish Normal Kubernetes Events outside the checkpoint and restore reconciliation path through a bounded worker queue. Warning events should remain synchronous, and queue saturation should be observable.

### Why is this needed?

Every Event API write currently blocks reconciliation. A slow Event API request can delay the workload even though Normal events are best-effort telemetry. Informational event delivery should not be part of the restore critical path.

### Acceptance criteria

- Normal event publication does not wait for the API request.
- Warning event publication remains synchronous.
- The queue is bounded and handles saturation explicitly.
- Concurrency behavior has race-test coverage.

agent

### Before submitting

hutm:
## Problem

Every Kubernetes Event write currently blocks checkpoint or restore reconciliation. A slow Event API request can delay the workload even though Normal events are best-effort telemetry.

## Why

Informational event delivery should not be part of the restore critical path. Warning events still need synchronous error reporting.

## Change

Publish Normal events through a bounded worker queue. Keep Warning events synchronous, copy queued Pod data, and drop a Normal event with a diagnostic when the queue is full.

This removes a blocking dependency from the critical path, but local end-to-end measurements did not show a statistically significant latency improvement.

## Testing

- Verified Normal event publication returns while the API request is blocked.
- Verified Warning event publication waits for the API request.
- Ran the complete agent test suite and event publisher race tests.

Fixes https://github.com/ai-dynamo/snapshot/issues/351


agent/internal/controller/event_publisher.go
```diff
@@ -0,0 +1,94 @@
...
+		log:       log,
+		component: component,
+		eventType: eventType,
+		reason:    reason,
+		message:   message,
+	}
+	select {
+	case p.pending <- event:
+	default:
+		p.log.V(1).Info("Normal Kubernetes Event queue is full; dropping best-effort event",
```
julienmancuso: The overall asynchronous approach looks sound, including keeping Warning events synchronous and copying the Pod before queueing it. One issue before this is ready: queue saturation is logged with `V(1)`, which is hidden under Snapshot’s default `info` configuration. That means dropped events would normally be invisible despite the acceptance criterion that saturation be observable. Could this use a rate-limited info-level signal or a metric? I’d also add a deterministic test that fills the queue and verifies that publication remains non-blocking while exercising this drop path.

PR #356, opened 2026-09-17
feat(agent): log restore controller phase timings

### LINKED ISSUE #350: Add restore controller phase timing

Log one structured timing summary per restore-controller pass with total, preflight, finalizer, and restore durations, plus the stage where the pass returned.

### Why is this needed?

Current timing starts inside the executor, leaving controller work such as preflight validation and finalizer updates unaccounted for. Without that attribution, operators cannot tell whether a slow restore spent time preparing the request or executing CRIU and CUDA restore.

### Acceptance criteria

- Each controller pass emits one structured timing summary.
- Preflight, finalizer, restore, total duration, and return stage are represented.
- Unit tests cover the structured fields.

agent

### Before submitting

hutm:
## Problem

Restore timing starts inside the executor, leaving controller work such as preflight validation and finalizer updates unaccounted for.

## Why

Without controller-stage timing, operators cannot tell whether a slow restore spent time preparing the request or executing CRIU and CUDA restore.

## Change

Log one structured timing summary per controller pass with total, preflight, finalizer, and restore durations, plus the stage where the pass returned.

## Testing

- Added coverage for the structured timing fields on a preflight failure.

Fixes https://github.com/ai-dynamo/snapshot/issues/350


agent/internal/controller/controller.go
```diff
@@ -484,18 +484,44 @@ func (w *NodeController) handleTerminalRestorePod(ctx context.Context, pod *core
 	return w.removeRestoreFinalizerWithEvent(ctx, pod)
 }
 
-func (w *NodeController) reconcileRestorePod(ctx context.Context, pod *corev1.Pod) bool {
+func (w *NodeController) reconcileRestorePod(ctx context.Context, pod *corev1.Pod) (requeue bool) {
 	podKey := fmt.Sprintf("%s/%s", pod.Namespace, pod.Name)
+	passStarted := time.Now()
```
julienmancuso: The structured timing is useful, but this starts after several controller-pass return paths. `processRestoreQueueItem` can return earlier for missing, already-handled, terminal, deleting, or inactive Pods, so those passes do not emit the promised summary. Could we either move the pass-level timing wrapper outward into `processRestoreQueueItem`, or narrow the PR and issue wording to say this covers each active restore attempt rather than every controller pass?

PR #357, opened 2026-09-17
perf(agent): reuse observed pod for non-DRA restores

### LINKED ISSUE #352: Reuse observed Pod metadata on non-DRA restores

Allow restore GPU discovery to reuse the Pod already observed by the node controller when the Pod identity matches and its immutable spec has no resource claims. Pods that may use DRA must retain the live Pod lookup because generated claim names arrive through status.

### Why is this needed?

The restore controller already holds the scheduled Pod, but GPU discovery performs another Pod GET before deciding whether DRA is relevant. For Pods without resource claims, that request is redundant and adds API traffic to the restore path. Actual GPU assignment should continue to come from kubelet PodResources.

### Acceptance criteria

- A matching Pod without resource claims performs no additional Pod GET.
- DRA-capable Pods retain the live API lookup.
- Pod identity is checked before cached metadata is used.
- Unit tests cover the optimized path.

agent

### Before submitting

hutm:
## Problem

The restore controller already has the scheduled Pod from its informer, but GPU discovery performs another Pod GET before deciding whether DRA is relevant.

## Why

For Pods without resource claims, the immutable Pod spec already proves that DRA lookup is unnecessary. Removing the redundant GET reduces API traffic and avoids an extra request on the restore path.

## Change

Reuse the observed Pod when its identity matches and it has no resource claims. Pods that may use DRA retain the live Pod lookup because generated claim names arrive through Pod status. GPU assignment still comes from kubelet PodResources.

Local phase measurements showed a smaller GPU-discovery preflight, but no statistically significant end-to-end latency improvement.

## Testing

- Added coverage verifying that a matching Pod without resource claims performs no Pod GET.

Fixes https://github.com/ai-dynamo/snapshot/issues/352


agent/internal/cuda/cuda.go
```diff
@@ -187,6 +188,31 @@ func DiscoverGPUs(ctx context.Context, clientset kubernetes.Interface, podName,
...
+		DiscoverVisibleGPUs,
+		log,
+	)
+}
+
+// DiscoverGPUsForPod uses the Pod already observed by the node controller to
+// avoid a redundant API read when the immutable Pod spec contains no resource
+// claims. Pods that may use DRA still take the live API path because generated
+// claim names are published asynchronously in Pod status.
+func DiscoverGPUsForPod(ctx context.Context, clientset kubernetes.Interface, pod *corev1.Pod, podName, podNamespace, containerName, hostProcPath string, pid int, log logr.Logger) (compat.GPUInfo, error) {
```
julienmancuso: The optimization itself looks reasonable, but this PR currently conflicts with the recently merged GPU device-path work. When rebasing, please preserve the new OCI environment handling passed into `DiscoverGPUs`; a mechanical resolution could drop the `NVIDIA_VISIBLE_DEVICES` behavior, so this helper will likely need to accept and forward the environment as well as the observed Pod. I’d also add table-driven coverage showing that Pods with resource claims, mismatched identities, and a nil observed Pod still take the live Pod GET path.

